PR #7, opened 2016-06-14
fix for script install / vm install

ezraholm50:
vm ships with ens33 and regular ubuntu distro's with eth0 as default NIC


enoch85:
Nice fix! Please use git via CLI and make the changes to the other VMs as well. Thanks!

PR #21, opened 2016-08-10
Suggested fixes by ShellCheck

enoch85:
Not tested cc @ezraholm50

### DIFF
--- a/static/change_mysql_pass.sh
+++ b/static/change_mysql_pass.sh
@@ -3,13 +3,13 @@
 # Tech and Me, ©2016 - www.techandme.se
 
 SHUF=$(shuf -i 17-20 -n 1)
-NEWMYSQLPASS=$(cat /dev/urandom | tr -dc "a-zA-Z0-9@#*=" | fold -w $SHUF | head -n 1)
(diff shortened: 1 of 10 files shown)


enoch85:
@ezraholm50 Can you please merge all your changes into this PR and I'll check it later today. Thanks!

static/change_mysql_pass.sh
```diff
@@ -3,13 +3,13 @@
 # Tech and Me, ©2016 - www.techandme.se
 
 SHUF=$(shuf -i 17-20 -n 1)
-NEWMYSQLPASS=$(cat /dev/urandom | tr -dc "a-zA-Z0-9@#*=" | fold -w $SHUF | head -n 1)
+NEWMYSQLPASS=$(cat < /dev/urandom | tr -dc "a-zA-Z0-9@#*=" | fold -w "$SHUF" | head -n 1)
```
enoch85: Shoulnd't this be a '>' instead?

ezraholm50: Nope said it should be this way, thought it was strange to, but after changing the error went away so, we'll see.

enoch85:
This PR failed. ShellCheck are wrong, we are right. WIll close.

Issue #28, opened 2016-08-30
ERROR: Site nextcloud_ssl_domain does not exist!


jsahoka:
Hey,
I just downloaded and installed the VM.
The letsencrypt portion of the installer does not seem to work.
This is the last output of the script : 

> **Enabling site 000-default.
> To activate the new configuration, you need to run:
>   service apache2 reload
> 
> Apache will now reboot
> 
> Press any key to continue...
> ERROR: Site nextcloud_ssl_domain does not exist!
> Site nextcloud_ssl_domain_self_signed already disabled

Could it have something to do with this : 

> 
> +---------------------------------------------------------------+
> |    Please enter the domain name you will use for Nextcloud:   |
> |    Like this: example.com, or nextcloud.example.com (1/2)     |
> +---------------------------------------------------------------+
> 
> nc.XXXX.com
> 
> Is this correct? nc.XXXX.com ([y]es or [N]o): y
> /etc/apache2/sites-available/.conf was successfully created
> 

It seems to create a .conf without a filename? Is that supposed to be nextcloud_ssl_domain.conf?

My knowledge of linux and apache is quite underwhelming so I might also be scrwewing something up here. 
Thanks in advance.

jsahoka:
Also, after the script exits it seems that activate-ssl.sh gets deleted, is that by design?

invader303:
so i have had the same Problem with two deployments i did within the last few days but got it fixed after some searching around.

what i did:

- cd /etc/apache2/sites-available
- nano nextcloud_ssl_domain_self_signed.conf and change at the end the following 
![image](https://cloud.githubusercontent.com/assets/21998250/18239011/8510a644-7342-11e6-9afb-911049346304.png)
- save as nextcloud_ssl_domain.conf
- a2ensite nextcloud_ssl_domain.conf
- Service apache2 reload

that should do the trick. good luck

jsahoka:
I was able to fix it by issuing :
sudo cp .conf nextcloud_ssl_domain.conf

Pe46dro:
The `ssl_conf` variable is defined before the definition of `domain` variable. It should be defined at line 168

enoch85:
Make a PR please. Thank you!

ezraholm50:
#31 

Just ran the installer without any problems. Will investigate it...

enoch85:
@all Please test with now when the PR is merged, thanks!

enoch85:
[...]

PR #31, opened 2016-09-08
SSL configuration creation

Pe46dro:
PR as requested #28 

Now the file generation will be made after domain variable definition and configuration will be tested with the correct name

Sorry for multiple commits but i made it from Github Web

### DIFF
--- a/lets-encrypt/activate-ssl.sh
+++ b/lets-encrypt/activate-ssl.sh
@@ -7,7 +7,6 @@ ADDRESS=$(hostname -I | cut -d ' ' -f 1)
 dir_before_letsencrypt=/etc
 letsencryptpath=/etc/letsencrypt
 certfiles=$letsencryptpath/live
-ssl_conf="/etc/apache2/sites-available/$domain.conf"
 SCRIPTS=/var/scripts
 
 # Check if root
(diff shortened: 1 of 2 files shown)


enoch85:
@ezraholm50 Please test this, doesn't seem right at first look. Thanks!

@Pe46dro Thanks for your PR!!

lets-encrypt/activate-ssl.sh
```diff
@@ -242,7 +245,7 @@ fi
 # Check if $certfiles exists
 if [ -d "$certfiles" ]; then
 # Activate new config
-        bash /var/scripts/test-new-config.sh
+        bash /var/scripts/test-new-config.sh $domain.conf
```
ezraholm50: Have you actually tested this?

Pe46dro: Proof of work: http://imgur.com/a/sDALF

I made two gist with the script so you can test it.
https://gist.github.com/Pe46dro/a6e26ce6c3bc6d0a70ff8053d80b5a3c
https://gist.github.com/Pe46dro/c9c1e84c1778f733d54d5954bb7cdd81

enoch85: `test-new-config.sh` is a totally different script that tests the configuration and has nothing to do with the domain name. @Pe46dro Please explain this fix.

enoch85:
I just checked this a little but more and understands the fix, not tested though. 

Lets's merge this anyway and see what response we get from other users. 

👍 From me.

Pe46dro:
Essentially before my change when you run the `lets-encrypt/activate-ssl.sh` file a `.conf` file was generated (only `.conf`, no name, because the `$domain` used in `$ssl_conf` definition was defined before FQDN input ) and when you run the `test-new-config.sh` you always try to enable the `nextcloud_ssl_domain.conf` configuration.

PR #40, opened 2016-09-28
WIP COLLABORA

enoch85:
cc @Pe46dro @ezraholm50 Please review and make necessary changes. Thanks!


oparoz:
1. You're missing the prefix when calling activate-ssl.sh
1. Why delete the script since it tells the user that there is the option to run it later?
1. Let's encrypt only works if the box is connected to the outside world, which may not always be the case
1. At the end of running the script, I didn't know what to do. Richdocument isn't installed. I see the docker running though

oparoz:
OK, so apparently, this doesn't work on a LAN anyway. Too bad ;(

enoch85:
@oparoz I have no clue here actually, it's based on @ezraholm50s work so far. But to answer your questions:

1. What prefix? 
2. WIP here, it depends on when we run it. If we decide to run it in the inital setup script, yes, then it's not neccesary to delete it, but if we run it in nextcloud_install then we would delete activate activate-ssl.sh as the the initial setup script downloads the latest version of the script again.
3. Well, this VM isn't built to be run locally. :)
4. It's not ready yet, and this will be a part of something bigger - which means something else will continue once this is run. 

Thanks for your comments! Please keep testing as we develop it further.

oparoz:
> What prefix?

You're copying the script to a folder but then call it as if it was in your path. 

> Well, this VM isn't built to be run locally. :)

Apart from that issue, it runs well on a LAN and I expect quite a few people use it that way, at least before it's ready to be connected to the Internet, but if Collabora doesn't work without it, then nothing you can do.

enoch85:
> You're copying the script to a folder but then call it as if it was in your path.
Aah yes ofc, blunt miss.

> it runs well on a LAN and I expect quite a few people use it that way,
Yeah, Nextcloud is installed before it's released, so running it in LAN would work.

enoch85:
Maybe we should move all the code to a separate folder just like we do with Let's Encrypt and adjust the Let's Encrypt part so that all the config files are set there directly instead? 

Also TODO:
- [ ] Download and activate Collabra Online Connector on successful setup 

cc @ezraholm50 @Pe46dro

Pe46dro:
Yes i think it's a good idea, to move in a separated folder and leave the final user run it at the end of the installation.
Should the script also check for nextcloud and collabora different domains?

From nextcloud website:
` For security reason we recommend to use a subdomain such as office.nextcloud.com `

ezraholm50:
Indeed we should so it is less prone to fails, if the cert is setup add another -d with office.samedomain and just tell the user they have to set the same ip as the root domain for office.samedomain. This way we rule out any coming user fails.

enoch85:
What I mean is to simply move all the files to a separate dir so that we don't mix the original Let's Encrypt scripts.

But it would be even better if we could get the scripts from Let's Encrypt original folder then just make the necessary changes to the Collabora Let's Encrypt files. That way we don't have to make changes in two places if something is changed in Let's Encrypt original folder. Makes sense?

oparoz:
> Makes sense?

Yes :)

enoch85:
@ezraholm50 I would really need your help here. Thanks!

ezraholm50:
Whatsup mate, I'll come online on irc

ezraholm50:
Only problems we have here are:

* User should have office.yourdomain.xx pointing to the IP already before the installation begins. (mind the delegation, notifing them on install could work/could not) Out Of The Box
* Add a notification that they need to setup office.domain.xx on their DNS page with $THEIRWANIP and run the script, [ if ] fails then notify to try it again 1 hour later (DNS)
* Have the original letsencrypt echo the domains to /var/scripts/DONOTDELETE-domains (or /etc/hosts...) and pull the already present domains, add them to the new ./letsencrypt-auto certonly -d $OLDDOMAINGREPPED -d office.domain.xx

Ark74:
Submitted websocket [update](https://github.com/nextcloud/vm/pull/40/commits/403bd37f946ba4c9ce81d3c27b08a9a849edefa5) for collabora script. 
Cheers!

enoch85:
Let's merge this and improve it even more after that.

@Ark74 Do you think you could get this in a working state? Would be awesome!

static/collabora.sh
```diff
@@ -120,14 +122,11 @@ fi
...
+sleep 10
 wget https://raw.githubusercontent.com/nextcloud/vm/master/lets-encrypt/activate-ssl.sh -P $SCRIPTS
-bash activate-ssl.sh
-rm activate-ssl.sh
-
-
-
-
-
+bash $SCRIPTS/activate-ssl.sh
```
ezraholm50: Is this intended to run at setup? And is the activate-ssl.sh modified already?

enoch85:
I'm thinking that we use Let's Encrypt seperatly and skip the SSL setup in this script as that's already done in activate-ssl.sh. 

Maybe we can export the domian variable in the Let's Encrypt script (as we already do in test-new-config.sh) and run Let's Encrypt before we setup the docker image? I'm on mobile now so I can't test.

ezraholm50:
Yes exactly what i thought, it seems we would only need to: sed -i 's|-d $domain|-d $domain -d $COLLABORA|g' activate-ssl.sh

And already set the path to /etc/letsencrypt/live/cloud-main-domain/xxx.pem for the vhost cert files.

enoch85:
Yup, I fixed it behind nginx last week actually. :grinning: 

I will try to make this better later tonight.

ezraholm50:
Omg, can you enlighten me? I really want to set it up aswell!

enoch85:
Just have to get home first. Make sure you'll be online on IRC later tonight. :)

ezraholm50:
No problem!

PR #51, opened 2016-10-20
contacts are now moved to Nextcloud

### LINKED ISSUE #48: Script to install contacts has to be modified
Contacts has moved to nextcloud
Contacts doesn't have any releases any more, so the package can't be fetched.

enoch85:
Still wrong version number though as 1.5.0.0 is not released yet. 

Fixes #48

@oparoz for review. Thanks!

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -478,7 +478,7 @@ apt-get update -q2
 apt-get install webmin -y
 
 # Nextcloud apps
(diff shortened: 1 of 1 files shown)


oparoz:
This is not enough (yet) because the releases have not been migrated. Packages created by tags do not contain a valid signature.

enoch85:
Ok, when will the releases be migrated?

enoch85:
> Ok, when will the releases be migrated?

cc @irgendwie

enoch85:
The version is fixed

PR #52, opened 2016-10-20
Update GPG key

### LINKED ISSUE #49: Use a PGP keyserver which doesn't require a hole in the firewall
I suggest you use this URL to check the validity of the Nextcloud package:
`hkp://p80.pool.sks-keyservers.net:80`

enoch85:
Fixes #49 

@oparoz Is this what you mean?


oparoz:
:+1: 

Yep, should work much better when behind a corporate firewall :)

PR #56, opened 2016-11-03
Use whiptail to install apps [WIP]

enoch85:
cc @nextcloud/vm


enoch85:
This is done in the install script instead. Alredy merged to master

Issue #50, opened 2016-10-20
Make a backup of network/interfaces before overwriting it


oparoz:
It would be a good idea to make a backup of /etc/network/interfaces before overwriting it, because sometimes the scripts are used outside the VM because not everything needs to be installed, etc. and overwriting an existing config can break things.

enoch85:
Not priority. Please make a PR. :)

Thanks for you issue!

enoch85:
@oparoz Now when thinking about it, the VM scripts are designed to be install on a clean system, and we overwrite it ti get the correct variables for example $IFACE. So why save it?

oparoz:
It's true that, because I didn't want everything to be managed by the script, I used it on a fresh install of Ubuntu server and that part of the script did not work and left the VM disconnected from the network.
Maybe there wouldn't have been a problem if we had booted the image, but since other things broke at install time, I'm just thinking that making a backup before touching critical config files isn't such a bad idea.

enoch85:
Please explain further about the bug you found so that we can fix it. The VM is designed to fit ALL systems, but at the same time you have to mount it correctly (Bridged mode i.e.) for it to work. 

Example: The VM uses ens33 as IFACE (network interface) when created by us, then you as a user downloads it and have another interface (let's say eth0), the script will fail with the ens33 IFACE so that's why we change it. In the install script we only check that the network is OK without changing anything as you probably use the install script to install the VM for the first time. In the startup script though (the one that is pre installed and run when the user boots the released VM for the first time) we do change the interfaces file becuase of the reason above and then test that the network is OK here: https://github.com/nextcloud/vm/blob/master/nextcloud-startup-script.sh#L42-L61

The "correct interface" is just the original interface on Ubuntu 16.04 with new IFACE, so that it breaks seems unlikley, and even if it did, you are not able to continue as we will stop you before any scripts are downloaded. That's by design. The scripts are fetching all the latest scripts from Github and it will break without them, but again - we test if your network connection is OK before running anything.

So where did it break?

enoch85:
Maybe we should add more checks? @oparoz

oparoz:
The interface was configured with a static IP and the script replaced the config with one which uses DHCP and that didn't work. That's how it broke.

I guess I could also write something which makes a backup of these files before trying to install NC...

enoch85:
[...]

PR #59, opened 2016-11-13
Avoid overwriting interfaces when installing on VPS

enoch85:
Please review @nextcloud/vm 

Possible fix for https://github.com/nextcloud/vm/issues/50

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -43,7 +43,6 @@ fi
 # Check network
 echo "Testing if network is OK..."
 sleep 2
-sudo ifdown $IFACE && sudo ifup $IFACE
 service networking restart
     wget -q --spider http://github.com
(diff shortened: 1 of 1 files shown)


nextcloud-startup-script.sh
```diff
@@ -39,16 +39,29 @@ then
     exit 1
 fi
 
+
+# Check network
+echo "Testing if network is OK..."
+sleep 2
+sudo ifdown $IFACE && sudo ifup $IFACE
```
Pe46dro: Usually when i use VPS i connect to it via SSH and i think that turn off the network interface where i'm connected It is not a great idea 😄

enoch85:
@Pe46dro Better?

enoch85:
This may cause issues if the user has another IFACE when mounting the VM, but let's merge and see if we get any issues from users.

PR #80, opened 2017-01-05
Integrante Nextant app

Ark74:
Hi,
I would like to submit this script that install solr and setup nextant app.
Feedback is welcome.
Cheers.

### DIFF
--- a/static/collabora.sh
+++ b/static/collabora.sh
@@ -97,8 +97,8 @@ else
   ProxyPassReverse    /hosting/discovery https://127.0.0.1:9980/hosting/discovery
 
   # Main websocket
-  ProxyPass   /lool/ws      wss://127.0.0.1:9980/lool/ws
-
+  ProxyPassMatch   "/lool/(.*)/ws$"      wss://127.0.0.1:9980/lool/$1/ws
+  
   # Admin Console websocket
   ProxyPass   /lool/adminws wss://127.0.0.1:9980/lool/adminws
 
--- a/static/nextant.sh
+++ b/static/nextant.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# Solr Server & Nextant App Installation
+
+# Setting variables
+NT_RELEASE=nextant-master-1.0.3.tar.gz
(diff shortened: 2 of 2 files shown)


Ark74:
How to just make a pull request for the last commit? 
(1d6a237)

enoch85:
Ok, I will merge this, but I will not integrate it automatically, not even download the script to begin with. This needs some testing first. 

Thanks for your PR! :)

enoch85:
@Ark74 Also, please add some testing and if arguments that exits on failure. Don't expect it to work every time.

Ark74:
Sure, i'll keep adding some scripting to check and exit on error.
This sketch runs straight forward after the install_production.sh script

enoch85:
@Ark74 I improved it a bit more. 

>This sketch runs straight forward after the install_production.sh script

No it doesn't, it needs to be added as a function here: https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L494-L577

enoch85:
But even better would be to add it to the startup-script here: https://github.com/nextcloud/vm/blob/master/nextcloud-startup-script.sh#L84 and here: https://github.com/nextcloud/vm/blob/master/nextcloud-startup-script.sh#L355

Ark74:
> No it doesn't, it needs to be added as a function here

No, i mean that on its current stage it can be (manually) run right after install_production and it will work.

Maybe this can be added to the 
* Contacts
* Calendar
* Spreed

list to be enabled or find its path the same way that collabora will.

Thanks again.

enoch85:
We will see, the I prefer to let the user choose. I started a poll here: https://twitter.com/techandme_se/status/817044233437716480

`nextcloud_install_production.sh` is just used to prepare Nextcloud, while the startup script is run when the VM is downloaded an mounted. I plan to improve the choice there as well, so that the user can choose to install several apps at the same time. But that's for later. If you want you can copy the code from the install script and put it in the startup script and ask the user if he/she wants to install Nextant and SpreedME, collabora is not stable yet, so we'll let that be. Or have you tested the collabora script recently? Is it stable? Does it work?

Btw, join #techandme on IRC so that we can chat there.

PR #86, opened 2017-01-07
update

ezraholm50:


ezraholm50:
My sincere apologies, wrong pull req trying to update @NextBerry

PR #90, opened 2017-01-15
Always keep local configuration when updating

enoch85:
Please review @nextcloud/vm

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -29,7 +29,7 @@ NCVERSION=$(curl -s --max-time 900 $NCREPO/ | tac | grep unknown.gif | sed 's/.*
 
 # System Upgrade
 apt update
-apt dist-upgrade -y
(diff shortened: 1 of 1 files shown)


enoch85:
Maybe a better option is to re-add all the custom configs in php.ini (this is the fix for that) if they don't exist after upgrade instead of forcing something by default?

Changes are made to max upload + redis.so and smbclient.so as extentions in php.ini...

Ark74:
That would be a wider approach. :)
BTW, i thought you refer to the nextcloud config.php not the php config.

enoch85:
> That would be a wider approach. :)

What would be?

Ark74:
Look up for strings like 
- redis.so
- smbclient.so
- upload_max_filesize = 1000M
- memory_limit = 512M

At php.ini right after `apt dist-upgrade` so if they are not, they get re-added, if they are, nothing else happen so that way wherever the user choose nextcloud would not be compromised.
Since the system is a nextcloud vm/box.

enoch85:
Ok, so you think that approach is better?

enoch85:
Let's merge this for now, and write a better solution later on.

PR #79, opened 2017-01-03
[WIP] Improve the Collabora script

enoch85:
I tested the script and there are no script errors, but why get a cert for the main domain? Makes no sense imo.

About using the standard Let's Encrypt scripts... no need as this is only for setting up the subdomain, doesn't have to be that advanced, but improvements are welcome.

Me and @ezraholm50 were discussing that we maybe should start with the Let's Encrypt scripts to setup the SSL config, and then use that address like $1 as we do in test-new.config.sh... It's an idea, but don't know if it's optimal.

Please review and test this script live please @nextcloud/vm

PR #91, opened 2017-01-16
New branch for another approach for collabora script

Ark74:
The script is almost complete, it has some issues when checking for the DNS resolution (commented out).
I'll keep testing, comments are welcome.

### DIFF
--- a/static/collabora.sh
+++ b/static/collabora.sh
@@ -1,61 +1,9 @@
 #!/bin/bash
-# Collabora auto installer
 
-## Variable's
-# Docker URL
-DOMAIN=$(whiptail --title "Techandme.se Collabora" --inputbox "Nextcloud url, make sure it looks like this: cloud\\.yourdomain\\.com" "$WT_HEIGHT" "$WT_WIDTH" cloud\\.yourdomain\\.com 3>&1 1>&2 2>&3)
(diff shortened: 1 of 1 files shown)


enoch85:
Thanks for the PR, but please use the new version I pushed to master just an hour ago. Made some changes there and added some checks. Will have a look at this tomorrow.

enoch85:
*Copied from https://github.com/nextcloud/vm/pull/79*
@Ark74 

Please remember, this VM is not for users like us, it's for users with NO knowledge in Linux, SSL, HTTP, HTTPS, Dockers, VMs or whatever.
That's right, we should also take into account that ssl certs bring some administration responsibility.
I'm thinking about splitting into

Standard mode
Advance mode
So the standard one uses Let's Encrypt and the advance could use the users certs by copying-pasting the content of the crt/key/pem files.

Do you want to be added as a maintainer?
I sure would like, but let me figure github a little more, i would not like to mess something up by ignorance.

enoch85:
@Ark74 Please use the latest master to fix the issues with the conflicts. Thanks!

enoch85:
@Ark74 I've fixed Collabora now so that it's stable. Will add even more checks to make it more user friendly. 

Sorry, but I don't see the purpose any longer for this PR. Feel free to continue or to close.

Thanks!

Ark74:
Your are right. And to be honest i haven't had the time to really work on it.
Let's continue from that one.

PR #95, opened 2017-01-18
Put MySQL password in .my.cnf 

enoch85:
To be able to login with "mysql -u root"

Also, remove $PW_FILE


enoch85:
Tested and works!

Issue #92, opened 2017-01-16
extended time to get web front end


razorfish-sl:
people complain that they cannot get the front end web up when logging in for variable amounts of time upto 30 seconds.

I have had same problem as well as issues related to showing the apps tab, taking hour to get a result.

This appears to be a combination of VM settings and network settings related to  packet size on the outgoing interface.

since the vm assumes it is running as a 'real' computer, it uses real computer setting with the MTU set to 1500 bytes, normally this is not an issue.
However the packet size is increased beyond this size in some VM setups, this ALWAYS causes the packets to fragment on external interfaces. Depending on the network settings this causes the packet no ends of issues.

simply setting the interface MTU to a smaller size should fix this problem for all

in nextcloud  linux CLI do:

sudo ifconfig -a   #to list networks

then
sudo ip link set dev {interface from above list} mtu 1400

it may be worth configuring the VM with a slightly smaller MTU to fix  this issue globally.

enoch85:
Thank you so much for posting this issue! 

A PR is welcome :)

enoch85:
Ok, a PR is here; https://github.com/nextcloud/vm/pull/96

And here is some confirming information about the issue: https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1359

razorfish-sl:
Thats the problem with the internet.... all the info is there ,
 it's just linking a problem to the right solution...

enoch85:
@razorfish-sl Are you happy with the fix? It broke the internet in the scripts so I had to comment it out: https://github.com/nextcloud/vm/commit/e24196af357a625edac5d1535c0a432a628f6eed

PR #96, opened 2017-01-18
set MTU to 1400

enoch85:
Fixes https://github.com/nextcloud/vm/issues/92

cc @razorfish-sl

### DIFF
--- a/static/ip.sh
+++ b/static/ip.sh
@@ -19,6 +19,7 @@ iface lo inet loopback
 iface $IFACE inet static
 pre-up /sbin/ethtool -K $IFACE tso off
 pre-up /sbin/ethtool -K $IFACE gso off
+mtu 1400 # Fixes https://github.com/nextcloud/vm/issues/92
 
 # Best practice is to change the static address
 # to something outside your DHCP range.


enoch85:
Tested and works!

Ark74:
Shouldn't the comment be on the previous line?
I've had some issues in the past with comments on the same line on /etc/network/interfaces

Cheers.

razorfish-sl:
ok, setting the MTU should not be done like above fix

something like:

pre-up  ip link set dev $IFACE  mtu 1430

but it must only be for "real" interfaces  not the loopback.

enoch85:
@razorfish-sl

I got this:

> ok, setting the MTU should not be done like above fix

something like:

pre-up ip link set dev $IFACE mtu 1430

but it must only be for "real" interfaces not the loopback.

in my email, but nothing here. Can you make a PR please?

razorfish-sl:
yep , i got some emails as well but nothing in the repository

enoch85:
@razorfish-sl Where should I set `pre-up ip link set dev $IFACE mtu 1430`? Can you make a PR please?

PR #100, opened 2017-01-28
Updater complains about assets folder

Ark74:
Maybe the folder is `updater` as pointed out here[0]
https://docs.nextcloud.com/server/11/admin_manual/installation/installation_wizard.html#setting-strong-directory-permissions


enoch85:
Thanks @Ark74 !

PR #107, opened 2017-01-31
Change UNIXUSER to $SUDO_USER

enoch85:
If a user installs with a different user, the script should work anyway. Though ncadmin is always the standard user and changed to NCUSER instead.


enoch85:
@ezraholm50 @Ark74 Please check this out. Will test it soon.

enoch85:
Last commit should be $SUDO_USER not daniel lol.

PR #114, opened 2017-02-04
Collabora enhancements

enoch85:
Fix some smaller things like adding the App and better failure messages to the user. Also add some more checks.

### DIFF
--- a/static/collabora.sh
+++ b/static/collabora.sh
@@ -45,7 +45,7 @@ calc_wt_size() {
 }
 
 # Notification
-whiptail --msgbox "Please before you start make sure port 443 is directly forwarded to this machine or open it!" "$WT_HEIGHT" "$WT_WIDTH"
+whiptail --msgbox "Please before you start, make sure that port 443 is directly forwarded to this machine!" "$WT_HEIGHT" "$WT_WIDTH"
 
 # Get the latest packages
 apt update -q2
@@ -128,12 +128,13 @@ fi
 
 # Check of docker runs and kill it
 DOCKERPS=$(docker ps -a -q)
-ISRUNNING=$(docker inspect -f {{.State.Running}} $DOCKERPS > /dev/null)
-if [ $ISRUNNING = "true" ]
+if [[ $DOCKERPS > 0 ]]
 then
-    echo "Removing old Docker instanses..."
+    echo "Removing old Docker instance... ($DOCKERPS)"
     docker stop $DOCKERPS
     docker rm $DOCKERPS
+else
+    echo "No Docker instanses running"
 fi
 
 # Disable RichDocuments (Collabora App) if activated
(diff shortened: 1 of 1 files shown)


static/collabora.sh
```diff
@@ -100,6 +114,23 @@ else
...
+    echo "Removing old Docker instanses..."
+    docker stop $DOCKERPS
+    docker rm $DOCKERPS
+fi
+
+# Enable RichDocuments (Collabora App)
+if [ -d $NCPATH/apps/richdocuments ]
+then
+    sudo -u www-data php $NCPATH/occ app:disable richdocuments
+    rm $NCPATH/apps/richdocuments
```
ChristophWurst: ``rm -r`` since it's probably a directory?

enoch85: Thanks!

enoch85:
@ChristophWurst Anything else?

enoch85:
Docker check fails:

```
See 'docker inspect --help'.

Usage:	docker inspect [OPTIONS] CONTAINER|IMAGE|TASK [CONTAINER|IMAGE|TASK...]

Return low-level information on a container, image or task
collabora.sh: line 132: [: =: unary operator expected
```

enoch85:
Tested and works! 

Left to do: request occ command for entering $SUBDOMAIN with occ.

ChristophWurst:
> enoch85 commented a day ago
> 
> @ChristophWurst Anything else?

Now that you merged the PR anyway, I'm curious why you even care about the extra work of preparing a PR if you do not request others to review your changes? I see that other PRs were merged in the same way. Usually we do PRs for Nextcloud projects so that at least two people have seen the changes (the author and a second person). If you merge the PRs yourself, it's like you commit and push to master directly.

enoch85:
@ChristophWurst Yeah, you're right. When @jnweiger and me first started to develop the ownCloud VM we decided to push to master directly, and since I don't know git that well I continued down that path.

Now I'm starting to learn git and just understood how to make PRs, verify your commits and such. Making PR's is very easy as you already know;

```
git checkout -b new-branch
git commit -S -a -m "commit message"
git push origin new-branch
```

Then make a PR out of that new branch.

Anyway, to answer your question. PR's are very good because others can comment and review like you just did, but it also takes time as the designated maintainers are often away or busy doing other things. In this case I wanted to push out the new fix asap as the collabora script is included in the production version and was failing atm. So I tested it myself, confirmed it to work, and then pushed it.

**TL:DR** I wanted to push this as fast as possible as there could be some risk that the collabora script was failing without these changes (checks and such). 

Hope that explains it. Thanks for your input. :)

ChristophWurst:
I see your point, that is a valid approach I would say. I'd always prefer having PRs over direct master commits/pushes anyway. So when you do merge your own PRs without a review, it would still be nice to leave a comment why you did so. This way others are able to catch up on what has been changed in the repository recently and why there's no review 😉

PR #117, opened 2017-02-06
Update nextcloud_install_production.sh

ezraholm50:
double wget of nextcloud-startup-script.sh @ (search for  # Get nextcloud-startup-script.sh)
+ some spaces to few or to many


enoch85:
Thanks!

PR #120, opened 2017-02-06
Automated networking

ezraholm50:
Automates the test_connection.sh script and only prompt nano /etc/network/interfaces when the connection fails. It uses the DHCP connection.

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -475,7 +475,8 @@ then
     sleep 1
     echo
     CONTEST=$(bash $SCRIPTS/test_connection.sh)
-    if [ "$CONTEST" == "Connected!" ] then
+    if [ "$CONTEST" == "Connected!" ] 
+    then
         echo "Connected!"
         echo
         echo "We will use the DHCP address for now, if you want to change it later then just"
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -477,16 +477,19 @@ then
     CONTEST=$(bash $SCRIPTS/test_connection.sh)
     if [ "$CONTEST" == "Connected!" ] 
     then
-        echo "Connected!"
+        # Connected!
+        echo -e "\e[32m$CONTEST\e[0m"
         echo
         echo "We will use the DHCP address for now, if you want to change it later then just"
-        echo "edit the interfaces file. If you experience any bugs, please report it here:"
+        echo "edit the interfaces file: sudo nano /etc/network/interfaces"
+        echo "If you experience any bugs, please report it here:"
         echo "https://github.com/nextcloud/vm/issues/new"
         echo -e "\e[32m"
         read -p "Press any key to continue..." -n1 -s
         echo -e "\e[0m"
-    else    
-        echo -e "\e[31mNot Connected!\e[0m\nYou should change your settings manually in the next step."
+    else
+        # Not connected!
+        echo -e "\e[31m$CONTEST\e[0m\nYou should change your settings manually in the next step."
         echo -e "\e[32m"
         read -p "Press any key to open /etc/network/interfaces..." -n1 -s
         echo -e "\e[0m"
@@ -500,7 +503,7 @@ then
     ifup $IFACE
     sleep 1
     echo
-    bash $SCRIPTS/test_connection.sh
+    $CONTEST
     sleep 1
 else
     echo "OK, then we will not set a static IP as your VPS provider already have setup the network for you..."
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -479,6 +479,9 @@ then
         echo "We will use the DHCP address for now, if you want to change it later then just"
         echo "edit the interfaces file. If you experience any bugs, please report it here:"
         echo "https://github.com/nextcloud/vm/issues/new"
+        echo -e "\e[32m"
+        read -p "Press any key to continue..." -n1 -s
+        echo -e "\e[0m"
     else    
         echo -e "\e[31mNot Connected!\e[0m\nYou should change your settings manually in the next step."
         echo -e "\e[32m"
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -476,6 +476,8 @@ then
     echo
     CONTEST=$(bash $SCRIPTS/test_connection.sh)
     if [ "$CONTEST" == "Connected!" ] then
+        echo "Connected!"
+        echo
         echo "We will use the DHCP address for now, if you want to change it later then just"
         echo "edit the interfaces file. If you experience any bugs, please report it here:"
         echo "https://github.com/nextcloud/vm/issues/new"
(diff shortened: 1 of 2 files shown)


nextcloud-startup-script.sh
```diff
@@ -474,15 +474,18 @@ then
...
-    echo -e "\e[0mIf the output is \e[31mNot Connected!\e[0m you should change\nyour settings manually in the next step."
-    echo -e "\e[32m"
-    read -p "Press any key to open /etc/network/interfaces..." -n1 -s
-    echo -e "\e[0m"
-    nano /etc/network/interfaces
+    CONTEST=$(bash $SCRIPTS/test_connection.sh)
+    if [ "$CONTEST" == "Connected!" ] then
+        echo "We will use the DHCP address for now, if you want to change it later then just"
+        echo "edit the interfaces file. If you experience any bugs, please report it here:"
+        echo "https://github.com/nextcloud/vm/issues/new"
```
enoch85: Add "Press any key to continue..." Otherwise the user won't have a chance to see the text.

ezraholm50:
Done

enoch85:
Nice, will test it.

nextcloud-startup-script.sh
```diff
@@ -476,6 +476,8 @@ then
     echo
     CONTEST=$(bash $SCRIPTS/test_connection.sh)
     if [ "$CONTEST" == "Connected!" ] then
+        echo "Connected!"
```
ezraholm50: Added this else they don't get confirmation its really connected opposed to "not connected" and its looks bad without.

Ark74:
I just tested a vm run with the _stable_ version and it turns out it won't have LAN connection.
I looked at the interfaces file and it sets static IP, and i thought it was due the missing the 
```
auto enp0s3
iface enp0s3 inet static
```
(_been **enp0s3** the lan interface_
`ls /sys/class/net`
will say which devices are available.)
but no luck.

My router has dhcp enabled so i guess that would be the issue, right?. Anyone having this issue?

ezraholm50:
This has nothing to do with this pull request. Please make a new issue. But indeed it should not set a static IP?

enoch85:
@Ark74 Where in the script did it fail? Did you run the install script or the startup-script?

Ark74:
It didn't fail. It runned well from start to end (one  reboot).
Then runs the startup script and _fixes_ the network interface to static.

After rebooting again I have no network connection.

nextcloud-startup-script.sh
enoch85: Make it green! :D

enoch85:
@Ark74 Strange! Did you manually set anything in the interfaces file? 

We have like 100 downloads per day, and the network errors are long gone. Nobody has reported such a error in months!

Ark74:
I did enable mtu value to test. Just that.

enoch85:
Yeah, but that's another issue. That's why I commented it out.

enoch85:
Yup, as I suspected, you can't run a if command in another if command. 

This script fails.

```
Setting hostname...
/var/scripts/nextcloud-startup-script.sh: line 487: syntax error near unexpected token `else'
/var/scripts/nextcloud-startup-script.sh: line 487: `    else    '
root@nextcloud:~# 
```

Ark74:
I don't think is that since I have disable it.
$IFACE is not returning a value.

I had to manually set enp0s3
I'm using  a VM with bridge network interface.

enoch85:
@Ark74 Please run the scripts with debug set to 1 and try to find out why this happens. I don't think this is common as we have no reports about it, but I would really like to know what is the cause for you so that we can fix it.

enoch85:
Fixed the errors, now it should work

Ark74:
I'll run another instance at night, and report.
I'm out of the city with no internet connection.

Ark74:
Tue Feb  7 18:38:09 2017
md5sum nextcloud_install_production.sh 
e1dcda25512a39fcc273aca2570d9d9f

Same result.
There is no interface declared on `/etc/network/interfaces
`

enoch85:
Funny that you say that, because leaving the static ip config out if th eqvation is actually a feature after I ran the scripts on Digitalocean. Everything is pre-configured on VPSes and we don't wanna touch that. :)

It must be something else.

Thank you for investigating.

Ark74:
>  leaving the static ip config out if th eqvation is actually a feature
This is right what i mean. If you choose No (to the VPS) it tries to set up static IP and thus the issue i'm getting.
I'll share my installation screen, it is 60% up already. It may take a couple of hours to complete.  :-/

Ark74:
Since it is gonna take longer than i'll be around. Here the (pre)uploaded video link: https://youtu.be/WFdylR0TK3o

enoch85:
> If you choose No (to the VPS) 

But if you are on a VPS you should answer "yes". I'm confused here...

enoch85:
Btw, the Youtube clip is still not working...ping @Ark74

enoch85:
@Ark74 Ok I saw the video. 

1. This is run locally, right?
2. If yes, login to the hypervisor directly instead and check which IP you have with ifconfig.

Ark74:
1. Yes
2.
 nc-ifconfig

>[I looked at the interfaces file and it sets static IP, and i thought it was due the missing the
auto enp0s3
iface enp0s3 inet static](https://github.com/nextcloud/vm/pull/120#issuecomment-277822382)

>[ $IFACE is not returning any value, so
iface [interface] inet static](https://github.com/nextcloud/vm/pull/120#issuecomment-277836901)

cat /etc/network/interfaces
cat-interfaces

enoch85:
Great! We are getting somewhere.

Do `lshw -c network | grep "logical name" | awk '{print $3; exit}'` and check the output.

> i got just a couple hours of sleep before the day begin. 

I know the feeling. Hunting bugs from dusk til dawn. ;)

enoch85:
Note to self, add check to see if iface is missing(?)

Ark74:
Found the issue.
(this is not my VM, but is the same)
```
  *-network
       descripción: Interfaz inalámbrica
       producto: AR9285 Wireless Network Adapter (PCI-Express)
       fabricante: Qualcomm Atheros
       id físico: 0
       información del bus: pci@0000:06:00.0
       **nombre lógico: wlan0**
       versión: 01
       serie: f0:7b:cb:64:86:28
       anchura: 64 bits
       reloj: 33MHz
       capacidades: pm msi pciexpress bus_master cap_list ethernet physical wireless
       configuración: broadcast=yes driver=ath9k driverversion=4.2.0-42-generic firmware=N/A ip=192.168.0.107 latency=0 link=yes multicast=yes wireless=IEEE 802.11bgn
       recursos: irq:18 memoria:e8100000-e810ffff
```
It depends on the language it is set.
If i have Spanish, then it is not `grep "logical name"` but `grep "nombre lógico"` 

There it is. So maybe force English or find another way to get the name of the interface.

enoch85:
GREAT! Nice catch! :clap: 

Well, the solution is ti change language after the network is set. Will fix that ASAP.

enoch85:
Commit: https://github.com/nextcloud/vm/commit/6136cf6a5fb4698aadff12a1431ec49e37f5e747

WIll also work on a failsafe in case the user is running the scripts from start with another language. PR is coming soon.

PR #124, opened 2017-02-08
Failover IFACE

enoch85:
If the user uses another language than English for system language then "logical name" isn't grepped correctly. This intends to fix that as we don't grep "logical name" but uses another method.

@Ark74 please review.


Ark74:
All seems in place, i have tested the commands and they returns the correct output.
:+1:

enoch85:
Did you run the whole script?

Please join #techandme on IRC @freenode so we can chat.

PR #126, opened 2017-02-08
Update collabora.sh

ezraholm50:
https://help.nextcloud.com/t/does-official-vm-image-contain-collabora-online-office/8427/20


enoch85:
OK for now, but please find a better fix @ezraholm50

PR #129, opened 2017-02-11
Workaround for uncommon download url

Ark74:
Here a patch for the issue with uncommon urls.
Please feel free to improve the approach.

### DIFF
--- a/static/nextant.sh
+++ b/static/nextant.sh
@@ -101,9 +101,21 @@ else
 fi
 
 # Get nextant app for nextcloud
-wget -q -P $NC_APPS_PATH $NT_DL
-cd $NC_APPS_PATH
-tar zxf $NT_RELEASE
+NT_DL_CK=$(wget -q -P $NC_APPS_PATH $NT_DL)
+if [ $? -ne 0 ]; then
+    echo "There seems to be an issue detecting the latest release version published"
+    echo "Please copy/paste the tar file url for the latest release download, manually"
+    echo "(e.g. https://github.com/nextcloud/nextant/releases/download/v1.0.3/nextant-master-1.0.3.tar.gz )"
+    read NT_DL_AL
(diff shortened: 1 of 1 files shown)


enoch85:
Thanks for you PR!

If I understand this correct, this check is the automatic download link works, and if not it asks the user to download it manually? 

Why is this needed? If the download link doesn't work, there is something wrong with GitHub, and it won't work to download it manually either.

Ark74:
It asks for the tar url download for the latest release.
Then it continues.

It is for testing purposes.

The main idea is to add some trigger to avoid the script from breaking with issues like this.

Ark74:
Anyway there are issues with nextant 1.0.5, since it won't test and save the configuration.

Ark74:
@enoch85 i would keep the 
`apt update -qq`

If you wanna run the script by its own.

enoch85:
@Ark74 Not so fast! :) I just asked yo a question, not to close it. 

Actually the script breaks if the tag is wrong. I just tested. @daita had forgot to change the tag, and just uploaded the new version and the script broke as it failed to download 

I fixed it now and also told him it's inmportant to tag the releases correctly, but this fix isn't so bad maybe... If that should happen again.

The reason for my question was that I don't like when users have to manually do something. Tech and Me should do as much automatically as possible- Make tech easier. ;)

enoch85:
I have an idea for a fix for this...

If the tag is wrong, the download will fail. If it fails, try tag before the newest tag. That way it would still download the app, but a lower version. Could also put this in a while loop, so that it tries all versions. If all versions fail, the trow an error with exit 1.

@Ark74 Are you up for the task?

enoch85:
Here are the version tag: https://github.com/nextcloud/vm/blob/master/static/nextant.sh#L7

It fetches the latest, but if you remove latest you get all tags...

Ark74:
Sounds like a good idea to get the earliest and correct tag.
l'm about to go off for today.
Thanks for the review, let's check that as soon as we hear from @daita

Cheers!

PR #130, opened 2017-02-11
Update collabora.sh

ezraholm50:
Fixes access not allowed on my installation and why have it listen on all addresses/domains when you need only one.

### DIFF
--- a/static/collabora.sh
+++ b/static/collabora.sh
@@ -178,7 +178,7 @@ then
 else
         touch "$HTTPS_CONF"
         cat << HTTPS_CREATE > "$HTTPS_CONF"
-<VirtualHost *:443>
+<VirtualHost $SUBDOMAIN:443>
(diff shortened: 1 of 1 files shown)


enoch85:
True. Let's get this in. 

Not tested**

branflakem3:
I think this change is not properly creating the /etc/apache2/sites-available/office.domain.com.conf file. This is what the first 2 lines of my file looked like:
`<VirtualHost office.domain.com:443>`
`  ServerName office.domain.com:443`

The correct syntax (which fixed my problem) should be:
`<VirtualHost *:443>`
`  ServerName office.domain.com:443`

With the subdomain in there my whole site would only serve the office.domain.com cert. This caused cloud.domain.com to not function.

enoch85:
@branflakem3 You have one config file for your cloud and one for the subdomain. Both are active.

PR #138, opened 2017-02-23
Solr >= 6.4.1 for nextant

Ark74:
Force the latest 6 release for Solr.

### DIFF
--- a/static/nextant.sh
+++ b/static/nextant.sh
@@ -42,7 +42,7 @@ echo "Starting to setup Solr & Nextant on Nextcloud..."
 sleep 3
 
 # Installing requirements
-apt update -qq
+apt update -q2
 apt install default-jre -y
 
 # Getting and installing Apache Solr
--- a/static/nextant.sh
+++ b/static/nextant.sh
@@ -3,7 +3,7 @@
 
 # Setting variables
 STATIC="https://raw.githubusercontent.com/nextcloud/vm/master/static"
-SOLR_VERSION=$(curl -s https://github.com/apache/lucene-solr/tags | grep -P -m 3 -o '<span class="tag-name">.+/\K.+?(?=</span>)' | grep -m 1 6)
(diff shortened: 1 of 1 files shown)


enoch85:
@Ark74 What's the difference? It's already getting the latest version...

enoch85:
Ok just tested. This makes it non automatic as it greps 6 and one. In that case we could just gor for a simple variable that states which version to get.

Please improve this to be fully automatic. Thanks!

enoch85:
Advice: look att the version function in the update script.

Ark74:
Nextant requires release 6. 
The stable branch will install the latests release (5.5.4),  but not the latest version (6.4.1).
The change i made only greps the latest release for the 6 version, it's not and automatic change but it will get you the latest 6 release.

The stable branch as for today install solr 5.5.4 which will not work.

enoch85:
@Ark74 Yes, I read what you wrote before so I'm aware of that. 

What I was asking was to always get the version with the "highest" number. I.e. 6410 is a higher value than 5540. So if you can get the script to calculate that and get the correct version number no matter what, then I'll merge. :)

You can alwasy ask in #bash on Freenode on IRC.

Ark74:
@enoch85 we shouldn't look the latest release available in time. 
We should look for the latest functioning release in time, being 6.X at the time.
Getting the latest will repeat the error, in case there is a new unsupported version. Let's say 7.X.
If we get the latest in this case:
5.X
6.X
7.X
We'll get errors on both 5 and 7, since there is no current support for those.
Cult did mention that we should be careful of bleeding edge releases.

enoch85:
@Ark74 So what you're saying is that only 6.X.X will work?

Ark74:
Yes, pretty much.
Until proved otherwise when 7 is released.

Ark74:
@daita  could you confirm or correct [this](https://github.com/nextcloud/vm/pull/138#issuecomment-282088302).

PR #139, opened 2017-02-23
Ask for keyboard layout change

### LINKED ISSUE #140: Ask users to change variables
I think that for the sake of automation/ease and user preference we should ask if the user wants/requires to change variables such as:
* Static IP / Over DHCP
* [System user password](https://github.com/nextcloud/vm/blob/Ark74-ask-kblo/nextcloud-startup-script.sh#L667)
* [NC user password](https://github.com/nextcloud/vm/blob/Ark74-ask-kblo/nextcloud-startup-script.sh#L683)
* Also keyboard layout changes #139 

I've proposed that every question only waits for certain time, if no interaction then the default value will be applied.
Which in most of the cases is a denial of the change.

Ark74:
Instead of forcing the keyboard change, ask user or skip it if not answer given.

Fixes #140

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -446,7 +446,7 @@ then
     echo -e "in your router later. It's included in this guide:"
     echo -e "https://www.techandme.se/open-port-80-443/ (step 1 - 5)"
     echo -e "\e[32m"
-    echo "We'll now setup the VM static IP"
+    read -p "Press any key to set static IP..." -n1 -s
     sleep 2
     echo -e "\e[0m"
     ifdown $IFACE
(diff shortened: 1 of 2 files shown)


enoch85:
Thanks for your PR @Ark74 

Please be consistent and use the same Yes/No function as here: https://github.com/nextcloud/vm/blob/407c5d7de95fa1bbf26e4a9ca0e7ee9df05b6d8d/nextcloud-startup-script.sh#L429-L437 his one catches all the digits, instead of just N or Y

Ark74:
Sure, i'm on it.

enoch85:
You still haven't added the function I was talking about...

enoch85:
Oh just saw it... The reason why the user have to hit any key, is to read the instructions about opening ports before. Please revert. Sorry, I should have checked before. Kind of tired now.

And still no function as described [here](https://github.com/nextcloud/vm/pull/139#issuecomment-281941994)

Ark74:
It is on both of the keyboard layout reconfiguration.
Install and startup scrip.

enoch85:
Tired now, but that's what I meant all along... you missed the function part. ---^

Ark74:
I'll study that part. Thanks.
Read you in the morning.

PR #143, opened 2017-02-23
Alternative to v7

Ark74:
Since we are not sure if Solr 7 will or will not be compatible, we could just restrict v5 for now.
Did i got it right @daita ?


enoch85:
Not relevant, already fixed.

PR #145, opened 2017-03-01
Change from curl to wget

enoch85:
We mostly use `curl` to test stuff, and `curl` doesn't really download stuff, it exists with 0 even if the domain isn't available. This means that other parts on the script may fail as we think everything is running smoothly even though it doesn't 

We've used the `wget` command in this PR with success and it does a correct check (it actually tries to download the domain), but I'm not sure it applies to everything so this has to be tested before release.


enoch85:
Tested and works!

PR #149, opened 2017-03-03
Optionally change mirrors

Ark74:
Enable the option to choose mirrors or don't.

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -69,6 +69,7 @@ else
 fi
 
 # Check where the best mirrors are and update
+echo
 echo "Some VPS providers have local download mirrors preconfigured, some don't."
 function ask_yes_or_no() {
     read -p "$1 ([y]es or [N]o): "
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -68,13 +68,31 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


Ark74:
Some VPS providers have faster mirrors that are not publicly available.
So in some cases it would be better to leave them as they are. 

What's your opinion on it?

enoch85:
Thanks for your PR @Ark74! :)

I've also thought about this, but decided not to implement it because it's confusing. Let's say that your totally new to UNIX and you get a question like this. I would always answer "yes I want to keep local mirrors" and then never get the best mirror.

I also noticed that the local mirrors are picked up by apt-select. Digitalocean for instance.

@nextcloud/vm What do you think?

enoch85:
Also, you could get presented with the 3 best mirrors and decide which one you want, or if you want to discard them I think that would actually be a better solution if we should go this path.

Ark74:
> you could get presented with the 3 best mirrors and decide which one you want
It would be equal to always change default mirrors, right?

enoch85:
>What about setting a timer with a pre-configured option for newbies?
Read has -t option to set time for an answer.

Experiment with it and see what you can come up with. :)

> It would be equal to always change default mirrors, right?

Yes, but we could put everything after `else` and make everything optional.

Ark74:
The latest commit, show the default host mirror. That should give a better idea, right?

Issue #144, opened 2017-02-25
Stick to stable kernel (4.4)


Ark74:
I'm testing the the installation script with CODE, i see it installs the latest hwe kernel, which  at this time is linux 4.8.0-34
The kernel 4.8 doesn't have AUFS support, since it's not an upstream stable release.

Ubuntu 16.04 install 4.4 kernel by default, which has aufs support and it is still a LTS kernel release.
This will result on a failure when installing the CODE container, unless the docker storage driver is changed manually.

Cheers!

enoch85:
Hmm, interesting.

@Ark74 Can you do some research on how to keep the latest kernel and still get CODE running?

enoch85:
Just remembered... I run kernel 4.8 on my production server, and I have no issues with CODE (Collabora)

Ark74:
I got the very ambiguous "Access Denied" so i'm trying with a clean installation with aufs precharged (linux-image-extra-virtual linux-image-extra-$(uname -r) ) and the 4.8 kernel installation disable.

I'm testing with a clean vm i'll report.

Ark74:
I found an issue, lest's encrypt won't create the certificate for the office.nextcloud.com domain, the one used for the docker image.
We should add -d $domain to

`./letsencrypt-auto certonly --agree-tos --webroot -w $NCPATH -d $domain`

enoch85:
@Ark74 Yes, it's creating the cert here: https://github.com/nextcloud/vm/blob/master/static/collabora.sh#L251

Ark74:
This is the history line i get every time:

=========================
Start script installation:

Change mirror: N (PR #149  )
Change kblayout: n
Select Apps:
(None)

Reboot 1 

Message Block
Press any key to start the script. Press CTRL+C to abort.

Use VPS : Y
Setup Keyboard: N

Select apps:
* Collabora

Input 
office domain: office.domain.com
nc domain: nc.domain.com

Check port 443: Ok

Run collabora script.
Add extra security: N

Set up Timezone
(UTC)
User system password update
NC password update

Congratulation message block
(Key to continue)

Setup SSL: y
Continue: y
Port 443 open: y
domain available: y

Enter domain: nc.domain.com
is it correct?: Y

input email: my@email.com
agree: A
Share mail: N

System reboot

**New login, scripts ends, no more follow on the CODE docker installation.**

=========================
[...]

PR #150, opened 2017-03-03
Enable SSL subdomain read

Ark74:
This prevent the subdomain ssl cert being called.
Solves https://github.com/nextcloud/vm/issues/144#issuecomment-282528993

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -704,9 +704,15 @@ rm -f *.zip.*
 done;
 
 # Install virtual kernels
-apt install linux-tools-virtual-hwe-16.04-edge linux-cloud-tools-virtual-hwe-16.04-edge  -y
-apt install linux-image-virtual-hwe-16.04-edge -y
-apt install linux-virtual-hwe-16.04-edge -y
+#apt install linux-tools-virtual-hwe-16.04-edge linux-cloud-tools-virtual-hwe-16.04-edge  -y
+#apt install linux-image-virtual-hwe-16.04-edge -y
+#apt install linux-virtual-hwe-16.04-edge -y
+apt install linux-tools-virtual-lts-xenial linux-image-virtual-lts-xenial linux-headers-virtual-lts-xenial \
+linux-image-extra-virtual-lts-xenial -y
+
+# Load aufs 
+modprobe aufs
+echo aufs >> /etc/modules
 
 # Set secure permissions final (./data/.htaccess has wrong permissions otherwise)
 bash $SCRIPTS/setup_secure_permissions_nextcloud.sh
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -704,14 +704,9 @@ rm -f *.zip.*
 done;
 
 # Install virtual kernels
-apt purge -y $(dpkg-query -W -f='${binary:Package}\n' 'linux-image-4*' 'linux-headers-4*' | grep -v 4.4.0-64-generic)
-apt install linux-image-virtual-lts-xenial linux-headers-virtual-lts-xenial -y
-apt install linux-image-extra-virtual-lts-xenial linux-tools-virtual-lts-xenial -y
-
-# Load aufs 
-apt install aufs-tools -y
-modprobe aufs
-echo aufs >> /etc/modules
+apt install linux-tools-virtual-hwe-16.04 linux-cloud-tools-virtual-hwe-16.04  -y
+apt install linux-image-virtual-hwe-16.04 -y
+apt install linux-virtual-hwe-16.04 -y
 
 # Set secure permissions final (./data/.htaccess has wrong permissions otherwise)
 bash $SCRIPTS/setup_secure_permissions_nextcloud.sh
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -704,13 +704,12 @@ rm -f *.zip.*
 done;
 
 # Install virtual kernels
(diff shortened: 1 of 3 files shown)


Ark74:
I'm still unable to get CODE working using the script. Let's do something
I'm still digging. 
Until it runs using the script let's wait to so it can be proved right or wrong.

Ark74:
Nailed!
I'll complete the changes, test once again after that.
If it works, man, have i earned my medal of the month :smiley:

enoch85:
cc @ezraholm50 

Ref: https://github.com/nextcloud/vm/pull/130

enoch85:
This means that we would revert to the 4.4 kernel. I'm not so found of that.

enoch85:
And @Ark74 as I menationed before, this config works for *you* but that doesn't mean it works for everyone. We need more testing here.

Also test without commenting out the 4.8 kernels please.

Ark74:
>We need more testing here.

I think i'll look for pre-existing aufs support, if enabled or available, then it won't touch the kernel, and things like that.
Please test this PR.

> Also test without commenting out the 4.8 kernels please.

I did, that's why i'm on this journey already.

Here the result with this PR.
vm-fresh-build

Please note that after the script stops, you need to manually download and run collabora.sh
Cheers!

enoch85:
After wich scripts stops?

Will test this on a DO VPS tonight.

nextcloud_install_production.sh
```diff
@@ -704,9 +704,14 @@ rm -f *.zip.*
 done;
 
 # Install virtual kernels
-apt install linux-tools-virtual-hwe-16.04-edge linux-cloud-tools-virtual-hwe-16.04-edge  -y
-apt install linux-image-virtual-hwe-16.04-edge -y
-apt install linux-virtual-hwe-16.04-edge -y
+apt purge -y $(dpkg-query -W -f='${binary:Package}\n' 'linux-image-4*' 'linux-headers-4*' | grep -v 4.4.0-64-generic)
```
enoch85: This seems to be the 4.4 kernel, am I right?

enoch85: Also, why grep the exact specific version? What if the version isn't exactly -64?

Ark74: This is part is kind of werid.
Ubuntu already has a 4.4.0-65 but if you install

`linux-image-virtual-lts-xenial linux-headers-virtual-lts-xenial`

 It will install the 4.4.0-64 version, **_which happens to be the one with aufs support_**, the -65 OTOH, doesn't.
I think that i'll add 
`grep aufs /proc/filesystems`

if it returns aufs, then i shouldn't touch the kernels, but if the return is empty it should try to configure the aufs support, by modifying the kernels and stuff, or something like it.

enoch85:
Error:

```
modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/4.4.0-65-generic/modules.dep.bin'
modprobe: FATAL: Module aufs not found in directory /lib/modules/4.4.0-65-generic
```

nextcloud_install_production.sh
enoch85: Yes, it's the 4.4 kernel. Why can't we use the 4.8 kernel? It's a newer kernel and logically it should be even better...

Ark74:
Just what i mention before about the specific 4.4.0-64 kernel required.

nextcloud_install_production.sh
Ark74: Please, do.
I got errors every time i tried.

Ark74: IIRC, Linux 4,8 is like Ubuntu 16.10 as 4.4 like 16.04.
Linux 4.4 is a LTS release.

enoch85: I'm looking into enabling aufs on 4.8 as we speak.

Ark74:
Here the flow I get when installing CODE/Collaboration.
https://github.com/nextcloud/vm/issues/144#issuecomment-283899350
 After second reboot it won't continue the Code/collabora install.

enoch85:
Yeah I saw that, but I don't understand really... It works for me.

nextcloud_install_production.sh
```diff
@@ -704,9 +704,14 @@ rm -f *.zip.*
...
+apt install linux-image-virtual-lts-xenial linux-headers-virtual-lts-xenial -y
```
enoch85: save for later

apt purge -y $(dpkg-query -W -f='${binary:Package}\n' 'linux-image-4*' 'linux-headers-4*' | grep -v 4.4.0-64-generic)

+apt install linux-image-virtual-lts-xenial linux-headers-virtual-lts-xenial -y
 +apt install linux-image-extra-virtual-lts-xenial linux-tools-virtual-lts-xenial -y
 +
 +# Load aufs 
 +apt install aufs-tools -y
 +modprobe aufs
 +echo aufs >> /etc/modules

enoch85:
I fixed so that modprobe works on 4.8 kernel now.

Problem is that the extra kernel packages needs to be installed and run before collabora script is run, so maybe we should to the installation of the kernel in rc.local and restart to get it activated? Hmmm...

So in other words, move the kernel install to rc.local and then restart and THEN run the starup script. That way it will be on the new kernel once collabora is run.

ezraholm50:
Don't receive notifications atm. Working on the server overhaul to ESXI should be participating in about a week again.

enoch85:
@ezraholm50 I tested with and without *:443 and it worked with a *. I merged your commit because I thought you had tested it.

Anyway, this works out of the box now. Great work @Ark74!

PR #153, opened 2017-03-05
Move Let's Encrypt script before installing Apps

enoch85:
@Ark74 Can you test this please?

Also try to do a failed run with SSL and see if it works.


enoch85:
Tested successfully with valid domains.

Now I will test with a domain that is not valid and try to break the script.

Ark74:
Sorry i'm jammed on a pile of work. On sunday! :-1:

enoch85:
Tested and works!

PR #155, opened 2017-03-05
change repo var

enoch85:


enoch85:
cc @Ark74

Ark74:
So does apt-cache policy follows the source.list?
Or does it uses the software properties like add-apt-repository that are not reflected on the sources file?

At the end preferred grepping the sources.list file since apt-select modifies exactly that file.
If you find it better, it's ok with me.

enoch85:
This isn't perfect and doesn't pick up DO for example... :/ But it's better than a empty output which the other gave sometimes becuase it's very specific and expects the 1:st line to contain the mirror address.

enoch85:
Trying to find a failsafe method... but it's hard.

enoch85:
Think I found it!

apt-get update | grep -m 1 Hit | awk '{ print $2}'

Issue #163, opened 2017-03-07
Let's encrypt not working, 443 not open, can't set domain (FQDN)


flotpg:
## Steps to reproduce
1.fresh boot  after unpack of vm image
2.vm gets ip
3.lets encrypt doesn't work, cause public ip isn't detected corrected

## Expected behaviour
want to set FQDN of public domain somewhere to get lets encrypt working

## Actual behaviour
Let's encrypt failes, cause it uses local IP

## Server configuration

**Server version: current** (see your admin page)

**VM version: current**
<img width="651" alt="nextcloud01" src="https://cloud.githubusercontent.com/assets/20798074/23666961/b4f038b6-035c-11e7-8cc9-36961c1d078e.png">
nextcloud says 443 isn't open, it is, i checked with netcat. How does it detect it's ip address? We've multiple WAN IPs.
<img width="1106" alt="nextcloud02" src="https://cloud.githubusercontent.com/assets/20798074/23667008/d7e1c5ba-035c-11e7-8201-2a24fa56d638.png">

Ark74:
It uses nmap to look for it.
I had a similar issue.
What happens if you visit your NC domain?

flotpg:
Hey Ark74,

this is what happens:
<img width="798" alt="netxcloud03" src="https://cloud.githubusercontent.com/assets/20798074/23667421/120f6ed0-035e-11e7-857f-9188b9dbb83f.png">

Ark74:
Ohh! the same issue.
Just ssh your server and add your domain to /var/www/nextcloud/config/config.php

And you'll be good to go :)

Ark74:
@enoch85 
I think this is related to the latest PR #161

Ark74:
I had that behavior too.

enoch85:
Aah ok, we need to update the config parameters based on what the user types as domain.

flotpg:
Ok, working now, but how to get /var/scripts/activate-ssl.sh working

i checked the script,
WANIP4=$(dig +short myip.opendns.com @resolver1.opendns.com)
ADDRESS=$(hostname -I | cut -d ' ' -f 1)

Both result in correct IPs...

enoch85:
Please test this: https://github.com/nextcloud/vm/pull/164

flotpg:
wget also fails, commented them out, now letsencrypt fails:
`Creating virtual environment...
Traceback (most recent call last):
[...]

PR #165, opened 2017-03-09
Rely on distro's LE

Ark74:
Replace git LE with distro LE certbot.

Possible fix for https://github.com/nextcloud/vm/issues/163

### DIFF
--- a/lets-encrypt/activate-ssl.sh
+++ b/lets-encrypt/activate-ssl.sh
@@ -202,9 +202,16 @@ else
 fi
 
 # Install letsencrypt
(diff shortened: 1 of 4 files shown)


enoch85:
Thank you very much @Ark74! Have you tested this against the current master?

@nextcloud/vm Need testers here.

enoch85:
Also check the collabora script as we use SSL there as well.

enoch85:
Please also add a check to check whether lets encrypt is already installed or not.

enoch85:
This installes LE 0.4.1, latest version is 0.12.0...

PR #169, opened 2017-03-10
let the user choose mirror

enoch85:
some mirrors are bad

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -93,7 +93,7 @@ else
   apt install python-pip -y
  pip install \
      --upgrade pip \
-     apt-select
+     apt-select -m up-to-date -t 5 -c
  apt-select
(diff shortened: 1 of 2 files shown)


nextcloud-startup-script.sh
```diff
@@ -93,7 +93,7 @@ else
   apt install python-pip -y
  pip install \
      --upgrade pip \
-     apt-select
+     apt-select -m up-to-date -t 5 -c
```
Ark74: does it clear bad ones?

enoch85: No, it fetches mirrors from a list provided by Ubuntu, But giving the user the option to choose at least minimizes the risk of getting a bad mirror. And with "bad" I mean some packages aren't found.

enoch85: This is the error sort of: http://askubuntu.com/questions/838981/local-apt-mirror-says-release-cant-be-found-but-its-there-what-am-i-missing

PR #170, opened 2017-03-12
try with webroot first

enoch85:
One of the benefits with the webroot method is that Apache wont have to stop before renewing the cert. This also makes it less potent to failures as Apache may not start when it's once stopped which means unwanted downtime. This needs to be fixed.

### DIFF
--- a/lets-encrypt/activate-ssl.sh
+++ b/lets-encrypt/activate-ssl.sh
@@ -280,7 +280,28 @@ SSL_CREATE
 fi
 
 ##### START FIRST TRY
+letsencrypt certonly \
+--webroot -w $NCPATH \
+--rsa-key-size 4096 \
+--renew-by-default \
+--agree-tos \
+-d $domain
 
+# Check if $certfiles exists
+if [ -d "$certfiles" ]
+then
+    # Activate new config
+    bash /var/scripts/test-new-config.sh $domain.conf
+    exit 0
+else
+    echo -e "\e[96m"
+    echo -e "It seems like no certs were generated, we do three more try."
+    echo -e "\e[32m"
+    read -p "Press any key to continue... " -n1 -s
+    echo -e "\e[0m"
+fi
+
+##### START SECOND TRY
 # Stop Apache to aviod port conflicts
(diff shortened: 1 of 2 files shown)


enoch85:
It fails with `unathorized`. Think we need to put something in the Apache conf file.

enoch85:
https://wiki.archlinux.org/index.php/Let%E2%80%99s_Encrypt

enoch85:
Seems like 000-default needs to be the NC root.

@Ark74 Like I wrote in the first post. With this we won't have to shut down Apache on every renewal, which is a good thing.

enoch85:
With only the previous domain conf activated  I get this when requesting a new one:

```
IMPORTANT NOTES:
 - The following errors were reported by the server:

   Domain: c12.domain.nu
   Type:   connection
   Detail: Could not connect to c10.domain.nu.well-known

   To fix these errors, please make sure that your domain name was
   entered correctly and the DNS A record(s) for that domain
   contain(s) the right IP address. Additionally, please check that
   your computer has a publicly routable IP address and that no
   firewalls are preventing the server from communicating with the
   client. If you're using the webroot plugin, you should also verify
   that you are serving files from the webroot path you provided.
```

enoch85:
Help wanted. **Goal**: Request certs with the webroot method.

Ark74:
I believe the standalone is better, since gives the product we need.
Webroot will be messy on every cerbot change and intrusive with configuration, also as you said ports.

Since more and applications will use LE/certbot, with the LE change the new tendency will be to use SSL/Secure connections everywhere.

I believe the best is to let LE identify the server and we should take it from there, and apache restart is expected when installing web platforms, and now more and more, web applications.
My two cents. I'm still trapped with a pile of non paying work :(, but i'll try to review the issue.

enoch85:
Thanks @Ark74!

enoch85:
Either way, we need to change the letsencryptrenew script as it uses and old standard. But let's decide on this first.

enoch85:
I've been thinking. Let's stick to the current config.

If anyone wants to work on this, feel free to do so. Closing for now.

Issue #177, opened 2017-03-20
Install process block when installing solr


BobSquarePants:
## Steps to reproduce
1. install VM with collabora, nextant
2.
3.

## Expected behaviour
no blocking during install process

## Actual behaviour
Stay block on that error:
[![Nextcloud_Community_VM_PRODUCTION-2017-03-20-06-.png](https://s28.postimg.org/vq9c5umyl/Nextcloud_Community_VM_PRODUCTION_2017_03_20_06.png)](https://postimg.org/image/sjesm82ih/)

## Server configuration

**Server version:** (see your admin page)
https://cloud.techandme.se/index.php/s/lTjoMr5smSYHIiM

**VM version** (master / released version)

**Updated from an older installed version or fresh install:**
Fresh

## Network
**Do you use DHCP?**
yes

**Is port 80 and/or 443 open?**
443 yes
80 was not

#### Screenshoots
[![Nextcloud_Community_VM_PRODUCTION-2017-03-20-06-.png](https://s23.postimg.org/w9rf9cp6z/Nextcloud_Community_VM_PRODUCTION_2017_03_20_06.png)](https://postimg.org/image/mcgegahl3/)

Ark74:
Did you complete the installation  and startup script?
What are you running the script in? 
VM, VPS?

I've just tested and Solr is installing and working fine.
Please try again, if you see any error during the installation process report it.

enoch85:
@Ark74 I've seen this behavior as well, don't know what's causing it tbh.

Ark74:
It happens only when installing both collabora & nextant.
If you only install Nextant it doesn't show up.

enoch85:
This is now fixed.

BobSquarePants:
Did you have already an idea when the new release will be out with this fix ?
Thx for the quick and good job

enoch85:
@BobSquarePants Easy, just download the latest script or extract the VM and run it again. It's in master and automatically downloaded.

PR #178, opened 2017-03-20
Update nextant.sh

enoch85:
Possible fix for https://github.com/nextcloud/vm/issues/177

### DIFF
--- a/static/nextant.sh
+++ b/static/nextant.sh
@@ -71,8 +71,7 @@ iptables -A INPUT -p tcp --dport 8983 -j DROP
 #sudo service iptables-persistent start
 #sudo iptables-save > /etc/iptables.conf
 
-service solr start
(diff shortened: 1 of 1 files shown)


enoch85:
ping @Ark74

Ark74:
I found that only when running nextant along with code installers, the issue was presented.
Should i download (this) nextant.sh on /var/scripts before starting the installation script?

enoch85:
> I found that only when running nextant along with code installers

What do you mean?

> Should i download (this) nextant.sh on /var/scripts before starting the installation script?

Yes please.

Also, would be great if you could pin point what the issue is here. I don't have so much time I'm afraid. Thanks Luis!

enoch85:
@Ark74 You rock! Thanks for testing!

PR #183, opened 2017-04-01
possible fix for servername

enoch85:


enoch85:
let's do this in the new code quality PR instead.

PR #184, opened 2017-04-02
improve SSL Labs report

enoch85:


enoch85:
Let's do this in the code quailty PR instead

PR #194, opened 2017-04-09
change method of trying to get cert

enoch85:
maybe move to lib.sh?


morph027:
Hm...as this is the only script which is using this, i would let it there...

enoch85:
OK, all all looks good to you?

morph027:
Jip...

PR #203, opened 2017-04-14
200 check for libsh

morph027:


enoch85:
Should work, not tested. :+1:

PR #216, opened 2017-05-08
not needed

enoch85:


enoch85:
@daita Please don't make changes to the release name any more. It breaks the Nextcloud VM. Thanks!

daita:
My bad ; I switched the release server to my local dev env and cleaned the script. It will stay like this right now :heart:

PR #221, opened 2017-05-17
Remove PW_FILE and put it in .my.cnf instead

enoch85:
Do not merge this until 12! This changes the behavior of where the MySQL pass is saved and breaks a lot of stuff if merged before a new VM is released.


enoch85:
Tested and works!

Issue #223, opened 2017-05-18
nextcloud_install_production.sh: not necessary OS upgrade breakes UTF8 support for samba mounts


daniela-waranie:
First: thanks a lot for this script.

You script requires Ubuntu 16.04 and updates kernel to 4.8.0. 
Why not requiring Ubuntu 16.10. Starting with a "clean" OS vs. OS upgrade is recommended on nearly every forum (as a general advice).

I am asking because with your script installes kernel 4.8.0-51 where nls_utf8.ko (kernel module) is missing in linux-image-extra-4.8.0-51-*.

As far as i can see there is no need to upgrade the OS to run nextcloud.
As far as i can see there is no need have hardware enablement packages within a standard virtual machine - no need to catch up with new hardware devices - no need to add "new hardware" to run nextcloud.
As far as i can see there is no need to install a virtual package optimized for virtual machines - no need to reduce the disc usage (which would only have an effect when removing the old 4.4.0 kernel after installing 4.8.0).

Please remove:

`apt dist-upgrade -y`
and
`apt install linux-tools-virtual-hwe-16.04 linux-cloud-tools-virtual-hwe-16.04  -y
apt install linux-image-virtual-hwe-16.04 -y
apt install linux-virtual-hwe-16.04 -y`

or please install a linux-image-* / linux-image-extra-*  which supports the UTF8 kernel module.
e.g.: linux-image-extra-4.8.0-22-generic

See here:
https://packages.ubuntu.com/search?suite=yakkety&arch=amd64&searchon=contents&keywords=nls_utf8.ko

daniela-waranie:
Background:
I try to run all my application servers as immutable systems. (app update = replacing the server favor to maintaining the server)
In case of nextcloud i mount in the data dir and link the application against a separate database server.

enoch85:
@daniela-waranie Thanks for your input. 

Ubuntu 16.10 is not LTS. 4.8 kernel is needed for Hyper-V to function properly.

There are reasons for my choices. ;)

daniela-waranie:
@enoch85 i am no Hyper-V expert, but Hyper-V support is included since Ubuntu 12.04 and just needs to be enabled:

> There are a few Linux distributions that come with LIS already installed, including Ubuntu – Ubuntu Hyper-V Integration Services have been built into Ubuntu Linux since Ubuntu 12.04 — so there's no need for a separate download and installation. For these distributions, only activation is required to use the full features of Hyper-V

[...]

PR #224, opened 2017-05-21
don't install 4.8 kernel (breaks VMXNET 3)

enoch85:
Also pleases the wishes in https://github.com/nextcloud/vm/issues/223


enoch85:
Tested and works!

PR #249, opened 2017-05-30
Onlyoffice WIP

enoch85:
https://github.com/ONLYOFFICE/Docker-DocumentServer

### DIFF
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -204,8 +204,8 @@ then
   # keep the host
   ProxyPreserveHost On
 
-  ProxyPass / http://127.0.0.1:9981 #(https?)
-  ProxyPassReverse / http://127.0.0.1:9981 #(https?)
+  ProxyPass / https://127.0.0.1:9981
+  ProxyPassReverse / https://127.0.0.1:9981
 </VirtualHost>
 HTTPS_CREATE
 
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -23,7 +23,8 @@ fi
 # Check if Collabora is running
 if [ -d "$NCPATH"/apps/richdocuments ]
 then
-    echo "It seems like Collabora is running, you can't run Collabora at the same time as you run OnlyOffice.
+    echo "It seems like Collabora is running."
+    echo "You can't run Collabora at the same time as you run OnlyOffice."
     exit 1
 fi
 
@@ -211,8 +212,8 @@ RequestHeader set X-Forwarded-Proto "https"
 ProxyRequests off
 
         ProxyPass / http://127.0.0.3:9090/
-        ProxyPassMatch "/(.*)/websocket"  wss://127.0.0.3:9091/$1/websocket
-
+        ProxyPassMatch "/(.*)/websocket"  wss://127.0.0.3:9091/\$1/websocket
+        
         <Location />
                 ProxyPassReverse /
         </Location>
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -253,9 +253,9 @@ then
     service apache2 restart
 # Activate SSL for OnlyOffice Docker (is this needed?)
     mkdir -p /app/onlyoffice/CommunityServer/data/certs
-    cp -f "$CERTFILES/$SUBDOMAIN/privkey.pem" /app/onlyoffice/CommunityServer/data/certs/
-    cp -f "$CERTFILES/$SUBDOMAIN/cert.pem" /app/onlyoffice/CommunityServer/data/certs/
-    cp -f "$CERTFILES/$SUBDOMAIN/chain.pem" /app/onlyoffice/CommunityServer/data/certs/
+    cp -f "$CERTFILES/$SUBDOMAIN/privkey.pem" /app/onlyoffice/CommunityServer/data/certs/onlyoffice.key
+    cp -f "$CERTFILES/$SUBDOMAIN/cert.pem" /app/onlyoffice/CommunityServer/data/certs/onlyoffice.crt
+#    cp -f "$CERTFILES/$SUBDOMAIN/chain.pem" /app/onlyoffice/CommunityServer/data/certs/
     cp -f "$DHPARAMS" /app/onlyoffice/CommunityServer/data/certs/
 # Install Onlyoffice App
     cd $NCPATH/apps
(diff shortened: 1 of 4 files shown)


enoch85:
Seems like this needs a second domain after all... Maybe revert to Collabora settings with a SUBDOMAIN and then reverse proxy the whole thing as in the Collabora setup?

oparoz:
You need a separate domain, a new docker and you can't' enable it at the same time as Collabora, so would need to bring the other docker down when switching.

enoch85:
@oparoz So basically the same setup as Collabora?

oparoz:
Yes

enoch85:
@oparoz Any idea how the Apache conf file should look like?

brantje:
Would a reverse proxy to a custom port solve the problem?

oparoz:
@brantje - Which problem?

@enoch85 AFAIK, their standard config from the readme works?

brantje:
This one: 
```
You need a separate domain, a new docker and you can't' enable it at the same time as Collabora, so would need to bring the other docker down when switching.
```
You can put OnlyOffice on port 8181 for example. With a nginx / apache reverse proxy you would be able to serve it via http / https.

oparoz:
Ah, that's not the issue.
The problem is that you can only have one viewer per media type, so both apps cannot co-exist.

brantje:
Seems like Nextcloud needs an "Open with..." dialog.

oparoz:
Yes, it should work like on Android

enoch85:
> The problem is that you can only have one viewer per media type, so both apps cannot co-exist.

That's not the issue here. The issue is to get it up and running at all. For Collabora we use the "standard" reverse proxy in Apache as you can see in the config. Question is, how should the proxying look like?

**There is currently a bounty of $50 for the one that solves this PR. Anyone interested?**

oparoz:
OK, I'll give you a proxy. But I don't understand what the problem is with the instructions they give.

oparoz:
Docker
```
$ sudo docker run -i -t -d -p 0.0.0.0:3080:80 --restart=always onlyoffice/documentserver
```
Vhost
```
<VirtualHost *:80>
#        ServerAdmin admin@example.com
        ServerName oo.mynextcloud.com
        ServerAlias onlyoffice

        ProxyPass / http://127.0.0.1:3080
        ProxyPassReverse / http://127.0.0.1:3080
</VirtualHost>
```

enoch85:
@oparoz Great thanks, and for SSL?

oparoz:
You don't need SSL, but you could just use 443 for the vhost and add the standard  SSL section, copied from Collabora. You'll need a different cert though...

enoch85:
@oparoz Can you test and confirm that it works and if not, change it until it works and I'll send you $50.

enoch85:
@oparoz Tried with both plain 80 and also 443, get this error:
screenshot from 2017-06-01 17-25-07

**Apache error log**

```
[Thu Jun 01 15:24:58.178678 2017] [proxy:error] [pid 10850] [client 109.124.180.118:51037] AH00898: Error reading from remote server returned by /
[Thu Jun 01 15:24:58.411276 2017] [proxy:error] [pid 10850] [client 109.124.180.118:51037] AH00898: DNS lookup failure for: 127.0.0.1:9981favicon.ico returned by /favicon.ico, referer: http://oo.domain.com
```

enoch85:
If I add /welcome to ProxyPass I get to index, but it still doesn't work.

enoch85:
Current docker:
`61fce9032a7a        onlyoffice/documentserver   "/bin/sh -c 'bash ..."   4 minutes ago       Up 4 minutes                  80/tcp, 0.0.0.0:9981->443/tcp   affectionate_hugle`

oparoz:
You're passing an http request to an https backend. You have to pick one or the other.
Since you went through the effort of configuring the docker to support SSL certificates, you should update the vhost.

enoch85:
@oparoz 

Current vhost:
```
<VirtualHost *:443>
  ServerName oo.danielhansson.nu:443

  <Directory /var/www>
  Options -Indexes
  </Directory>

  # SSL configuration, you may want to take the easy route instead and use Lets Encrypt!
  SSLEngine on
  SSLCertificateChainFile /etc/letsencrypt/live/oo.danielhansson.nu/chain.pem
  SSLCertificateFile /etc/letsencrypt/live/oo.danielhansson.nu/cert.pem
  SSLCertificateKeyFile /etc/letsencrypt/live/oo.danielhansson.nu/privkey.pem
  SSLOpenSSLConfCmd DHParameters /etc/letsencrypt/live/oo.danielhansson.nu/dhparam.pem
  SSLProtocol             all -SSLv2 -SSLv3
  SSLCipherSuite ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-$
  SSLHonorCipherOrder     on
  SSLCompression off

  # Encoded slashes need to be allowed
  AllowEncodedSlashes NoDecode

  # Container uses a unique non-signed certificate
  SSLProxyEngine On
  SSLProxyVerify None
  SSLProxyCheckPeerCN Off
  SSLProxyCheckPeerName Off

  # keep the host
  ProxyPreserveHost On

  ProxyPass / https://127.0.0.1:9981
  ProxyPassReverse / https://127.0.0.1:9981
</VirtualHost>
```

Current docker
```
CONTAINER ID        IMAGE                       COMMAND                  CREATED             STATUS              PORTS                             NAMES
9931a228044b        onlyoffice/documentserver   "/bin/sh -c 'bash ..."   5 minutes ago       Up 5 minutes        80/tcp, 127.0.0.1:9981->443/tcp   fervent_joliot
```

Result: https://oo.danielhansson.nu

enoch85:
Solved it: https://www.bdjl.de/localhost/?tag=onlyoffice

apps/onlyoffice.sh
```diff
@@ -0,0 +1,294 @@
...
+SSLProxyCheckPeerCN Off
+SSLProxyCheckPeerName Off
+
+# contra mixed content warnings
+RequestHeader set X-Forwarded-Proto "https"
+
+# basic proxy settings
+ProxyRequests off
+
+        ProxyPass / http://127.0.0.3:9090/
```
brantje: You use http and wss (websocket secure like https but then for sockets), you you want this?

enoch85: @brantje If I change to https it fails. Any idea?

enoch85: `ONLYOFFICE not reached. Please contact admin`

enoch85:
**Tested and works!**

Let's improve this along the way. HTTPS all the way would be preferred.

PR #252, opened 2017-06-05
Nmap retention model

toby00001:
-Edited script to retain nmap IF previously found on system

### DIFF
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -75,8 +75,6 @@ fi
 if [ "$(dpkg-query -s nmap 2> /dev/null | grep -c "ok installed")" == "1" ]
 then
     NMAPSTATUS=preinstalled
-else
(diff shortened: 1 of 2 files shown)


apps/onlyoffice.sh
```diff
@@ -70,29 +70,55 @@ else
...
 fi
 
-# Check if 443 is open using nmap, if not notify the user
+# Check to see if user already has nmap installed on their system
+
+if [ "$(dpkg-query -s nmap 2> /dev/null | grep -c "ok installed")" == "1" ]
+then
+    NMAPSTATUS=preinstalled
+else
+    NMAPSTATUS=notinstalled
```
enoch85: you don't need `notinstalled` as that variable isn't used. So you can remove the `else`. :)

enoch85:
@toby00001 Thanks for you contribution. Can you please add the same fix for OnlyOffice?

PR #253, opened 2017-06-07
delete pre-defined email

enoch85:
cc @LukasReschke 

Account is also deleted, will inform all users in my contacts (~2800 of them)


enoch85:
cc @jospoortvliet as well.

Issue #268, opened 2017-06-13
test-new-config.sh doesn't set /etc/hosts information, only /etc/hostname


ChuckNozzle:
I have 3 brand new NC12 VM's running and all three have SSL certificates obtained using the activate-ssl.sh script. I have noticed that the /etc/hosts file has not been populated with the FQDN  and hostname, and the hostname in /etc/hostname is set to what is basically the FQDN, which is that way because test-new-config.sh writes that.

I am very new at this, but this doesn't seem correct to me?

I am looking into this as I am having possibly unrelated DNS issues which are causing intermittent connection issues, and this just stood out as it wasn't what I expected to see, so I want to know if this is OK, and what, if any, effect this may have on connectivity.

## Expected behaviour
I expect /etc/hosts to be populated as follows:
127.0.0.1 localhost
127.0.1.1 hostname.domain.com hostname

I also expect /etc/hostname to contain just the hostname.

## Actual behaviour
All three VM's have the following in /etc/hosts:
127.0.0.1 localhost
127.0.1.1 nextcloud

/etc/hostnames contains hostname.domain.com.

## Server configuration

**OS: Ubuntu 16.04.1**

**Server version: 12.0.0**

**VM version: Production Release**

**Fresh install**

## Network
**Do you use DHCP?: Yes. However static IP is set in /etc/network/interfaces.**

**Is port 80 and/or 443 open?: Yes. 443.**

enoch85:
Does it not work as expected if you don't change the /etc/hosts file?

ChuckNozzle:
I haven't changed the /etc/hosts file yet, as I like to understand the implications of the changes before I make them.

In general, it works fine if I connect via IP address on my LAN(as it should), but I get badly intermittent connections when I use the domain names to connect, either on my LAN or from outside.

I then started looking into any settings on my server, or on the VM's that have any relation to DNS, and when these settings didn't look like what I was expecting to see, I thought I should query it to see if they were correct, or if they could be contributing to my connection issues.

Having said that, the connectivity issues could also be related to either a problem with the CNAME's that I am currently using as alias's for the DDNS to keep the domains pointed at the VM's, or a DNS caching/propagation issue, so I am trying to rule things out in my search for a solution.

If you feel that this shouldn't be contributing to DNS connectivity issues in a material way, then I am happy for now to continue looking into the ISP and domain hosting side of things.

enoch85:
Thanks for your explanation. 

[...]

PR #269, opened 2017-06-13
Set hostname in /etc/hosts

enoch85:
Possible fix for https://github.com/nextcloud/vm/issues/268


enoch85:
This is better either way, tested and works. 

(For the future it could be "nextcloud" instead of `hostname -s`)

PR #271, opened 2017-06-15
set domain in /etc/hosts

enoch85:


enoch85:
This should solve problems when using DDNS.

PR #277, opened 2017-06-22
[WIP] Change to PostreSQL

enoch85:
Signed-off-by: Daniel Hansson <daniel@techandme.se>

PostgreSQL is safe and faster + no need for utf8mb4 which is something MySQL made up because they failed.

Much work to be done here.


MariusBluem:
Really? Nextcloud is recommending MariaDB/MySQL on docs.nextcloud.com for the setups ... it is much more used - and also more tested then 🙈

Ark74:
Maybe adding it as an option, could be a better idea.
But change it as a default i don't believe that's a really great move.

Or maybe make a postgres branch for "high performance" or whatever.

enoch85:
This is an experimental branch, I'll not merge to master. I'm thinking maybe offering another VM in the Tech and Me Shop. We'll see.

From an enterprise perspective this is much better. Not as much settings as Postgres have it all built in already so no need for tweaks. Not like MySQL or MariaDB where you have to **add** support for utf8bm4 and stuff like that.

Also, like I said, Postgres is **MUCH** faster. Did some tests and wow, it flies!

Ark74:
In deed it does, some other high intensive platforms use Postgres as database for the same reason, such as Odoo.
But mysql is much better integrated on the current web services environment.
Nevertheless the point about postgres remains.

enoch85:
Though I am really tempted to change to PostgreSQL...

@MorrisJobke @nickvergessen @LukasReschke @juliushaertl @rullzer and others, how much will break (if any) if I change to PotgreSQL? Are many apps depending on MySQL?

MorrisJobke:
> @MorrisJobke @nickvergessen @LukasReschke @juliushaertl @rullzer and others, how much will break (if any) if I change to PotgreSQL? Are many apps depending on MySQL?

Nothing™ ... it runs on multiple sites with Postgres. We recommend MySQL because we had good experiences with it especially in clustered environments.

enoch85:
@MorrisJobke 

> because we had good experiences with it especially in clustered environments.

So what you're saying is that Postgres doesn't work as good as MySQL/MariaDB in clustered environments?

Another questsion regarding git, how do I move this PR to be merged to another branch and follow the master branch? As I said I'm thinking of maybe offering this as an alternative to MariaDB which is the default in 12.0.1 (not released yet).

What I'm thinking is:
1. create another branch called PSQL
2. Close this PR
3. Clone branch postgresql
4. `git co --track -b PSQL master`

Would that be enough?

enoch85:
I'll create a new branch fot this. Closing this PR.

PR #280, opened 2017-06-23
Add info about RAM

enoch85:
Let's try again...


enoch85:
Hmm, it works. Ok, never mind @morph027

PR #310, opened 2017-07-23
change command for restarting mariadb

enoch85:


enoch85:
Look at this later.. not satisfied with this fix, but it works.

PR #349, opened 2017-08-25
Enable HTTP/2 server wide

MichaIng:
- In favour of https://help.nextcloud.com/t/http-2-on-apache-2-4-27-debian-9/19885/4?u=michaing
- Since Apache version 2.4.17, "`mod_http2`" is shipped but not enabled by default.
- Even enabled, the HTTP/2 protocol is not used by default by the web server.
- It brings "`/etc/apache2/mods-available/http2.load`" but no "`/etc/apache2/mods-available/http2.conf`", so we can use that to enable/disable the protocol server wide together with the mod, as "`*.load`" and "`*.conf`" are symlinked together into "mods-enabled" and included into "`apache2.conf`".
- "`Protocols h2 h2c http/1.1`" defines the order in which the protocols are prioritize, where "`h2`" stands for HTTP/2 for https/ssl and "`h2c`" for h2**clear**, thus the clear text non ssl/http.
- "`H2Direct on`" enables less overhead on establishing HTTP/2 connections, but might have some compatibility issues. Further discussion/testing about that might be reasonable.
- Further information: https://httpd.apache.org/docs/2.4/mod/mod_http2.html, https://httpd.apache.org/docs/2.4/howto/http2.html

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -411,10 +411,21 @@ SSL_CREATE
     echo "$SSL_CONF was successfully created"
 fi
 
+# Enable HTTP/2 server wide
+touch /etc/apache2/mods-available/http2.conf
+cat << HTTP2_ENABLE > /etc/apache2/mods-available/http2.conf
+<IfModule http2_module>
+        Protocols h2 h2c http/1.1
+        H2Direct on
+</IfModule>
+HTTP2_ENABLE
+echo '"/etc/apache2/mods-available/http2.conf" was successfully created'
+
 # Enable new config
 a2ensite nextcloud_ssl_domain_self_signed.conf
 a2ensite nextcloud_http_domain_self_signed.conf
 a2dissite default-ssl
+a2enmod http2
 service apache2 restart
 
 whiptail --title "Which apps/programs do you want to install?" --checklist --separate-output "" 10 40 3 \
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -412,14 +412,14 @@ SSL_CREATE
 fi
 
 # Enable HTTP/2 server wide
-touch /etc/apache2/mods-available/http2.conf
-cat << HTTP2_ENABLE > /etc/apache2/mods-available/http2.conf
+touch "$HTTP2_CONF"
+cat << HTTP2_ENABLE > "$HTTP2_CONF"
 <IfModule http2_module>
         Protocols h2 h2c http/1.1
         H2Direct on
 </IfModule>
 HTTP2_ENABLE
-echo '"/etc/apache2/mods-available/http2.conf" was successfully created'
+echo "$HTTP2_CONF was successfully created"
 
 # Enable new config
 a2ensite nextcloud_ssl_domain_self_signed.conf
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -411,6 +411,10 @@ SSL_CREATE
     echo "$SSL_CONF was successfully created"
 fi
 
+## Add ask_yes_no here
+
+# if ....
+
 # Enable HTTP/2 server wide
 cat << HTTP2_ENABLE > "$HTTP2_CONF"
 <IfModule http2_module>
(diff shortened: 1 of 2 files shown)


MichaIng:
Shall we create `HTTP2_CONF = "/etc/apache2/mods-available/http2.load"` in [lib.sh](https://github.com/nextcloud/vm/blob/master/lib.sh#L55)

enoch85:
> Shall we create HTTP2_CONF = "/etc/apache2/mods-available/http2.load" in lib.sh?

Yes, please do and change it to a variable instead. Thank you very much! I will test this and get back to you.

nextcloud_install_production.sh
```diff
@@ -412,14 +412,14 @@ SSL_CREATE
 fi
 
 # Enable HTTP/2 server wide
-touch /etc/apache2/mods-available/http2.conf
-cat << HTTP2_ENABLE > /etc/apache2/mods-available/http2.conf
+touch "$HTTP2_CONF"
```
enoch85: `touch` is not needed as `cat` is passed as `>` which creates a new file.

enoch85:
Remember this is Ubuntu 16.04, not Debian 9. Seems there is more to it than just creating a conf and expect everything to work. :/

Maybe we need to add Ondrejs PPA, but I'd rather not. https://www.youtube.com/watch?v=xoJxXYDW2M8

MichaIng:
Hmm, since apache 2.4.17 there should habe nothing changed about the module name. Can you check `/etc/apache2/mods-available/` if the related `*.load` file is there?

All guides about that are the same for Debian and Ubuntu as far as I could see, should work for both. So far the module seems not to be there (perhaps new apache version solves it) or with a different name.

enoch85:
This is what I see. 

screenshot from 2017-08-25 21 18 52

Bummer, seems like they didn't include it in 16.04: https://wiki.ubuntu.com/XenialXerus/ReleaseNotes#HTTP.2F2_support_in_Apache_httpd

So, should we wait until 2018 when 18.04 is out, or just add Ondrejs PPA until then? I mean, how much do we gain with HTTP2, is or much or just nice to have?

MichaIng:
Also just found it: https://www.digitalocean.com/community/questions/enable-http2-in-apache-on-ubuntu-16-04

There seem to be a repo for apache2 with http2 included. But yeah that's a question if one want to rely on such private repos: https://launchpad.net/~ondrej/+archive/ubuntu/apache2

enoch85:
I trust Ondrej, so no problem with his PPA. 

What I'm thinking is the LTS support. I mean, what happens in 1 year when the user decides to upgrade to 18.04? OK if you chose to install it yourself, but if you don't know that PPA is used it could cause problems longer down the road. What do you think?

I just tested the PPA btw, and it works. It also creates a `.load` file.

MichaIng:
Would be interesting to know, if Ubuntu 17.04 also doesn't ship it. Could be reasonable if they decided to explicitly remove it for stability/compatibility.

To be true, I personally did not recognize any performance gain by browsing my small RPi2 single server nextcloud. But indeed I didn't do any benchmark or what.
My motivation was more about tinkering around and trying/advance the newest technology.

@18.04 upgrade: Yeah that's a point. On the other hand, how high is the chance, that 18.04 still does not include http2 yet? Maybe there is some roadmap about it? Could be also found out via 17.04: If this includes the module, 18.04 will most likely also 😄.

Did you test the PR with the PPA?

enoch85:
> Did you test the PR with the PPA? 

Sort of. I ran the lines of this PR and it dodn't work. Then I added the PPA, upgraded Apache and then it worked. So yes, adding the PPA in this PR will work.

You know what? Let's make it an option. So please use the `asl_yes_no` function and ask if the user wants HTTP2 or not. If they want it, install Apache with the PPA, if not, just do a regular install.

Sounds like a good compromise?

MichaIng:
Sounds good. Also that way people can be informed about the PPA and that they in case need to use it again after ubuntu upgrade.

enoch85:
Great! Here's a good example of the funtion: https://github.com/nextcloud/vm/blob/master/nextcloud-startup-script.sh#L86

So the question should be: "Do you want to add HTTP2 and thereby Onrdrejs PPA?"
if no
Do nothing as regular apache is already installed
if yes
add the PPA, then do `apt update -q4 spinner_loading` and `apt upgrade apache2 -y` and then enable the rest as of this PR.

Feel free to add text if you want.

MichaIng:
My text become a bid longer 😆 . Feel free to adjust it. Oh forgot "-y" to the upgrade.

enoch85:
Working on the final touches as we speak.

enoch85:
@MichaIng Thanks for this, really appreciated!

I will test it tomorrow as well and maybe add even more tweaks to it. Feel free to do the same.

enoch85:
Tested and works. Let's merge this!

PR #352, opened 2017-08-28
grep logfile path

enoch85:


enoch85:
cc @nachoparker FYI on how to improve it. Get the path directly from config.php

PR #361, opened 2017-09-07
update snap to latest

johnfelipe:


enoch85:
Great! Have you tested this? What are the benefits? Does it solve any problem?

enoch85:
Ping @johnfelipe 

Using `--edge` could be unstable, so that's why I'm asking if it makes any improvements, or fix any issues?

enoch85:
ping @johnfelipe Can you answer above questions please?

johnfelipe:
Great! Have you tested this? YES
What are the benefits? I can use in wifi and 4g network in my office 
Does it solve any problem? for moment solve all issue with spreedme and coturn

enoch85:
@johnfelipe Did you notice it was unstable?

Also, which specific issues did you have prior to this fix?

johnfelipe:
Did you notice it was unstable? for me wasn´t

Also, which specific issues did you have prior to this fix? is not issues because is working for me in ubuntu 16 own server and digitalocean, suggest change -edge parameter.

enoch85:
Well, I'm not convinced, but OK. We can always revert.

PR #369, opened 2017-09-21
DHPARAMS 4096 to allow for IE11/Win7 to work instead of Page Cant Be …

gabasync:
…Displayed

ssllabs.com shows a standard IE11/Win7 setup will not work. For companies that do not allow alternative popular Browsers this can be a non go for internal company deployment. Lowering the DH to 4096 may alleviate this issue. Below is the ssllabs.com lab results:

IE 11 / Win 7  R		Client does not support DH parameters > 4096 bits
RSA 4096 (SHA256)   |  TLS 1.2  |  TLS_DHE_RSA_WITH_AES_128_GCM_SHA256  |  DH 8192

### DIFF
--- a/lets-encrypt/activate-ssl.sh
+++ b/lets-encrypt/activate-ssl.sh
@@ -248,7 +248,7 @@ if [ -d "$CERTFILES" ]
     # Generate DHparams chifer
     if [ ! -f "$DHPARAMS" ]
     then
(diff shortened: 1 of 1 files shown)


enoch85:
Yup, this is good. After reading the LE forum I think this is the correct way to go. Thank you for your PR! Could you please do the same in the PosgreSQL branch?

PR #376, opened 2017-09-24
make check open port a function

enoch85:
And include the Nmap check


enoch85:
Needs testing.

- Will LE work (since that extra step on $domain).
- Does $1 and $2 work everywhere?

enoch85:
Tested and works. 

Not tested when $1 is failing though and $2 is needed..

PR #391, opened 2017-10-04
Update onlyoffice.sh

steve-s-71:
Modified order of SSL Cipher Suites to improve connection security with certain browsers, e.g. Firefox (ECDHE-RSA-AES256-GCM-SHA384 instead of -SHA256)


enoch85:
Congratulations to your first PR and welcome as a contributor to Nextcloud!

I will check this later, but seems good. Thanks!

enoch85:
We will fix this together instead. :)

Issue #396, opened 2017-10-04
Show text in whiptail msgbox instead of plain text


enoch85:
Make msgbox a function.

```

. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)

msg_box() {
local PROMPT="$1"
    whiptail --msgbox "${PROMPT}" "$WT_HEIGHT" "$WT_WIDTH"
}

msg_box "This is a whiptail msgbox :)"
```

Results in
![screenshot from 2017-10-04 23 01 34](https://user-images.githubusercontent.com/4511254/31199557-13044540-a958-11e7-8674-db6307ac54fa.png)

PR #406, opened 2017-10-31
fix whiptail msgbox instead of boring text

enoch85:
Fixes https://github.com/nextcloud/vm/issues/396

**Left to do:**
- [ ] Replace `echo` with `msg_box` in all scripts


enoch85:
Let's start with the most used scripts and add more scripts as time goes.

PR #407, opened 2017-10-31
fix whiptail msgbox instead of boring text [ postgresql ]

enoch85:


enoch85:
Let's start with the most used scripts and add more scripts as time goes.

Issue #431, opened 2017-12-17
Script Full Text Search


enoch85:
Script installation of Full Text Search as requested here: https://github.com/daita/fullnextsearch/issues/1 

https://github.com/nextcloud/fulltextsearch/wiki/Basic-Installation

**UPDATE**
Script is here: https://github.com/nextcloud/vm/blob/improvments/apps/fulltextsearch.sh

enoch85:
Install Elastic Search:

```
check_command apt-get install openjdk-8-jre
check_command curl -L -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.1.2.deb
check_command dpkg -i elasticsearch-6.1.2.deb
check_command /etc/init.d/elasticsearch start

# Install plugin
if [ -d /usr/share/elasticsearch ]
then
    cd /usr/share/elasticsearch
fi
bin/elasticsearch-plugin install ingest-attachment
```

enoch85:
https://gist.github.com/ondrej-kvasnovsky/9363975 (Install steps)

enoch85:
@Ark74 Would you mind do some researching, or maybe even try to make a script our of this? I'm happy to assist.

Ark74:
I've been wanting to work this out, but haven't got the time yet.
I was thinking about starting on removing Solr so old instances could implement this new engine too.
Next week i'll make time to take a look on this.
Cheers!

BetaLeaf:
Can't wait! 

Also, make sure you make a script to uninstall nextant & solr. I already know how, but other end users may not know how.

enoch85:
I've started a PR here: https://github.com/nextcloud/vm/pull/446

Ark74:
Wow!  Looks like you nailed. Great work.

I'll help on the testing then. 
Cheers!

enoch85:
@Ark74 There are still a few things left until perfection. We need to secure it with SSL, and also check the SHA against the package that is in ../fulltextsearch-files. We also need a proper YML file file passwords and username that needs to be put in the GUI automatically: https://github.com/nextcloud/vm/pull/446#issuecomment-359592078

Can you do the finishing touches?

enoch85:
@Ark74 Can you please test this and add the finishing details before 13.0.0 is released? Would be awesome! It's live in this branch: https://github.com/nextcloud/vm/blob/nc13-appinstall/apps/fulltextsearch.sh

Ark74:
[...]

PR #446, opened 2018-01-20
WIP Full Text Search

enoch85:
Fixes https://github.com/nextcloud/vm/issues/431


enoch85:
It's now in a working state, but much to be fixed until it's final. Set values directly in the app. Maybe @daita can help out with which values that have to be set in both GUI and elasticsearch.yml?

enoch85:
Merged to nc13-appinstll already. Closing this one.

PR #475, opened 2018-02-21
prevents ssl error

Ark74:
By updating ssl cert, this error is prevented in case previous java configuration was performed on the same server installing ES.


enoch85:
Hmm, that's done automatically for me, but OK. Better safe than sorry.

PR #476, opened 2018-02-21
Fix variables on occ command

Ark74:
This should fix the static values for the variables on the occ command.
Cheers!

### DIFF
--- a/apps/fulltextsearch.sh
+++ b/apps/fulltextsearch.sh
@@ -149,8 +149,9 @@ chown -R www-data:www-data $NC_APPS_PATH
 
 # Final setup
 # Add password and user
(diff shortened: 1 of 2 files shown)


enoch85:
Thanks, but please but all the varbles in lib.sh.

Ark74:
Well, the command is inside hard quotes, that's why I tried this way.
I'll think on something.

Please any other idea is pretty welcome. :)

enoch85:
A much cleaner solution :)

enoch85:
Will have to test the quoting later.

enoch85:
Fixed quoting. Everything expands correctly.

PR #525, opened 2018-05-05
changed ncdata volume to ZFS instead of BTRFS

penzoiders:
removed maintenance part since not needed on ZFS, eventually checking it later on.
Tested (the single script), ended up with a correctly mounted and persistent across reboot ZFS volume in /mnt/ncdata.
enjoy ZFS.

### DIFF
--- a/static/format-sdb.sh
+++ b/static/format-sdb.sh
@@ -27,6 +27,7 @@ mkdir -p "$MOUNT_"
 isMounted() { findmnt -rno SOURCE,TARGET "$1" >/dev/null;} #path or device
 isDevMounted() { findmnt -rno SOURCE        "$1" >/dev/null;} #device only
 isPathMounted() { findmnt -rno        TARGET "$1" >/dev/null;} #path   only
+isDevPartOfZFS() { zpool status | grep "$1" >/dev/null;} #device memeber of a zpool
 
 if isPathMounted "/mnt/ncdata";      #Spaces in path names are ok.
 then
@@ -52,6 +53,13 @@ then
 msg_box "/dev/sdb1 is mounted and need to be unmounted before you can run this script."
 exit 1
 fi
+
+if isDevPartOfZFS "sdb";
+then
+echo "/dev/sdb is a member of a ZFS pool and needs to be removed from any zpool before you can run this script."
+exit 1
+fi
+
 
 # Get the name of the drive
 SDB=$(fdisk -l | grep sdb | awk '{print $2}' | cut -d ":" -f1 | head -1)
--- a/static/format-sdb.sh
+++ b/static/format-sdb.sh
@@ -56,7 +56,7 @@ fi
 
 if isDevPartOfZFS "sdb";
 then
-echo "/dev/sdb is a member of a ZFS pool and needs to be removed from any zpool before you can run this script."
+msg_box "/dev/sdb is a member of a ZFS pool and needs to be removed from any zpool before you can run this script."
 exit 1
 fi
 
--- a/static/format-sdb.sh
+++ b/static/format-sdb.sh
@@ -10,6 +10,9 @@ true
 # Check if root
 root_check
 
+# Check if ZFS utils are installed
+install_if_not zfsutils-linux 
+
 LABEL_=ncdata
 MOUNT_=/mnt/$LABEL_
 format() {
@@ -70,10 +73,18 @@ msg_box "Formatting $SDB when you hit OK.
 
 *** WARNING: ALL YOUR DATA WILL BE ERASED! ***"
     check_command wipefs -a -f "$SDB"
-    check_command parted "$SDB" --script -- mklabel gpt
-    check_command parted "$SDB" --script -- mkpart primary 0% 100%
     sleep 0.5
-    check_command mkfs.btrfs -q "$SDB"1 -f -L "$LABEL_"
+    check_command zpool create -f -o ashift=12 "$LABEL_" "$SDB"
+    check_command zpool set failmode=continue "$LABEL_"
+    check_command zfs set mountpoint="$MOUNT_" "$LABEL_"
+    check_command zfs set compression=lz4 "$LABEL_"
+    check_command zfs set sync=disabled "$LABEL_"
+    check_command zfs set xattr=sa "$LABEL_"
+    check_command zfs set primarycache=all "$LABEL_"
+    check_command zfs set atime=off "$LABEL_"
+    check_command zfs set recordsize=128k "$LABEL_"
+    check_command zfs set logbias=latency "$LABEL_"
+
 else
 msg_box "It seems like /dev/sdb does not exist.
 This script requires that you mount a second drive to hold the data.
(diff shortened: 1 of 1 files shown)


static/format-sdb.sh
```diff
@@ -93,43 +104,12 @@ then
     sed -i 10q /etc/fstab > /dev/null
 fi
 
-# Mount it in fstab
-UUID=$(blkid /dev/sdb1 | awk '{ print $3 }')
-FSTAB="$UUID     $MOUNT_     btrfs   defaults 0       2"
-echo "# ncdata mount" >> /etc/fstab
-echo "$FSTAB" >> /etc/fstab
-check_command mount -a
-
```
enoch85: It seems to me that it doesn't stay persistent across reboots if you don't put it in fstab as well.

Run the script, reboot, and check `df -h`.

static/format-sdb.sh
```diff
@@ -70,10 +73,18 @@ msg_box "Formatting $SDB when you hit OK.
 
 *** WARNING: ALL YOUR DATA WILL BE ERASED! ***"
     check_command wipefs -a -f "$SDB"
```
enoch85: In the previous script I could run it as many times as I wanted. It then reformatted the drive. Could you add that here as well?

enoch85:
@penzoiders Thanks a lot!

Could you please take a look at the comments?

static/format-sdb.sh
enoch85: Sorry, my bad, missed the last line. Please ignore this.

penzoiders:
Ok, will take a look in the afternoon.
Will do a forceful wipe that'll work with the zfs volume.
I'll check that zfs stays mounted across reboots (I had tested it without specific mount point.. I'll double check)

static/format-sdb.sh
enoch85: Just noticed that the "isMounted" functions doesn't work so it doesn't recognize that the drive is mounted on a second run.

static/format-sdb.sh
penzoiders: Ok, I was pretty sure that was working. Nevermind

static/format-sdb.sh
penzoiders: Yes, in fact I had to modify that 
```
-if grep "$UUID" /etc/fstab
+if grep "$LABEL_" /etc/mtab
```

enoch85: sudo zfs list -pH -o mountpoint | grep -q '^/mnt/ncdata$' && echo yes

enoch85:
Added some more comments, I want it to be as failsafe as possible, so please work with more tests and checks. Thanks a lot!

penzoiders:
Since you want it failsafe I noticed that it's better to use `sync=enabled` (I'm used to run VMS on a very reliable and rollback capable distributed file system, but that's not the default for everyone, yet).

static/format-sdb.sh
```diff
@@ -70,15 +73,23 @@ msg_box "Formatting $SDB when you hit OK.
...
-    check_command parted "$SDB" --script -- mklabel gpt
-    check_command parted "$SDB" --script -- mkpart primary 0% 100%
     sleep 0.5
-    check_command mkfs.btrfs -q "$SDB"1 -f -L "$LABEL_"
+    check_command zpool create -f -o ashift=12 "$LABEL_" "$SDB"
+    check_command zpool set failmode=continue "$LABEL_"
+    check_command zfs set mountpoint="$MOUNT_" "$LABEL_"
+    check_command zfs set compression=lz4 "$LABEL_"
+    check_command zfs set sync=enabled "$LABEL_"
```
penzoiders: sync changed to enabled to be more safe on not highly reliable hosts.

static/format-sdb.sh
penzoiders: tested again on a new machine. persistent mount is working.
selezione_999 887

penzoiders:
I think we should also keep it as simple as possible.
I had just turned this script from btrfs to zfs.. now I'm going to fix the wiping part.

do you want me to le the user choose among free and not mounted drives?
or choose among everything (except / and /boot) and eventually savagely wipe and zfs it?

enoch85:
> do you want me to let the user choose among free and not mounted drives?

If only `/dev/sdb` exists, then continue without prompt, else ask for input, but exclude `/dev/sda´.

static/format-sdb.sh
enoch85: Please check the isMounted functions as well. We need to be able to check if `/dev/sbx*` is mounted or `/mnt/ncdata` before we proceed with anything. Right now those functions are broken due to a different way of mounting the volume. 

That's why I posted the command.

enoch85: - `/dev/sda` ofc

enoch85:
We could postpone release until a later date if needed.

static/format-sdb.sh
```diff
@@ -72,6 +72,10 @@ then
 msg_box "Formatting $SDB when you hit OK.
 
 *** WARNING: ALL YOUR DATA WILL BE ERASED! ***"
+    if zpool list | grep ncdata > /dev/null
+    then
+        check_command zpool destroy ncdata
+    fi
```
enoch85: Please `grep $LABEL_` instead, if possible. Thanks!

static/format-sdb.sh
penzoiders: going to check those functions in a while... fixed the ability to correctly re-wipe everything by destroying ncdata zpool if present.

static/format-sdb.sh
enoch85: Thanks! :+1:

enoch85:
OK, so now it feels kind of secure and failsafe. This still needs to be fixed: https://github.com/penzoiders/vm/blob/postgresql/static/format-sdb.sh#L25-L54

enoch85:
Just found a bug:

```
cannot set property for 'ncdata': 'sync' must be one of 'standard | always | disabled'
zfs set sync=enabled ncdata failed
```

enoch85:
Fixed it

enoch85:
@penzoiders Are you working on [this](https://github.com/nextcloud/vm/pull/525#issuecomment-386874221)?

penzoiders:
yes sir, just fixed it...

ups, sorry for the sync parameter discovered now while testing again the whole thing.

it should be ok for now if you want to publish..

I'd love to do a better "find a suitable drive and use that for ncdata" but today I'm out of time.
Will eventually work on it later on (and also provide compatibility for VirtIO devices for people using it with paravirtualized devices in KVM)

static/format-sdb.sh
```diff
@@ -82,7 +82,7 @@ msg_box "Formatting $SDB when you hit OK.
     check_command zpool set failmode=continue "$LABEL_"
     check_command zfs set mountpoint="$MOUNT_" "$LABEL_"
     check_command zfs set compression=lz4 "$LABEL_"
-    check_command zfs set sync=enabled "$LABEL_"
+    check_command zfs set sync=always "$LABEL_"
```
penzoiders: I would have gone with "standard" anyway, always can have severe performance hits

penzoiders:
ups.. very sorry for the "dirty" code with bad intents and the `echo` from my unit tests.
let me know if you need anything else

enoch85:
@penzoiders 

> I'd love to do a better "find a suitable drive and use that for ncdata" but today I'm out of time.
Will eventually work on it later on (and also provide compatibility for VirtIO devices for people using it with paravirtualized devices in KVM)

That would be suuper awesome. I'm really greatful for what you've done here. Thanks!

Just tested and it works i'm a similar way as before. Feels nice to use ZFS instead of BTRFS actually, so thanks for the heads up.

Will do some final changes and then merge.

penzoiders:
I have to thank you for all the work, it was a pleasure to help out.
Will definitely write the part for safe disk selection and VirtIO support.

please consider to use `sync=standard` otherwise you'll get performance issues.

enoch85:
Tested and works!

PR #542, opened 2018-05-17
changed network connectivity test (

penzoiders:
restarting just network-manager failed badly on a VPS with static IP coming from VM contextualization scripts) this works better and brings up routing too.

	modified:   lib.sh

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -202,15 +202,20 @@ done
 test_connection() {
 install_if_not dnsutils
 install_if_not network-manager
-check_command service networking restart
+check_command service network-manager restart
 ip link set "$IFACE" down
 wait
 ip link set "$IFACE" up
 wait
-check_command service networking restart
+check_command service network-manager restart
 echo "Checking connection..."
 sleep 3
 if ! nslookup github.com
+then
+    echo "Trying to restart networking service..."
+    check_command service networking restart
+fi
+if  ! nslookup github.com
 then
 msg_box "Network NOT OK. You must have a working network connection to run this script
 If you think that this is a bug, please report it to https://github.com/nextcloud/vm/issues."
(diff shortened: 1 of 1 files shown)


enoch85:
We use network-manager and not networking, so :-1: 

Maybe if one fails, then test with the other?

enoch85:
And thanks for the PR btw. :)

penzoiders:
what happens on my server is that restarting network-manager only after setting ip link down/up is that the iface is up and ip is correctly set but the default route is not there, so connectivity fails.

maybe it's ok, before giving test=failed, as last resource try a `systemctl networking restart` and see if it gets up.

BTW restarting networking should restart all the network stack including network manager (not 100% sure about it but worth a test)

enoch85:
Tested on Ubuntu 16.04 (which doesn't hold network-manager by default) and it works!

enoch85:
@penzoiders Do you want to be a maintainer for this repo?

enoch85:
And this is why I voted :-1:  https://github.com/nextcloud/vm/issues/547

Can we make it even better with `elif` instead?

penzoiders:
@enoch85 I admit that I don't have an amount of spare time to be a full time maintainer on this, but since I often use this script I'll be glad to help out.
Tell me what you prefer, I can keep on sending PR as I find something, of course if there's a bug on something I push I'll try to be as quick as possible in fixing it.

For #547 I'll suggest to check whether is loaded or not the networking service, and only in that case restart it (of course if connectivity test fails, if successful, just skip it).
--> do you want a PR for that patch?

enoch85:
I have an [ongoing PR](https://github.com/nextcloud/vm/pull/548) that you could help me test. 

Regarding being maintainer, just keep sending PRs and try to help with issues if you have time. :)

Issue #545, opened 2018-05-18
Make TURN (Talk) happen!


enoch85:
https://github.com/nextcloud/talk-android/pull/169

cc @mario

enoch85:
@mario Now my calendar is pretty clear. Will start to look at this next week.

penzoiders:
actually moved now to #562

PR #561, opened 2018-06-07
Install Talk together with TURN behind NAT

enoch85:
Fix https://github.com/nextcloud/vm/issues/545

Take the next step and install a NATed TURN server instead of just installing and enabling the app.

Help are welcome!

### DIFF
--- a/apps/talk.sh
+++ b/apps/talk.sh
@@ -16,17 +16,113 @@ debug_mode
 TALK_CONF="/etc/turnserver.conf"
 
 echo "Installing Talk..."
-check_open_port 443
-check_open_port 80
+check_open_port 443 $TALKDOMAIN
+check_open_port 80 $TALKDOMAIN
+
+# Install Apache2
+install_if_not apache2
+
+# Enable Apache2 module's
+a2enmod proxy
+a2enmod proxy_wstunnel
+a2enmod proxy_http
+a2enmod ssl
+
+# Create Vhost for Talk online in Apache2 (Taken from Collabora, edit values later)
+if [ ! -f "$TALKPROXY_CONF" ];
+then
+    cat << TALKPROXY_CREATE > "$TALKPROXY_CONF"
+<VirtualHost *:443>
+  ServerName $TALKDOMAIN:443
+  
+  <Directory /var/www>
+  Options -Indexes
+  </Directory>
+  # SSL configuration, you may want to take the easy route instead and use Lets Encrypt!
+  SSLEngine on
+  SSLCertificateChainFile $CERTFILES/$TALKSUBDOMAIN/chain.pem
+  SSLCertificateFile $CERTFILES/$TALKSUBDOMAIN/cert.pem
+  SSLCertificateKeyFile $CERTFILES/$TALKSUBDOMAIN/privkey.pem
+  SSLOpenSSLConfCmd DHParameters $DHPARAMS
+  SSLProtocol             all -SSLv2 -SSLv3
+  SSLCipherSuite ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS
+  SSLHonorCipherOrder     on
+  SSLCompression off
+  # Encoded slashes need to be allowed
+  AllowEncodedSlashes NoDecode
+  # Container uses a unique non-signed certificate
+  SSLProxyEngine On
+  SSLProxyVerify None
+  SSLProxyCheckPeerCN Off
+  SSLProxyCheckPeerName Off
+  # keep the host
+  ProxyPreserveHost On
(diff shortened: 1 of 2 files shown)


enoch85:
cc @mario 

Does this have to be on a separate subdomain for it to work with NAT or would it be OK to reuse all the settings (TLS and such) from the Nextcloud TLS setup?

mario:
Well, I was telling you to have a separate subdomain which would then proxy to turn/stun. Is that not doable/optimal from your perspective?

No idea about proxy to be honest. We can try and see? :D

apps/talk.sh
```diff
@@ -1,13 +1,36 @@
+#!/bin/bash
+
+# Tech and Me © - 2017, https://www.techandme.se/
```
mario: 2017, really? xD

apps/talk.sh
```diff
@@ -1,13 +1,36 @@
...
+
+echo "Installing Netdata..."
```
mario: What is netdata? xD

apps/talk.sh
enoch85: Hehe, thanks!

apps/talk.sh
```diff
@@ -25,12 +48,15 @@ no-tlsv1_1
 cipher-list="ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AES:RSA+3DES:!ADH:!AECDH:!MD5"
 no-loopback-peers
 no-multicast-peers
+TALK_CREATE
+    echo "TALK_CONF was successfully created"
+fi
 
----- end of EOF -------
-
+# Retsart the TURN server
```
mario: Restart

apps/talk.sh
```diff
@@ -25,12 +48,15 @@ no-tlsv1_1
...
 sudo systemctl restart coturn
 
-
+# Install and enable app from App Store 
+install_and_enable_app spreed
 
 # occ_command for setting values in app (check with @mario which ones that are avaliable)
```
mario: available

enoch85:
> Well, I was telling you to have a separate subdomain.

Thanks! It works, just run the same stuff as for Collabora then. :+1:

apps/talk.sh
enoch85: [CTRL+C] [CTRL+V]. :)

enoch85:
So, @mario which `occ` commands are available?

mario:
Btw. you set port as 443 - that will conflict with apache/nginx. Why not run it on a regular port and do a virtualhost like I suggested?

enoch85:
> Btw. you set port as 443 - that will conflict with apache/nginx. Why not run it on a regular port and do a virtualhost like I suggested?

Didn't know that you couldn't use port 443, don't you want to use Apache/Nginx? That's how it's done in Collabora. Reverse Proxy to localhost, and they can be run on the same port externally.

enoch85:
Welcome to join the discussion @penzoiders :)

mario:
Ah I misunderstood, ignore me.

No commands I'm afraid available yet. File a bug? :D

enoch85:
> No commands I'm afraid available yet.

Hehe, that should be in the top of the list. :) Make it more available, easier to install.

> File a bug?

Sure!

mario:
Btw do you configure http2?

enoch85:
> Btw do you configure http2?

It's optional [when installing](https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L361-L388), but not default.

mario:
But why? xD

enoch85:
> Buy why

Because it may break things along the way. Sure Ondrejs PPA is pretty rock solid, but things like Spreed won't work (according to the contributor of the code.)

mario:
Talk works via http2 for sure.

enoch85:
OK, great, then users that wants it can have it. :)

Still, as long as HTTP2 isn't default in Ubuntu I won't force it.

penzoiders:
OK I've pulled everything to my fork and started working on it.
do you agree if we go for setting up just the turn server on another port (as I did) suggesting a widely accessible port (I used one that is commonly reachable even by public hotspots, or we let user choose) without messing with proxying coturn (that I don't even know if is possible)?

it you can investigate on the occ_command to set the STUN TURN srever values it would be great, I'll focus on the rest.

penzoiders:
I've checked and coturn is not proxable, or at least it's near black magic. so screw-it and go for a custom port. I'm baking the script

mario:
Really? o.O Looking at the code I haven't seen the commands o.O

penzoiders:
@mario I didn't push anything yet.
I'm just unit-testing as hell. now I'll fire up a test vm.

enoch85:
> or we let user choose

Let's set a default port, what if the user sets port 21, 993, 587, 23 or yeah, you get it... :)

penzoiders:
I'm almost done the occ commands are pissing me off.. instead of updating the field is creating a line per charachter. tried both plain and json and json_pretty  options.

need lillte help here:
```
    STUN_SERVERS_STRING="[$TURN_DOMAIN:$TURN_PORT]"
    TURN_SERVERS_STRING="[{"server":"$TURN_DOMAIN:$TURN_PORT","secret":"$TURN_SECRET","protocols":"udp,tcp"}]"
    occ_command config:app:set spreed stun_servers --value="$STUN_SERVERS_STRING" --update-on
ly --output json
    occ_command config:app:set spreed turn_servers --value="$TURN_SERVERS_STRING" --update-on
ly --output json
```


> Let's set a default port, what if the user sets port 21, 993, 587, 23 or yeah, you get it... :)
I actually use 587 since it's reachable from nearly any connection... do you have suggestions for other?
usually NextcloudVM is a dedicated one, don't have SMTP in it, this is reserved for SMTP over TLS... what do you think? for me it's ok.

penzoiders:
ok, found the issue: backslash galore!
```
    STUN_SERVERS_STRING="[\"$TURN_DOMAIN:$TURN_PORT\"]"
    TURN_SERVERS_STRING="[{\"server\":\"$TURN_DOMAIN:$TURN_PORT\",\"secret\":\"$TURN_SECRET\",\"protocols\":\"udp,tcp\"}]"
    occ_command config:app:set spreed stun_servers --value="$STUN_SERVERS_STRING" --update-only --output json
    occ_command config:app:set spreed turn_servers --value="$TURN_SERVERS_STRING" --update-only --output json
```

almost there guys

enoch85:
> I actually use 587

I thought you where joking at first. :) 

I'm not a big fan of using a standard port for something else, like Talk. Use something like 50001 instead in that case.

> almost there guys 

You're awesome! Let's get the real `occ` commands in later when @mario or someone else develops it.

penzoiders:
sorry guys, done another pull request #562 had no permissions on this repo... I'll suggest to let the user choose the port at this point, it's best option IMO.  have a test on this, everything is unit tested, should work out of the box.

enoch85:
> should work out of the box.

Where did the TLS cert generation go? 

Let's close this in favor of the other one.

penzoiders:
> I thought you where joking at first. :)
>
>I'm not a big fan of using a standard port for something else, like Talk. Use something like 50001 instead in that case.


I don't find it to be an issue, ok may not be polite, but that port is almost always open... if you cannot use 443 then go for anything.

Certificates are the ones of the standard nextcloud instance. maybe just put a check on "is there a Let's encrypt cert" before continuing

PR #562, opened 2018-06-07
Talk App install with TURN/STUN server over TLS

penzoiders:
Sorry for opening another PR but I had no permissions to edit the current one.

Fix https://github.com/nextcloud/vm/issues/545


penzoiders:
@enoch85 what do you think about letting user choose the TURN port? probably it's the best option since one needs to know wich port it's using and to avoid issues it will be better for him/her to forward it directly to the NC box without messing with crossed port forwardings.

enoch85:
So 443 can't be used?

Yes, let the user choose between 3 options. Use a whiptail for it and let 587 be one of them.

penzoiders:
> So 443 can't be used?

no, you'll need another IP to use that since apache is locking it and you can't proxy coturn

> Yes, let the user choose between 3 options. Use a whiptail for it and let 587 be one of them.

I'll go for a free digit with suggested default (ppl will go for that one 90% of cases) eg: 5001 or whatever
now I'll look at what ports are used by the standard vm and I'll make you a list to check against to eventually prompt the user with "hey pal, dont be an asshole, set another port"

penzoiders:
maybe coturn is proxable but if so is a huge PITA and I'll avoid it.

enoch85:
@penzoiders Do whatever you feel is best. I'll review it later. Thanks a lot btw. :)

penzoiders:
this is the logic, then you can snap in where you prefer with your format/style.. but this does the job.
I've audited 10 different vms I've running and those are all the ports (except some monitoring agent) that I've found in listening state.

```bash

containsElement () {
  local e match="$1"
  shift
  for e; do [[ "$e" == "$match" ]] && return 0; done
  return 1
}

NONO_PORTS=(22 25 53 80 443 3306 5432 7983 8983 10000)
if containsElement $TURN_PORT "${NONO_PORTS[@]}"
then
        echo "Choose another port..."
else
        echo "Port should be free..."
fi
```

enoch85:
@penzoiders Thanks! I was thinking a msg_box and a regular `read` (user input).

Are we about done now? Have you tested it in whole?

penzoiders:
Not yet tested as of now.. yesterday I fell asleep. Will try a shot today but my schedule is freaking right.
Msg_box and read is fine. If you can add it I'll test out in coffee break.

penzoiders:
@enoch85 in about 1 hour and half I should be able to test the script as-is now. if you manage by that time to put the msg_box thing for the user-inputed port it would be perfect. so we can then spin up vms to test the RC version of this script

enoch85:
I have no idea it my fix works, and I can't test. :( 

But it explains how I'm thinking at least.

penzoiders:
awesome. I'll be testing this as soon as I end my current task and have a huge coffeee

enoch85:
Hehe: https://stackoverflow.com/a/8574392

penzoiders:
> Hehe: https://stackoverflow.com/a/8574392

=D Late night copy-paste from stack overflow is one of my favourite sports

enoch85:
Haha! I found it searching for something else. 

alt text

enoch85:
Just for reference: https://blog.netways.de/2017/08/16/setting-up-a-turn-server-for-nextcloud-video-calls/

penzoiders:
> Just for reference: https://blog.netways.de/2017/08/16/setting-up-a-turn-server-for-nextcloud-video-calls/

I used that before, then I ended up with my other config that works on UDP too and has better logging (out of ShitstemD)

enoch85:
> that works on UDP too

Hmm, maybe we should improve this then?

penzoiders:
sorry, the new config is tested and working on udp too. what I've put in the script yesterday is already OK.

enoch85:
@penzoiders So did you test the scripts? I guess they don't exit with 0?

penzoiders:
I'm testing right now... It's a bit of a PITA to test because of all the lib.sh fetched by hardcoded link in all scripts, I cloned my repo and did a savaga `sed` to use the local copy + changed `GITHUB_REPO` variable.

it's installing right now. feeback in minutes.

enoch85:
Just put all the vars inside the script itself and comment out the curl at the top. :)

penzoiders:
>  Just  put all the vars inside the script itself and comment out the curl at the top. :)

manual editing for this test is always needed, `sed` is my friend.. I'll set up an alias for this kind of tricks

penzoiders:
it failed to download the script, maybe due to my shitty editing. I'll try to install it after it finishes. I got 2 more test VMs ready in the line..

penzoiders:
> Just do a snapshot and revert when it fails. Much quicker 

I'm about pushing VMs into production =P and testing this automation instead of doing manually

enoch85:
Whatever suits you the best. :) :+1:

apps/talk.sh
```diff
@@ -105,9 +105,15 @@ echo "TURN_CONF was successfully created"
...
+# Warn user to open port
+msg_box "Open $TURN_PORT on your firewall or your TURN/STUN server won't work!
+After you hit OK the script will check for the firewall and eventually exit on failure.
+To run again the setup, after fixing your firewall:
+sudo wget $APP/talk.sh
+sudo bash talk.sh"
 
 # Check if the port is open
-check_open_port "$TURN_PORT" "$TURN_DOMAIN"
+check_open_port "$TU RN_PORT" "$TURN_DOMAIN"
```
enoch85: typo?

enoch85: I actually don't get this..? In the LE scritps we just do like [this](https://github.com/nextcloud/vm/blob/master/lets-encrypt/activate-ssl.sh#L47-L52) and [this](https://github.com/nextcloud/vm/blob/master/apps/collabora.sh#L36) in the Collabora script.

enoch85: Better to check if the port is open and exit if it's not IMHO. That's why I wanted to put the check in the beginning as we are using the Nextcloud URL anyway it doesn't matter if it's in the top of the script. I mean, why start installing stuff if it will fail in the end anyway? :)

penzoiders:
```bash
if [ -d "$NC_APPS_PATH"/spreed ]
```

in my vm template I had not spreed there, enabled it manually and disabled to simulate behaviour.
can you please fix it that if not there it just downloads again the spreed subfolder?

penzoiders:
OK All working now!
if you can fix the "download if spreed folder is missing" thing I'll have cofffee and try again.
you can merge as of now IMHO.

enoch85:
> in my vm template I had not spreed there,

So what you are saying i that you didn't have the app in `/var/www/nextcloud/apps/spreed`? That's the default path. I never heard of anyone having it somewhere else, is that even possible?

penzoiders:
nevermind, I'll fix it myself in my vm template. (I had partially installed, just to go with startup script vms).. better starting from scratch each time.  go and merge.

enoch85:
> go and merge.

Thanks a lot! :+1: Will have to test it first and improve some variables and functions. I don't like taking the domain name from the config.php for instance. It's not generic. :)

penzoiders:
OK, now the mechanism is tested, and everything is fine. I have to stop here 'cause I got ultra late on my schedule, but is always a pleasure working with you and this is a awesome project.
Let's catch up in the next 24h feature rush =P

cheers!

enoch85:
OK, fixed some bugs and now everything is working except the last line:

```
Config value stun_servers for app spreed set to ["test.danielhansson.nu:587"]
Config value turn_servers for app spreed not updated, as it has not been set before.
sudo -u www-data php /var/www/nextcloud/occ config:app:set spreed turn_servers --value=[{"server":"test.danielhansson.nu:587","secret":"v7X9Wj4WOXHu4eJ3whhblgnX7O","protocols":"udp,tcp"}] --update-only --output json failed
```

enoch85:
Also, the port I set was 3333, still 587 was set in the script.

enoch85:
@penzoiders OK, so the port is correct now and the values are correct. Will test one last time.

enoch85:
This is ready to be merged. Will change the repo then do it.

ZaxLofful:
Sorry to bug, are these fixes available via the updater or do I need to export my DB and then reinstall NextCloud? I recently updated to the last version of NextCloud from version 12 (more than a year old) and I am unable to use talk.

The official documentation says it should just work, and thenI saw these because udpates since I favorited NextCloud/vm

enoch85:
@ZaxLofful Please report a separate :+1:  issue, don't write in old PRs. Thanks!

Issue #535, opened 2018-05-09
NextCloud VM on Xenserver


apatek1975:
Hello,
I am having issues with getting the current NextCloud VM to install properly on Xenserver.
When I import the appliance and start it up, it attempts to run the startup script but fails. It takes me to the purple screen that says "It appears something went wrong with the update". It says my data directory is not writable and throws an exception that says "Environment not properly prepared and references line 148 in applicatin.php file. It also says sudo -u www-data php /var/www/nextcloud/occ -V failed. I attempted many different permission changes etc, but nothing seemed to resolve this. I also made sure the data volume was not set to read only in xenserver.

enoch85:
Have you made two disks according to the instructions?

apatek1975:
Yes. When I import the OVA into xenserver it does create two disks. The
data drive is mounted as /mnt/ncdata.

I was actually able to get it working by changing the permissions on the
data drive to be owned by www-data user I think. By default everything  was
root:root. I am going to try it again from scratch because I made a lot of
different changes trying to get it to work.

enoch85:
That's strange because this is how the permissions look like on a freshly mounted OVA:

**/mnt/ncdata**
```
total 28
drwxrwx--- 5 www-data www-data    9 maj 10 14:30 ./
drwxr-xr-x 3 root     root     4096 maj  7 23:45 ../
drwxr-xr-x 7 www-data www-data    7 maj 10 14:30 appdata_ocnnap7p5mox/
drwxr-xr-x 2 www-data www-data    3 maj  7 23:46 files_external/
-rw-r--r-- 1 root     www-data  324 maj  7 23:46 .htaccess
-rw-r--r-- 1 www-data www-data    0 maj  7 23:46 index.html
drwxr-xr-x 3 www-data www-data    3 maj  7 23:46 ncadmin/
-rw-r--r-- 1 www-data www-data    0 maj 10 14:30 nextcloud.log
-rw-r--r-- 1 www-data www-data    0 maj  7 23:46 .ocdata
```

apatek1975:
I just tried a fresh install. It fails the startup script with the
following information:

ncadmin@192.168.1.103's password:
Welcome to Ubuntu 18.04 LTS (GNU/Linux 4.15.0-20-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Thu May 10 20:24:40 CEST 2018

[...]

PR #563, opened 2018-06-08
Xenserver/XCP Hypervisor Support

zalonic:
Hello,

While reviewing some bits we came across this [comment](https://github.com/nextcloud/vm/issues/535).
We have been using Xenserver and XCP for a while and thought we would post how we have been installing this on ours.

We've cloned the format-sdb.sh script and amended it to work with Xen/XCP, if it makes it easier to maintain we can merge the scripts together with some variables into one format.sh

Please let me know your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ Keep asking questions so that we can add them here.
 <br />
 **A:** You can download the latest release [here](https://my.vmware.com/web/vmware/free#desktop_end_user_computing/vmware_workstation_player/12_0).
 
+**Q: Can I use Xenserver or XCP-ng?**
+<br />
+**A:** Yes you can, however, these have to be manually setup, the OVA is not configured for these Hypervisors.
+
 **Q: The downloaded file is just a few kilobyte, or corrupted**
 <br />
 **A:** This is due to heavy load on the server. Please try again in a few minutes.
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -171,8 +171,14 @@ check_command apt install -y \
 # Install VM-tools
 install_if_not open-vm-tools
 
+# Check if Xenserver
+if [ /dev/xvdb ]; then
+# Format /dev/xvdb to host the ncdata
+run_static_script format-xvdb
+else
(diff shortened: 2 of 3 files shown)


zalonic:
Okay maybe put a pin in this until we see fix the build failure.

enoch85:
What happened here?

zalonic:
Sorry, slight disagreement over this our end, my account was logged in so they posted the changes I had made for the Xenservers we use, however, it doesn't actually work correctly because... although it now recognised the XVDB disks and will install correctly on a Xenserver, if you try it on a standard VMware or VBox is no longer sees SDB.

I told my guy to close this until I had actually worked out how to do this correctly and then we can post it as a pull request.

Sorry about this.

enoch85:
@zalonic No problems! Would be great if you could fix Xenserver support.

I suggest a function in the existing script, not adding another one. Looking forward to your contribution!

penzoiders:
we should also include KVM VirtIO support (eg: `/dev/vdX` disks) let's fix this togheter.

enoch85:
@penzoiders Sure! Been having some bug reports about Xen and it would be awesome to extend the support!

zalonic:
Hi all, right think I've worked it out. Just making the changes and will upload as soon as I get a chance. I will modify the entire fork I have to download the scripts from my repository so we can run the install from there to test and then if all goes well I will revoke all my amendments except the disk modifications for the commit.

@penzoiders can you confirm how the disks work on KVM as I'm not familiar, are the devs xva, xvb and when partitioned xva1, xvb1, xvb2?

zalonic:
Sometimes... I dislike people... well I've created the script and it works on HyperV and Xen, KVM is still untested as I don't know exactly how the disks are labelled in the system.

However somewhere between me posting my last comment and this one, one of my dev guys has deleted the Repo because he thought it wasn't needed [angry face].

Blocked all of them from my account now as I can't have that happening again.

Unfortunately, this means I will have to open a new pull request.

enoch85:
@zalonic 

> open a new pull request.

No problem, you can open as many as you want ;) :+1:

PR #566, opened 2018-06-14
Add Nextcloud Talk info

Ivansss:
What do you think about changing the last message in your script to this one @enoch85 ?
This way you don't have to keep track of the available mobile clients (and possible future clients) because we'll keep that page updated with the links to all available clients.

### DIFF
--- a/apps/talk.sh
+++ b/apps/talk.sh
@@ -143,6 +143,7 @@ fi
 
 if [ -d "$NC_APPS_PATH"/spreed ]
 then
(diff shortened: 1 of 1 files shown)


enoch85:
Sure!

enoch85:
Sorry, removed the sign-off by mistake. I always remove all commits and squash it, did the same thing here without thinking.

Ivansss:
No problem 😉

PR #603, opened 2018-08-06
Change the FTS Docker name

enoch85:
cc @Ark74 

Not tested

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -110,8 +110,8 @@ APACHE2=/etc/apache2/apache2.conf
 [ ! -z "$ES_INSTALL" ] && NCADMIN=$(sudo -u www-data php $NCPATH/occ user:list | awk '{print $3}')
 [ ! -z "$ES_INSTALL" ] && ROREST=$(tr -dc "A-Za-z0-9" < /dev/urandom | fold -w "$SHUF" | head -n 1)
 [ ! -z "$ES_INSTALL" ] && DOCKER_INS=$(dpkg -l | grep ^ii | awk '{print $2}' | grep docker)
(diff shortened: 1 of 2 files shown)


lib.sh
```diff
@@ -110,8 +110,8 @@ APACHE2=/etc/apache2/apache2.conf
 [ ! -z "$ES_INSTALL" ] && NCADMIN=$(sudo -u www-data php $NCPATH/occ user:list | awk '{print $3}')
 [ ! -z "$ES_INSTALL" ] && ROREST=$(tr -dc "A-Za-z0-9" < /dev/urandom | fold -w "$SHUF" | head -n 1)
 [ ! -z "$ES_INSTALL" ] && DOCKER_INS=$(dpkg -l | grep ^ii | awk '{print $2}' | grep docker)
-[ ! -z "$ES_INSTALL" ] && nc_rores6x="ark74/nc_fts:latest"
-[ ! -z "$ES_INSTALL" ] && rores6x_name="ark74/nc_fts"
+[ ! -z "$ES_INSTALL" ] && nc_fts="ark74/nc_fts:latest"
+[ ! -z "$ES_INSTALL" ] && fts_es_name="fts_es63x"
 # Talk
```
enoch85: Great idea!

Will you change the name on your Docker so that the script won't break?

enoch85:
Would be even better if it was generic without any version. As little maintenance as possible is better if you ask me. :)

Ark74:
`fts_esror` ?

enoch85:
> fts_esror 

Perfect!

enoch85:
Merge?

lib.sh
Ark74: It's already set.

enoch85:
Let's do it!

PR #609, opened 2018-08-07
add calculat_max_children to all apps with docker

enoch85:


enoch85:
Bad idea since it will break the scripts for them who doesn't run php-fpm

We do the check in the startup-script anyway.

PR #633, opened 2018-08-17
prepare for nextcloud 14

enoch85:


enoch85:
NC 14 is due for tomorrow, no harm is done with merging this now.

--> MERGE!

PR #636, opened 2018-08-21
Replace phppgadmin with adminer

danielkesselberg:


enoch85:
Thanks a ton! Missed that one!

PR #646, opened 2018-08-31
Fixed

Ark74:
From fts 0.99.1 app forwards this issue is fixed.


lib.sh
```diff
@@ -114,7 +114,7 @@ APACHE2=/etc/apache2/apache2.conf
 [ ! -z "$ES_INSTALL" ] && NCADMIN=$(sudo -u www-data php $NCPATH/occ user:list | awk '{print $3}')
 [ ! -z "$ES_INSTALL" ] && ROREST=$(tr -dc "A-Za-z0-9" < /dev/urandom | fold -w "$SHUF" | head -n 1)
 [ ! -z "$ES_INSTALL" ] && DOCKER_INS=$(dpkg -l | grep ^ii | awk '{print $2}' | grep docker)
-[ ! -z "$ES_INSTALL" ] && nc_fts="ark74/nc_fts:latest"
```
enoch85: Why not use the latest tag?

enoch85:
Let's merge this, we can add the `latest` tag later if needed.

lib.sh
Ark74: docker pulls will look by default for latest tag in case there is no"latest" tag, then it will pull the "newest" image pushed.
Once that said, I'm tagging "latest" to the latest tested version, so no need to add latest on the command.

Cheers!

PR #652, opened 2018-09-04
change check for sudo user

enoch85:
needs testing, especially with a pure root account


Ark74:
Testing...
sudo_root

Ark74:
If this pure root was in order to fix FTS issue with password change, then it didn't work.
1. Clean install
2. Enable root password  (sudo passwd root)
3. Login with root
4. Run sudo_user branch
fts1-4

enoch85:
You shoudn't run the script with a pure root user. That's not by design and you will get warned if you do so.

I'm testing the master branch now to see if the fixes I made earlier made any difference.

Worst case, move the password stuff before installing of apps.

Ark74:
Oh man, then did I misunderstood this?

> needs testing, especially with a pure root account

I thought that was the purpose of this test.
I'll leave it at that.

Ark74:
I'll second this;

> Worst case, move the password stuff before installing of apps.

enoch85:
> Oh man, then did I misunderstood this?

I was thinking if you are running from DO, you are prompted to change user, but in case you don't this should work anyway. 

BTW, this has nothing to do with FTS failing afaik.

Btw, join me on Talk: https://cloud.techandme.se/call/db33x6wj

PR #668, opened 2018-09-28
minor tuneup

Ark74:
avoid need of user interaction

### DIFF
--- a/apps/collabora.sh
+++ b/apps/collabora.sh
@@ -100,7 +100,7 @@ fi
 
 # Install Collabora docker
 docker pull collabora/code:latest
-docker run -t -d -p 127.0.0.1:9980:9980 -e "domain=$NCDOMAIN" --restart always --cap-add MKNOD collabora/code
+docker run -t -d -p 127.0.0.1:9980:9980 -e "domain=$NCDOMAIN" --restart always --name code --cap-add MKNOD collabora/code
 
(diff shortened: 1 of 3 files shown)


enoch85:
Nice change! Thanks! Can you do the same for OnlyOffice please?

apps/collabora.sh
```diff
@@ -100,7 +100,7 @@ fi
 
 # Install Collabora docker
 docker pull collabora/code:latest
-docker run -t -d -p 127.0.0.1:9980:9980 -e "domain=$NCDOMAIN" --restart always --cap-add MKNOD collabora/code
+docker run -t -d -p 127.0.0.1:9980:9980 -e "domain=$NCDOMAIN" --restart always --name code --cap-add MKNOD collabora/code
```
enoch85: Maybe "collabora" is better suited. I mean what is "code", some PHP code or what? I know, but users may not. :D

Ark74: I preferred CODE since a lot of users feel tricked with the limitations that "collabora" has, when they are using CODE not Collabora.
Whatever suits best.

Ark74: collabora-de :laughing:

enoch85: OK, fine, let's go with that.

Please make the same changes to onlyoffice.sh, thanks :)

PR #669, opened 2018-09-28
fixes for fts

enoch85:
The user can't be in CAPITAL letters and it doesn't matter if it exists or not in Nextcloud, so make it random.


enoch85:
cc @Ark74 FYI

Ark74:
> The user can't be in CAPITAL letters and it doesn't matter if it exists or not in Nextcloud, so make it random.

Ohh! 
I wasn't aware of that :open_mouth: 
Great to know :+1:

Issue #680, opened 2018-10-05
Implement PageKite


enoch85:
https://twitter.com/techandme_se/status/1048077706775203842

@penzoiders @pagekite

enoch85:
cc @pagekite

enoch85:
@penzoiders @pagekite I've started some veeeeeery basic scripting. Feel free to dig in!

BjarniRunar:
Just a quick comment: I really do NOT see PageKite as an "alternative" to Let's Encrypt (that wording was used on the PR).

PageKite is an alternative to punching holes in your firewall or setting up a Tor Hidden Service.

From a security point of view, you absolutely still want Let's Encrypt and end-to-end TLS, even if you're using PageKite. From a performance point of view, just relying on PageKite's wildcard certs (SSL terminated at the relay) may be better, but if you've already got a working Let's Encrypt flow you should absolutely use them together.

enoch85:
@BjarniRunar Thanks for your input. As I'm very novice in PageKite, please help me with the PR. 

It's just a basic PR to get things started, and to get you guys motivated. :) It' easier to edit when things are in place already imo.

To me PageKite is an alternative to using your own domain, which Let's Encrypt will give you, so hence the wording. Could defiantly be improved a lot though.

BjarniRunar:
You can also use your own domain with PageKite. :smile_cat:   I'll be happy to help as soon as I have a bit more time (I am travelling at the moment). I'm leaving this browser tab open...

enoch85:
@BjarniRunar 

> I'm leaving this browser tab open...

Is it still open? :)

BjarniRunar:
Haha, it is now! I'll take a look today and try and send you some advice.

BjarniRunar:
To re-iterate: PageKite is mostly useful for people who don't have a public IP address, and/or don't want to (or cannot for some reason) configure port forwarding in their router. For example if people want to just try spinning up OwnCloud in a VM on their Windows desktop - it's a great fit for that sort of experimenting.

I took a look, but since I haven't tried the Nextcloud setup process, I'm a little bit unsure. But let me have a go... since PageKite is very flexible, it may not make sense to support all the different possible setups at first.

If you assume people go to https://pagekite.net/ first and create an account, then you could get by just asking them for a) the name of their "kite" (e.g. owncloud-user.pagekite.me) and b) the PageKite service shared secret.

[...]

PR #701, opened 2018-10-22
include PageKite as an alternative

enoch85:
Fixes https://github.com/nextcloud/vm/issues/680


ZaxLofful:
What is it an alternative to?

enoch85:
> What is it an alternative to? 

Let's Encrypt with your own domain.

PR #727, opened 2018-11-27
Let user select the device for the nextcloud data

ggeorgg:
One problem I see is that the user can fill in any text. There is no input validation...

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -103,17 +103,15 @@ fi
 install_if_not netplan.io
 install_if_not network-manager
 
-# Format /dev/sdb to host the ncdata
-
-# Let the user decide which disk to use
+# Format the second disk
 msg_box "This VM is designed to run with two disks, one for OS and one for DATA.
 
-You will now get the option to decide which disk you want to use, or run the automatic script that will choose the available disk automatically."
-if [[ "yes" == $(ask_yes_or_no "Do you want to choose disk by yourself?") ]]
+You will now get the option to decide which disk you want to use for DATA, or run the automatic script that will choose the available disk automatically."
+if [[ "no" == $(ask_yes_or_no "Do you want to choose disk by yourself?") ]]
 then
-	run_static_script format-chosen
-else
 	run_static_script format-sdb
+else
+	run_static_script format-chosen
 fi
 
 # Change DNS system wide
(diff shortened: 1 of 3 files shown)


enoch85:
Thanks! 

Input validation is a must btw as this script is crucial for everything to work.

More options will also generate more errors as users probably will do it the wrong way. Let me think about this.

ggeorgg:
I modified the commit. Please have try now. What do you think?

enoch85:
DEVTYPE is still changed in [here](https://github.com/nextcloud/vm/pull/727/files#diff-8bff978ff8cc8677ee9420cf74d79a61R96) which means that your code won't have any effect. Please make it an option instead. Like; "Would you like to choose your own disk, or would you like to choose the second disk automatically?"

Also please fix the travis errors: https://travis-ci.org/nextcloud/vm/builds/461460259?utm_source=github_status&utm_medium=notification

enoch85:
Sorry, I was a bit tired this morning. It says DISKTYPE, not DEVTYPE. :)

enoch85:
Actually, when thinking about it, I think it's better to make this a seperate script and ask the user which script to run. 

1. Automatically = format-sdb.sh
2. Choose yourself = format-chosen.sh

ggeorgg:
I think it's not worth to make another script. It's not that much code and it's not a completly other thing. So I would leave it in this script here.
The new lines need to be after the "checks what Hypervisor disks are available" because I wanted to have a default value for DEVTYPE.

enoch85:
OK, I haven't tested this, have you?

Why I would want a second script is because it's easier when if comes to the IF argument. Adding this adds more complexity which also raises the risk of a failure. The whole logic changes with this change, so it would be easier to separate it into two scripts, where you can have two different logic's.

Makes sense?

ggeorgg:
I added another script. I am not that happy with this solution because there is to much redundant code now...

ggeorgg:
I have just tested the code snipped in a tests file, but not yet the whole script.

enoch85:
That's not what I meant at all.

I can do the changes and you can test them. I'll have a look later today.

enoch85:
Sorry, the diffs weren't updated. That's a correct change.

enoch85:
@ggeorgg I tested he script now and it seems to work. If you are willing to maintain format-choosen.sh then I will merge this. I that OK by you?

ggeorgg:
I will give my best ;)

PR #735, opened 2018-12-08
ask_yes_or_no questions not working

ggeorgg:
I just wanted to test the script to select the disk and I tried it using
`curl -sSL https://raw.githubusercontent.com/nextcloud/vm/master/nextcloud_install_production.sh | sudo bash`
but unfortunately I was not questioned the yes_or_no questions... I think the proposed change will avoid this problem but I have not tested it directly like this.
Do you also see this problem? Or is it only me?

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The script will mount and format the drive. Please select Manually Format & Moun
 
 #### Installation
 1. Get the latest install script from master and install it with a sudo user:<br>
-`curl -sSL https://raw.githubusercontent.com/nextcloud/vm/master/nextcloud_install_production.sh --progress-bar && sudo bash nextcloud_install_production.sh`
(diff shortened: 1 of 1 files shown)


enoch85:
Please test if it works. I'm busy with some Wordpress development right now. 

Thank you very much!

ggeorgg:
Here the reason why the old method is not working for me: [stackechange](https://unix.stackexchange.com/questions/180383/piping-a-script-with-read-to-bash)
And this is how It looks like on my virtualbox when testing:
virtualbox_ubuntu server_09_12_2018_10_03_02
(The content of the test file on the right hand side. Commands on the left window pane)

enoch85:
Use plain wget instead works. Any specific reason you would want `curl`?

ggeorgg:
No specific reason for curl on my side. I just stayed with as it has already been used before.
I tested the latest commit and can confirm that the yes_or_no questions work like this.

PR #736, opened 2018-12-09
cannot stat /etc/netplan/01-netcfg.yaml

ggeorgg:
On my system the file /etc/netplan/01-netcfg.yaml is not available after sudo apt install netplan.io. 
We should check if the file exists befor making a backup, otherwise the script exists without a real reason.

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -103,10 +103,14 @@ You must have a working network connection to run this script.
 
 You will now be provided with the option to set a static IP manually instead."
 
-    # Copy old interfaces file
-msg_box "Copying old netplan.io config file file to:
-/tmp/01-netcfg.yaml_backup"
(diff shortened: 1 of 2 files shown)


enoch85:
Are you running Ubuntu Server 18.04?

Anyway, please do an if command instead.

enoch85:
Sorry, saw now that it was `if`

enoch85:
OK, so now it copies all files instead of just one.

ggeorgg:
Yes I am using Ubuntu 18.04.1 LTS:
lsb_release -a gives me this on a new installation:
grafik
I am also surprised as you never had this problem?

Anyway, I can confirm that this commit does what it is expected to do.

enoch85:
No I never had this issue, but this it better anyway, so thanks for your PR! :D

PR #737, opened 2018-12-10
enabledPreviewProviders not correctly set

ggeorgg:
I guess nobody has ever tested this but setting a string value needs to be done with double quotes.

Otherwise nextcloud config will have four backslashes instead of two (see screenshots below).

[nextcloud occ command](https://docs.nextcloud.com/server/14/admin_manual/configuration_server/occ_command.html#setting-a-single-configuration-value)
Test:
grafik

grafik

grafik

grafik


enoch85:
Thanks a lot @ggeorgg! You rock! :rocket:

PR #740, opened 2018-12-12
Don't fail hard when app is not installable 

ggeorgg:
Prevent script from failing if NC app can not be installed due to wrong max-version of app (may happen if we move to a newer release of NC) or due to false app name (which should never happen).


enoch85:
@ggeorgg I'm very thankful for all your contributions! Keep em coming! :+1: :100:  :rocket:

PR #741, opened 2018-12-12
Replace almost all echos by printf function with color

### LINKED ISSUE #621: Use print functions instead of echo
Let's be consistent and use 3 `printf` `functions()` instead of `echo`.

```
printf_green() {
printf "%b" "${IGreen}$1\n${Color_Off}"
}

printf_red() {
printf "%b" "${IRed}$1\n${Color_Off}"
}

printf_normal() {
printf "%b" "${Color_Off}$1\n${Color_Off}"
}
```

# Example 
`printf_green hello`

Would print "hello" in green

ggeorgg:
I want to use a colored output for all echos from our scripts to make it easier to spot if an output is coming from our scripts or from other commands.
This should close #621 
Only the script nextcloud.sh and instructions.sh have echos left. I am not quite sure if we should replace them like all the other echos.

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -17,7 +17,7 @@ is_process_running dpkg
 # Install curl if not existing
 if [ "$(dpkg-query -W -f='${Status}' "curl" 2>/dev/null | grep -c "ok installed")" == "1" ]
 then
-    print_text_in_color "$Cyan" "curl OK"
+    print_text_in_color "$Green" "curl OK"
 else
     apt update -q4 & spinner_loading
     apt install curl -y
@@ -26,7 +26,7 @@ fi
 # Install lshw if not existing
 if [ "$(dpkg-query -W -f='${Status}' "lshw" 2>/dev/null | grep -c "ok installed")" == "1" ]
 then
-    print_text_in_color "$Cyan" "lshw OK"
+    print_text_in_color "$Green" "lshw OK"
 else
     apt update -q4 & spinner_loading
     apt install lshw -y
@@ -35,7 +35,7 @@ fi
 # Install net-tools if not existing
 if [ "$(dpkg-query -W -f='${Status}' "net-tools" 2>/dev/null | grep -c "ok installed")" == "1" ]
 then
-    print_text_in_color "$Cyan" "net-tools OK"
+    print_text_in_color "$Green" "net-tools OK"
 else
     apt update -q4 & spinner_loading
     apt install net-tools -y
@@ -219,7 +219,7 @@ cat << HTTP2_ENABLE > "$HTTP2_CONF"
     H2Direct on
 </IfModule>
 HTTP2_ENABLE
-print_text_in_color "$Cyan" "$HTTP2_CONF was successfully created"
+print_text_in_color "$Green" "$HTTP2_CONF was successfully created"
 a2enmod http2
 restart_webserver
 
@@ -403,7 +403,7 @@ then
 
 </VirtualHost>
 HTTP_CREATE
-    print_text_in_color "$Cyan" "$HTTP_CONF was successfully created"
+    print_text_in_color "$Green" "$HTTP_CONF was successfully created."
 fi
 
 # Generate $SSL_CONF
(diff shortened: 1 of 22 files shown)


enoch85:
WOW, Thanks!

This must be tested though. All the changes in all places.

nextcloud_install_production.sh
```diff
@@ -35,7 +35,7 @@ fi
 # Install net-tools if not existing
 if [ "$(dpkg-query -W -f='${Status}' "net-tools" 2>/dev/null | grep -c "ok installed")" == "1" ]
 then
-    echo "net-tools OK"
+    print_text_in_color "$Cyan" "net-tools OK"
```
enoch85: I think that when stuff are "OK" the text should be Green instead.

nextcloud_install_production.sh
```diff
@@ -212,14 +212,14 @@ check_command apt install -y \
...
-echo "Enabling HTTP/2 server wide..."
+print_text_in_color "$Cyan" "Enabling HTTP/2 server wide..."
 cat << HTTP2_ENABLE > "$HTTP2_CONF"
 <IfModule http2_module>
     Protocols h2 h2c http/1.1
     H2Direct on
 </IfModule>
 HTTP2_ENABLE
-echo "$HTTP2_CONF was successfully created"
+print_text_in_color "$Cyan" "$HTTP2_CONF was successfully created"
```
enoch85: Or when stuff are sucessfull, also Green.

static/modsecurity.sh
```diff
@@ -93,6 +93,6 @@ check_command sed -i 's|# SecDefaultAction "phase:2,nolog,auditlog,pass"|SecDefa
 
 if [ -f /etc/modsecurity/whitelist.conf ]
 then
-    echo "ModSecurity activated!"
+    print_text_in_color "$Cyan" "ModSecurity activated!"
```
enoch85: This is a sucessfull status, so Green.

enoch85:
My five cents:

Informative: $Cyan
Successful or OK: $Green
Failed: $Red

There are more echos than in this PR, would be nice to change it everywhere. nextcloud.sh and instructions can be tested by running the certain script and put this function at the top.

enoch85:
Tested some basic stuff, and it all seems to work. Good job @ggeorgg!

enoch85:
Call for testing and help @nextcloud/vm

static/modsecurity.sh
ggeorgg: Totally agree. Thats also what I had in mind. Just wanted to make one change after another.

ggeorgg:
Interessting for me would now be how to test these changes when they are not in master yet. The problem is that the lib is downloaded again for every single script and will then get the script from master branch.

Maybe we should set a default value for the variable $GITHUB_REPO but make it possible to change it with a name-value pair when executing the main script?

enoch85:
> Maybe we should set a default value for the variable $GITHUB_REPO but make it possible to change it with a name-value pair when executing the main script?

Naah, then I think it's better if we use your repo as DEV and change GITHUB_REPO to that instead (including the curl command).

Usually I test it locally.

enoch85:
So, the new repo would be `https://raw.githubusercontent.com/ggeorgg/vm/replace_echo`

ggeorgg:
Yes I understand, but I don't like this idee because I would have to change the repo name in each file to test it. Let me think about another solution to be able to easily get another branch from this repository for test cases.

enoch85:
I just tested this locally, and it seems to work. 

Will do one more test the merge.

ggeorgg:
Sorry, I don't understand how test it locally. How do you avoid getting the scripts from master branch?

enoch85:
>  scripts from master branch?

I change the curl command to the DEV repo.

ggeorgg:
But you have to do it for all files, right?

enoch85:
> But you have to do it for all files, right? 

Yes, usually I test one by one.

ggeorgg:
ok, I understand. Thats something I didn't want to do.

enoch85:
> ok, I understand. Thats something I didn't want to do.

Let me just finish up here, I'll merge this soon.

enoch85:
OK, I'm satisfied. Will merge once the checks passed.

Thank you very much @ggeorgg! :D

enoch85:
> Thats something I didn't want to do.

And yeah, regarding that. If you have a better idea I'm all ears!

ggeorgg:
You are welcome ;)

PR #747, opened 2018-12-17
change color

enoch85:


enoch85:
All good, maybe should change to IRed as well, but let's wait with that.

enoch85:
@ggeorgg That new `print_text_in_color` function is genius! Thanks for that. Found some more places where it was not set.

PR #763, opened 2019-01-16
Automatically caulculate_php_fpm

enoch85:
This is an attempt to automatically calculate the values for PHP-FPM.

Input is welcome.

cc @nextcloud/vm


enoch85:
Tested and works as expected.

I'm still open for improvements though.

PR #785, opened 2019-02-12
remove duplicate chain

AEITS:
The debian default chain is causing the ssh chain content to be
duplicated in sshd resulting in duplicate sets of rules


enoch85:
Thanks for this PR! Welcome to the Nextcloud VM repo! :+1:

PR #787, opened 2019-02-18
check if redis.so is enabled

enoch85:
This is needed as in older versions of the VM redis.so wasn't added. This adds redis.so in the correct place if:
- the php.ini file exist
- if it's not already added (if not it adds it)
- if pecl is used (which it is since we update to it)

It it's compatible with:
- php7.0 apache
- php7.2 apache
- php7.2 fpm ($PHP_INI)

 cc @nextcloud/vm for feedback


enoch85:
Tested and works!

PR #829, opened 2019-05-02
Adding 2 features (allow selection of single or dual drive setup, and selection of DNS resolver)

thejohnha:


enoch85:
OK, so this is a better approach, but, `--radiolist` doesn't work, `--checklist` does.

Problem is that with `--checklist` you can choose multiple options which is not what we want here.

If someone can solve so that `--radiolist` works, we'll merge this PR. I tried now for like 2 hours myself.

enoch85:
Tested and works as expected.

enoch85:
Thanks @thejohnha for the initial PR!

PR #836, opened 2019-05-16
add PHP version as a variable

enoch85:
For easier migration to new PHP version in later OS upgrades

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -274,7 +274,7 @@ sed -i "s|DocumentRoot /var/www/html|DocumentRoot $HTML|g" /etc/apache2/sites-av
  sed -i '14i\    # http://lost.l-w.ca/0x05/apache-mod_proxy_fcgi-and-php-fpm/' /etc/apache2/sites-available/000-default.conf
  sed -i '15i\   <FilesMatch "\.php$">' /etc/apache2/sites-available/000-default.conf
  sed -i '16i\    <If "-f %{SCRIPT_FILENAME}">' /etc/apache2/sites-available/000-default.conf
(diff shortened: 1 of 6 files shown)


nextcloud-startup-script.sh
```diff
@@ -274,7 +274,7 @@ sed -i "s|DocumentRoot /var/www/html|DocumentRoot $HTML|g" /etc/apache2/sites-av
  sed -i '14i\    # http://lost.l-w.ca/0x05/apache-mod_proxy_fcgi-and-php-fpm/' /etc/apache2/sites-available/000-default.conf
  sed -i '15i\   <FilesMatch "\.php$">' /etc/apache2/sites-available/000-default.conf
  sed -i '16i\    <If "-f %{SCRIPT_FILENAME}">' /etc/apache2/sites-available/000-default.conf
- sed -i '17i\      SetHandler "proxy:unix:/run/php/php7.2-fpm.nextcloud.sock|fcgi://localhost"' /etc/apache2/sites-available/000-default.conf
+ sed -i '17i\      SetHandler "proxy:unix:/run/php/php'$PHPVER'-fpm.nextcloud.sock|fcgi://localhost"' /etc/apache2/sites-available/000-default.conf
```
enoch85: Needs to be tested, does this actually work?

enoch85:
Maybe use "${PHPVER}" instead?

enoch85:
TODO: Remove quoiting where it's not needed (like if it's already quoted)

nextcloud-startup-script.sh
enoch85: Works!

enoch85:
Tested and works

PR #843, opened 2019-06-01
fix Sabre\DAV\Exception\BadRequest

enoch85:


enoch85:
Maybe change to `mod_reqtimeout.c`?

PR #845, opened 2019-06-06
improve docker_prune_this

enoch85:
make it possible to remove any docker image by name, rather than just the hard coded ones.

### DIFF
--- a/apps/fulltextsearch.sh
+++ b/apps/fulltextsearch.sh
@@ -68,13 +68,12 @@ then
 fi
 
 # Check & install docker
-apt update -q4 & spinner_loading
 install_docker
 set_max_count
 mkdir -p "$RORDIR"
-if docker ps -a | grep "$fts_es_name"
+if does_this_docker_exist "$nc_fts"
 then
-    docker stop "$fts_es_name" && docker rm "$fts_es_name" && docker pull "$nc_fts"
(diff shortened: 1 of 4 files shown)


apps/fulltextsearch.sh
```diff
@@ -68,13 +68,12 @@ then
...
 # Check & install docker
-apt update -q4 & spinner_loading
 install_docker
 set_max_count
 mkdir -p "$RORDIR"
-if docker ps -a | grep "$fts_es_name"
+if does_this_docker_exist "$nc_fts"
 then
-    docker stop "$fts_es_name" && docker rm "$fts_es_name" && docker pull "$nc_fts"
+    docker_prune_this "$nc_fts"
```
enoch85: fyi @Ark74 it not removes everything on a new install as well

Ark74: But the issue you reported should not happen on new installs since they have no data volume nor container running.
From what we tested, fts have issues when there is mixed data, 6.x/7.x, so starting from scratch is necessary for this new release.

I've have upgraded from es7.0 to es7.1 with no issues once I had re-indexed everything from scratch.

enoch85: The script does a re-indexing, removing the container, but I'm not sure about the volume. This issue fixes that anyway. The VM is all about automation you know. ;)

PR #862, opened 2019-07-03
Allow to pass the DIG test in some edge cases

enoch85:
**WIP**

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -233,10 +233,10 @@ domain_check_200() {
         print_text_in_color "$IGreen" "DNS seems correct when checking with dig!"
     elif [[ "$(dig +short "${1}" @resolver1.opendns.com)" != "$WANIP4" ]]
     then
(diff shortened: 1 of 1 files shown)


enoch85:
@bobdig, please test this.

bobdig:
@enoch85 Would love to but don't know how this all works.

enoch85:
OK, I'll merge this since I tested it here.

@bobdig Please download the latest activate-ssl.sh in about 15 minutes and run it just to see if it's possible to pass the test.

Thanks!

bobdig:
@enoch85 So I did a fresh install and the files got downloaded automatic. There was a nice Screen showing the problem with the two different IP addresses and a solution was given. 
But also there was a screen telling me, that there is a newer version of something, so I went with the newer one. And after that the first try of creating the cert failed. But at the given second try it succeed. 
I'll send you the sreenshots.
But all in all it worked!

PR #886, opened 2019-07-17
notify users in GUI

enoch85:
This needs to be fixed first, else it's not reliable since it's not possible to generate notifications for more users than one at the time: https://github.com/nextcloud/server/issues/16443


szaimen:
```
notify_user_gui() {
USER=$(occ_command user:list | awk '{print $3}';)
for user in $USER
do
    if [[ $(occ_command user:info $user)  == *"- admin"* ]]
    then
        args+="$user "
    fi
done
NCADMIN="${args[@]}"
for admin in $NCADMIN
do
    occ_command notification:generate -l "$2" "$admin" "$1"
done
}
```
This should work

szaimen:
I guess we have to test the function with admin user ids that contain spaces. (If that is possible)

enoch85:
Let's close this in favor of: https://github.com/nextcloud/vm/pull/1030

PR #891, opened 2019-07-18
improve check_open_port()

enoch85:


enoch85:
Tested and works as expected!

Issue #904, opened 2019-07-30
Change to i-colors everywhere.


enoch85:
Right now, some colors for `print_text_in_color()` are not with the "i"-color. Find and replace all the colors using the i-color instead of just plain Red, Cyan, Green.

zalonic:
Hi @enoch85, long time no commit.

I have opened a Pull Request to adjust the colouring scheme to High Intensity for all the remaining standard colours.

PR #906, opened 2019-08-13
High Intensity Colours Everywhere

zalonic:
Replaced all standard colours with there High Intensity counterparts.
Also removed a few redundant `{ }` braces.

RE: Change to i-colors everywhere. #904


enoch85:
Thanks a ton @zalonic!

Issue #917, opened 2019-08-24
Re-evaluate apcu for memcache.local


kesselb:
Please re-evaluate apcu for memcache.local: https://docs.nextcloud.com/server/16/admin_manual/configuration_server/caching_configuration.html#recommendations-based-on-type-of-deploymentre-evaluate

enoch85:
Why? :)

enoch85:
I can also add that the VM was using APCu from the get go, and it was replace by Redis which has worked extremely well.

kesselb:
Apcu is a php module and therefore available with low performance overhead. Redis is a external service and for each request a connection is required. There are some more explanations at the doc.

enoch85:
Haven't checked, but would it be possible to run both APCu and Redis at the same time?

kesselb:
> Haven't checked, but would it be possible to run both APCu and Redis at the same time?
> 
> APCu for local and Redis for distributed.

Yes. It's the recommended setup for "Organizations with single-server and clustered setups". But if the current setup works well for your users don't change it. Never change a running system :smile: I just wanted to mention it.

enoch85:
I will think about it.

Worth testing to implement at least and see if it works well. If that's the case I'll put it in production. :rocket:

enoch85:
Naah, seems like a "no go" for this. Did my usual stress test with the desktop client and:
1. Got ALOT of locked files
2. The desktop client didn't even upload all files.
3. Seems like APCu crashed.

Also saw that there were some issues when using PHP-FPM (which we do in the VM), so I'm not convinced yet. Maybe you can help me on this @kesselb?

kesselb:
Anything in the logs (Nextcloud, PHP)?

enoch85:
OK,[ I'm convinced](https://github.com/owncloudarchive/pi-image/issues/24), let's make this work!

Will do some more testing with the defaults.

enoch85:
For the tuning part: https://markus-blog.de/index.php/2018/05/30/how-to-install-apcu-object-cache-in-php7-and-integrate-in-nextcloud/

enoch85:
Check stats with this: https://raw.githubusercontent.com/krakjoe/apcu/master/apc.php

github-tomster:
hello world

is it just me:
```
PHP modules
This instance is missing some recommended PHP modules.
For improved performance and better compatibility it is highly recommended to install them:
- apcu
```
or 🤷🏻‍♂️ 

[...]

PR #918, opened 2019-08-24
add APCu

enoch85:
Needs extensive testing before merging.

cc @kesselb as you proposed it.

Just grabbed some random config, and that doesn't seem to work. Maybe something totally different is optimal for Nextcloud?

Check here for further tweaks: https://www.php.net/manual/en/apc.configuration.php

If merged, it's a fix for https://github.com/nextcloud/vm/issues/917

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -434,7 +434,7 @@ run_static_script redis-server-ubuntu
 
 # Enable igbinary for PHP 
 # https://github.com/igbinary/igbinary
-if is_this_installed pecl
+if is_this_installed "$PHPVER"-dev
 then
     if ! yes no | pecl install -Z igbinary
     then
@@ -453,7 +453,7 @@ restart_webserver
 fi
 
 # APCu (local cache)
-if is_this_installed pecl
+if is_this_installed "$PHPVER"-dev
 then
     if ! yes no | pecl install -Z apcu
     then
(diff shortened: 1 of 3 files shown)


kesselb:
I would go with the [default configuration](https://www.php.net/manual/en/apcu.configuration.php) as a start. We can always tweak it later.

enoch85:
@kesselb Ready for testing. :)

enoch85:
Been testing it for a while now in all possible ways. I feel confident. Added to both my clouds, and it seems faster in thumbnail loading.

**MERGE**

kesselb:
Where can I grap the latest vm?

enoch85:
> Where can I grap the latest vm?

I can make one for you :)

enoch85:
@kesselb Here you go: https://cloud.hanssonit.se/s/zjsqkrSpzqJGE9N?path=%2FTESTING 

I kept both versions so that you can compare the difference. Would be interesting if you did some real benchmarking.

enoch85:
@kesselb Did you have any time to test yet?

kesselb:
Unfortunately not. "Nextcloud_VM_www.hanssonit.se_17.0.0b2_APCu_TESTING" does not come with apcu :thinking:

enoch85:
Hmm, need to make a new release.

enoch85:
@kesselb New release is out! Same URL.

PR #923, opened 2019-09-01
warn user about passwords change

enoch85:
I get many emails regarding that passwords are wrong, warn the user about whats going on, and give the option to change passwords later on.


enoch85:
This kind of contradicts changing the password at all....

Issue #924, opened 2019-09-01
ask_yes_or_no: command not found


kesselb:
![VirtualBox_vm with apcu_01_09_2019_18_47_05](https://user-images.githubusercontent.com/3902676/64079554-0a136400-cce9-11e9-8bee-a49958628f51.png)

Testing box with apcu from the apcu pull request.

enoch85:
That tells me you lost internet connectivity as it fetches all the variables from this Github repo.

I tried that script yesterday, and it worked without a hitch.

Thanks for the heads up though.

kesselb:
Yes. The box was without internet. But this is the dialog to configure the network.

enoch85:
Best way to test is to download the latest beta from the link I gave you in another issue. Testing from branches won't work as the curl command will be in master anyway.

I will do one more test just to check.

kesselb:
OK. I was using the box "Nextcloud_VM_www.hanssonit.se_17.0.0b2_APCu_TESTING".

enoch85:
I see what's wrong here, I'm fixing that one now. Thanks for this issue!

enoch85:
Solved in 17.0.0

PR #926, opened 2019-09-01
add all the variables here in case the user doesn't have internet

enoch85:
cc @kesselb

Fix for https://github.com/nextcloud/vm/issues/924

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -607,6 +607,7 @@ do
         "Static IP")
             clear
             run_static_script static_ip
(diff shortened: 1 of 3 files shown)


enoch85:
@kesselb Please test this without internet, and see if everything checks out OK. 

Thanks!

enoch85:
Sorry, this is not done yet. I have a better solution.

enoch85:
OK, so this will start working in the next release (17.0.0) since we need lib.sh locally for it to work. cc @kesselb

enoch85:
Tested and works as expected! Will start working in 17.0.0.

PR #930, opened 2019-09-02
the first setup can only be run once

enoch85:


enoch85:
@kesselb Don't know how to reach you, so I write here :)<

17.0.0B4 is just released.  as a VM. Please test APCu and stuff like we talked about before.

PR #941, opened 2019-09-13
Missing `)` on line 512 added

BetaLeaf:
Cosmetic Fix
  - Missing `)` on line 512 added.


enoch85:
Thanks you very much for this PR! It's always nice with some extra eyes.

BetaLeaf:
No problem. Thank you for providing an easy to use script to set everything up with all the bells and whistles.

PR #959, opened 2019-10-03
Fixed issue where HTTP to HTTPS redirect leaves out the / and ultimately fails to work correctly

BetaLeaf:
Currently, when redirecting from HTTP to HTTPS, it fails if there is anything after $domain because it does not add back the `/`. 
Example: `http://example.com/apps/files/` redirects to `http://example.comapps/files/`
This is wrong!

I fixed it by adding the `/` on line 120.

Thanks for your time and consideration and have a great day!

### DIFF
--- a/lets-encrypt/activate-ssl.sh
+++ b/lets-encrypt/activate-ssl.sh
@@ -117,7 +117,7 @@ then
     cat << SSL_CREATE > "$ssl_conf"
 <VirtualHost *:80>
     ServerName $domain
-    Redirect / https://$domain
+    Redirect / https://$domain/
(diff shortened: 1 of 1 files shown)


enoch85:
Thanks!

Please fix this in `nextcloud_install_production.sh` as well.

BetaLeaf:
There is nothing to fix in `nextcloud_install_production.sh` as it does not use Redirect because SSL has not yet be enabled 😉

enoch85:
Right, forgot. Merge!

BetaLeaf:
No problem.

Issue #982, opened 2019-11-18
Update form Nextcloudvm 17.0 to 17.1 via update.sh deletes redis and went wrong


horvan:
```bash
joachim@vmd44994:/var/scripts$ sudo bash update.sh
mysql-common set on hold.
mysql-client set on hold.
mysql-client-5.7 set on hold.
mysql-client-core-5.7 set on hold.
mysql-server set on hold.
mysql-server-5.7 set on hold.
mysql-server-core-5.7 set on hold.
mysql-sandbox set on hold.
mysql-source-5.7 set on hold.
mysql-testsuite set on hold.
mysql-testsuite-5.7 set on hold.
mysql-utilities set on hold.
mysql-workbench set on hold.
mysql-workbench-data set on hold.
mysqltcl set on hold.
mysqltuner set on hold.
mariadb-common set on hold.
mariadb-server set on hold.
mariadb-client set on hold.
mariadb-client-10.1 set on hold.
mariadb-client-core-10.1 set on hold.
mariadb-plugin-connect set on hold.
mariadb-server-10.1 set on hold.
mariadb-plugin-cracklib-password-check set on hold.
mariadb-plugin-gssapi-client set on hold.
mariadb-plugin-gssapi-server set on hold.
mariadb-plugin-mroonga set on hold.
mariadb-plugin-oqgraph set on hold.
mariadb-plugin-spider set on hold.
mariadb-plugin-tokudb set on hold.
mariadb-server-core-10.1 set on hold.
mariadb-test set on hold.
mariadb-test-data set on hold.
Reading package lists... Done
Building dependency tree
Reading state information... Done
Calculating upgrade... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
mysql-client was already not hold.
mysql-client-5.7 was already not hold.
mysql-client-core-5.7 was already not hold.
mysql-server was already not hold.
mysql-server-5.7 was already not hold.
mysql-server-core-5.7 was already not hold.
mysql-sandbox was already not hold.
mysql-source-5.7 was already not hold.
mysql-testsuite was already not hold.
mysql-testsuite-5.7 was already not hold.
mysql-utilities was already not hold.
mysql-workbench was already not hold.
mysql-workbench-data was already not hold.
mysqltcl was already not hold.
mysqltuner was already not hold.
Canceled hold on mysql-common.
If you want to upgrade MySQL/MariaDB, please run 'sudo apt update && sudo apt dist-upgrade -y'
[...]

PR #983, opened 2019-11-18
possible fix for #982

enoch85:


enoch85:
Not a good idea, since we can't confirm that everyone is running the PHP version set in $PHPVER

PR #984, opened 2019-11-19
stop postgresql before dump

enoch85:


enoch85:
Won't work since the DB is stopped.

Issue #670, opened 2018-09-28
Vagrant & Nextcloud


gjgd:
In the README it is written that a good way to support the development was to "Write scripts so that this can be installed with Vagrant or similar".

Not sure if it's the right place, I just want to advertise that I wrote a script to easily spawn an Ubuntu 16.04 VM with Nextcloud v14 installed using Vagrant. 

https://github.com/gjgd/vagrant-nextcloud

enoch85:
Great stuff!

What I'm looking for is a way to:
1. Create an Ubuntu 18.04.X
2. Download the install script from Github and run it with all the default (non-interactive)
3. When run exits with 0, then package it and export it as OVA through VMware.

Do you think that's possible?

enoch85:
Related: https://github.com/nextcloud/vm/issues/193

enoch85:
Just tested and the current scripts need to be adapted to Vagrant because Vagrant can't handle whiptails and such.

enoch85:
Let's keep this in the original issue.

xelarate86:
Could this be done with Packer? I can take a pass at it, perhaps

enoch85:
@xelarate86 Thank you very much for your offer! I've already tried in packed actually but it failed because the install script needs root permissions, and I never figured out how to do that in a "packer secure" manner.

The repo is private, but I can make it public if you decide to continue on this.

enoch85:
Uploaded the files here: https://cloud.hanssonit.se/s/p6EdBDB9nmpfXai

enoch85:
Looking at it again I think it was due to that it needed `sudo` from the `ncadmin` user, and there are a lot of prompts to be answered. The defaults should work though (e.g. `yes no | bash script.sh`) but I can't recall exactly what the issue was now to be honest.

Feel free to investigate.

enoch85:
> Just tested and the current scripts need to be adapted to Vagrant because Vagrant can't handle whiptails and such.

Duuh, reading this issue again I can see why I never gave it more effort. 

I don't know if Vagrant evolved, but maybe. :)

josephdpurcell:
I created https://github.com/gjgd/vagrant-nextcloud/issues/1 since that repo is out of date; nextcloud is now on v17.

Very simply I did:

```
vagrant init
vim Vagrantfile
# Change the box to `ubuntu/bionic64`
  config.vm.box = "ubuntu/bionic64"
# Add a public IP: you can either do this, a local IP, or port forward
[...]

PR #993, opened 2019-12-10
Update readme to feature vagrant script

josephdpurcell:
See https://github.com/nextcloud/vm/issues/670


enoch85:
@josephdpurcell Thanks for the PR, though I'm thinking we should build this ourselves, not use another repo for it. 

Please make a PR where you add a folder named "build" (or another smart word) and put the code there instead.

Issue #1001, opened 2019-12-27
Change the preview script to match nextcloudpi


szaimen:
Hello,

I recently saw the [blogpost](https://ownyourbits.com/2019/06/29/understanding-and-improving-nextcloud-previews/) by nachoparker concerning the previewgenerator.
Would be really nice if you could change the previewgenerator.sh script to match the settings, that nachoparker recommends:

![image](https://user-images.githubusercontent.com/42591237/71531826-bdcfdd80-28f0-11ea-9197-3a88858a448a.png)

Also an option to choose, if I want to install imagick or not would be really nice, because it is not a must have if you get the settings right.

Maybe also beeing able to choose the formats I want to enable for preview during the setup would be really appreciated.

Also I think 4 am is a better time to run the previewgenerator.

Beeing able to choose a user for that the cronjob is running would also be really nice.

enoch85:
Merged

PR #1002, opened 2019-12-27
Refactor and improve Previewgenerator

szaimen:
See: https://github.com/nextcloud/vm/issues/1001

steps:
- [x] change preview settings
- [x] let us choose the preview-filetypes
- [x] let us choose if we want to install imagick
- [x] run the previewgenerator at 4 am
- [x] define a nextcloud-user for which the cronjob ist running
- [x] be able to run the script a second time

### DIFF
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -33,7 +33,9 @@ msg_box "In the next step you can choose to install a package called imagick to 
 
 Please note that this will put your server at risk as imagick is known to have several flaws.
 
-You can check this issue to understand why: https://github.com/nextcloud/vm/issues/743"
+You can check this issue to understand why: https://github.com/nextcloud/vm/issues/743
+
+Note: if you choose not to install imagick, it will get removed now."
 if [[ "yes" == $(ask_yes_or_no "Do you want to install imagick?") ]]
 then
     # Install imagick
@@ -110,6 +112,10 @@ then
     clear      
 
 else
+    # remove imagick
+    apt purge php-imagick -y
+    apt purge libmagickcore-6.q16-3-extra -y
+    
     # Choose file formats fo the case when imagick is not installed.
     whiptail --title "Choose file formats" --checklist --separate-output "Now you can choose for which file formats you would like to generate previews\nSelect or unselect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4 \
     "PNG" ON \
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -70,7 +70,7 @@ fi
 msg_box "In the next step you can choose to install a package called imagick to speed up the generation of previews and get support for more filetypes.
 
 The currently supported filetypes are:
--PNG
+- PNG
 - JPEG
 - GIF
 - BMP
@@ -256,7 +256,7 @@ then
 else
     while true
     do
-    PREVIEW_USER=$(whiptail --inputbox "Enter the Nextcloud user for which you want to run the preview-generation" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    PREVIEW_USER=$(whiptail --inputbox "Enter the Nextcloud user for which you want to run the preview-generation" $WT_HEIGHT $WT_WIDTH 3>&1 1>&2 2>&3)
     if ! $(occ_command user:list | grep $PREVIEW_USER) | awk '{print $3}'
         then
             msg_box "It seems like the user you entered ($PREVIEW_USER) doesn't exist, please try again."
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -212,6 +212,7 @@ then
     if [[ "yes" == $(ask_yes_or_no "So do you want to choose a nextcloud-user?") ]]
     then
         nextcloud_user=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
+        export nextcloud_user
         if [[ "yes" == $(ask_yes_or_no "Is this correct? $nextcloud_user") ]]
         then
             break
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -255,14 +255,14 @@ then
 else
     while true
     do
-    PREVIEW_USER=$(whiptail --inputbox "Enter the Nextcloud user for which you want to run the preview-generation" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
-    if ! occ_command user:list | grep "$PREVIEW_USER" | awk '{print $3}'
+        PREVIEW_USER=$(whiptail --inputbox "Enter the Nextcloud user for which you want to run the preview-generation" "$WT_HEIGHT"$
+        if [ -z $(occ_command user:list | grep "$PREVIEW_USER" | awk '{print $3}') ]
         then
             msg_box "It seems like the user you entered ($PREVIEW_USER) doesn't exist, please try again."
         else
             break
         fi
-     done
+    done
      # Add crontab for $PREVIEW_USER
      crontab -u www-data -l | { cat; echo "0 4 * * * php -f $NCPATH/occ preview:pre-generate $PREVIEW_USER >> $VMLOGS/previewgenerator.log"; } | crontab -u www-data -
      touch "$VMLOGS"/previewgenerator.log
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -219,7 +219,8 @@ then
     occ_command config:system:set jpeg_quality --value="60"
     occ_command config:app:set preview jpeg_quality --value="60"
     
-    msg_box "In the last step you can define a nextcloud-user for which you want to run the preview-generation. "
+    msg_box "In the last step you can define a nextcloud-user for which you want to run the preview-generation. 
+    If you not choose to, it will run in the default mode for all nextcloud-users."
     if [[ "yes" == $(ask_yes_or_no "So do you want to choose a nextcloud-user?") ]]
     then
         nextcloud_user=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
@@ -236,7 +237,8 @@ then
             fi
         fi
     else
-        sleep 1
+        nextcloud_user=""
+        export nextcloud_user
     fi
     
     # Add crontab
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -212,7 +212,10 @@ then
     if [[ "yes" == $(ask_yes_or_no "So do you want to choose a nextcloud-user?") ]]
     then
         nextcloud-user=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
-        whiptail --ok-button Done --msgbox "Is this nextcloud-user correct? $nextcloud-user" 10 30
+        if [[ "yes" == $(ask_yes_or_no "Is this correct? $nextcloud-user") ]]
+        then
+            break
+        fi
     else
         sleep 1
     fi
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -211,9 +211,9 @@ then
     msg_box "In the last step you can define a nextcloud-user for which you want to run the preview-generation. "
     if [[ "yes" == $(ask_yes_or_no "So do you want to choose a nextcloud-user?") ]]
     then
-        nextcloud_user=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
-        export nextcloud_user
-        if [[ "yes" == $(ask_yes_or_no "Is this correct? $nextcloud_user") ]]
+        nextclouduser=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
+        export nextclouduser
+        if [[ "yes" == $(ask_yes_or_no "Is this correct? $nextclouduser") ]]
         then
             break
         fi
@@ -222,7 +222,7 @@ then
     fi
     
     # Add crontab
-    crontab -u www-data -l | { cat; echo "0 4 * * * php -f $NCPATH/occ preview:pre-generate $nextcloud_user >> /var/log/previewgenerator.log"; } | crontab -u www-data -
+    crontab -u www-data -l | { cat; echo "0 4 * * * php -f $NCPATH/occ preview:pre-generate $nextclouduser >> /var/log/previewgenerator.log"; } | crontab -u www-data -
     touch /var/log/previewgenerator.log
     chown www-data:www-data /var/log/previewgenerator.log
 
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -211,8 +211,8 @@ then
     msg_box "In the last step you can define a nextcloud-user for which you want to run the preview-generation. "
     if [[ "yes" == $(ask_yes_or_no "So do you want to choose a nextcloud-user?") ]]
     then
-        nextcloud-user=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
-        if [[ "yes" == $(ask_yes_or_no "Is this correct? $nextcloud-user") ]]
+        nextcloud_user=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
+        if [[ "yes" == $(ask_yes_or_no "Is this correct? $nextcloud_user") ]]
         then
             break
         fi
@@ -221,7 +221,7 @@ then
     fi
     
     # Add crontab
-    crontab -u www-data -l | { cat; echo "0 4 * * * php -f $NCPATH/occ preview:pre-generate $nextcloud-user >> /var/log/previewgenerator.log"; } | crontab -u www-data -
+    crontab -u www-data -l | { cat; echo "0 4 * * * php -f $NCPATH/occ preview:pre-generate $nextcloud_user >> /var/log/previewgenerator.log"; } | crontab -u www-data -
     touch /var/log/previewgenerator.log
     chown www-data:www-data /var/log/previewgenerator.log
 
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -46,30 +46,148 @@ install_and_enable_app previewgenerator
 # Run the first preview generation and add crontab
 if [ -d "$NC_APPS_PATH/previewgenerator" ]
 then
+
     # Enable previews (remove the # to enable the specific preview)
     occ_command config:system:set enable_previews --value=true --type=boolean
     occ_command config:system:set preview_libreoffice_path --value="/usr/bin/libreoffice"
-#    occ_command config:system:set enabledPreviewProviders 0 --value="OC\\Preview\\PNG"
-#    occ_command config:system:set enabledPreviewProviders 1 --value="OC\\Preview\\JPEG"
-#    occ_command config:system:set enabledPreviewProviders 2 --value="OC\\Preview\\GIF"
-#    occ_command config:system:set enabledPreviewProviders 3 --value="OC\\Preview\\BMP"
-#    occ_command config:system:set enabledPreviewProviders 4 --value="OC\\Preview\\XBitmap"
-#    occ_command config:system:set enabledPreviewProviders 5 --value="OC\\Preview\\MarkDown"
-#    occ_command config:system:set enabledPreviewProviders 6 --value="OC\\Preview\\MP3"
-#    occ_command config:system:set enabledPreviewProviders 7 --value="OC\\Preview\\TXT"
-#    occ_command config:system:set enabledPreviewProviders 8 --value="OC\\Preview\\Illustrator"
-#    occ_command config:system:set enabledPreviewProviders 9 --value="OC\\Preview\\Movie"
-#    occ_command config:system:set enabledPreviewProviders 10 --value="OC\\Preview\\MSOffice2003"
-#    occ_command config:system:set enabledPreviewProviders 11 --value="OC\\Preview\\MSOffice2007"
-#    occ_command config:system:set enabledPreviewProviders 12 --value="OC\\Preview\\MSOfficeDoc"
-#    occ_command config:system:set enabledPreviewProviders 13 --value="OC\\Preview\\OpenDocument"
-#    occ_command config:system:set enabledPreviewProviders 14 --value="OC\\Preview\\PDF"
-#    occ_command config:system:set enabledPreviewProviders 15 --value="OC\\Preview\\Photoshop"
-#    occ_command config:system:set enabledPreviewProviders 16 --value="OC\\Preview\\Postscript"
-#    occ_command config:system:set enabledPreviewProviders 17 --value="OC\\Preview\\StarOffice"
-#    occ_command config:system:set enabledPreviewProviders 18 --value="OC\\Preview\\SVG"
-#    occ_command config:system:set enabledPreviewProviders 19 --value="OC\\Preview\\TIFF"
-#    occ_command config:system:set enabledPreviewProviders 20 --value="OC\\Preview\\Font"
+
+    # Choose file formats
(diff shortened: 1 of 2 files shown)


szaimen:
@enoch85 I would like to get some early feedback on this. Is my first PR here on GitHub. :)

enoch85:
Thanks for your PR! <3

You need to fix the checks here: https://travis-ci.org/nextcloud/vm/builds/630331851?utm_source=github_status&utm_medium=notification

I will have a closer look when I'm back from vacation.

szaimen:
@enoch85 Travis should be fixed by now.

Actually I don't know how I could do this:
- let us choose the preview-filetypes
- define a nextcloud-user for which the cronjob ist running
Maybe do you have some example code which I can look at?

But this could also be something for another PR.

Of course this can wait till after your vacation.

apps/previewgenerator.sh
```diff
@@ -60,23 +74,32 @@ then
...
-    # Install needed dependencies
-    install_if_not ffmpeg
-    install_if_not libreoffice
-    install_if_not php-imagick
-    install_if_not libmagickcore-6.q16-3-extra
+    msg_box "In the last step you can choose to run the previewgenerator now. 
     
-    # Pre generate everything
-    occ_command preview:generate-all
+    Please note: If you choose to not doing it now, then you have to do it manually before the cronjob at the next day at 4 am can continue.
```
enoch85: Remove this, there shouldn't be an option *not* to run the pre-generate command.

apps/previewgenerator.sh
```diff
@@ -16,18 +16,32 @@ debug_mode
...
-if [[ "no" == $(ask_yes_or_no "Do you still want to continue?") ]]
+It can speedup the feel of Nextcloud by a lot."
+if [[ "no" == $(ask_yes_or_no "So do you want to install the previewgenerator?") ]]
 then
     exit
 else
     sleep 1
 fi
 
+msg_box "In the next step you can choose to install a package called imagick to speed up the generation of previews and get support for many filetypes.
```
enoch85: I'm not sure but doesn't the previewgenerator depend on imagick? In other words, if not installed previewgenerator will not function?

apps/previewgenerator.sh
szaimen: Done.

apps/previewgenerator.sh
szaimen: No, I am currently running the preview:generate-all command on my instance without imagick installed and it works nicely.

szaimen:
@enoch85 It is ready for review again. 
I know, that there are some mistakes now, but I don't know how to fix them. 
Maybe you could help me out after your vacation?

enoch85:
There are more errors that needs to be fixed: https://travis-ci.org/nextcloud/vm/builds/630885829?utm_source=github_status&utm_medium=notification

Please have a look.

enoch85:
Also, problem with *not* installing Imagick here is that if not all filetypes are supported, then I'm sure we'll get a bug report saying: Previewrgenerator for XYZ is not working" because the user didn't enable Imagick and expect everything to work anyway. Most users don't read documentation, or even the text in the `msg_box`, they just expect everything to work.

As this adds more complexity it's very important to be more clear on exactly what happens and why.

Example, if you come from Windows and never saw a console box before, you would be very unfamiliar with everything, also take in to consideration that you don't read the informatic boxes - what would the end result be in 6 months from now when you decide to add support for SVG (since you got better in Nextcloud and  now knows what the `config.php` file does) and you forgot to add Imagick, or didn't add it because "it wasn't forced / needed) at the time? Answer: a bug report, a frustrated user, and a worse VM experience because the user will think it was the VMs fault, and not her/his own decision 6 months ago.

What I'm trying to say, you have to think of all scenarios when designing this stuff. :)

szaimen:
> There are more errors that needs to be fixed: https://travis-ci.org/nextcloud/vm/builds/630885829?utm_source=github_status&utm_medium=notification
> 
> Please have a look.

Actually I am not an experienced programmer and currently don't know, how to fix Travis.

What I have done is copying the part from [here](https://github.com/nextcloud/vm/blob/9c73939ac395f0091e7ca33c73311eb869dfffcf/lets-encrypt/activate-ssl.sh#L63) downwards and replaced everything with "domain" in it through "user" and pasted it [here](https://github.com/nextcloud/vm/pull/1002/commits/d5aa95e577a694a98c5abc5bf2f3ea0ded9179c1).
What I wanted to do there is letting the user enter a nextcloud-user for which the cronjob and the preview-generation at the end is running.
But that is probably not what I have done with the code I have added.

So could you maybe give me some code examples on how I could do it right?
Would be really appreciated!

szaimen:
> Also, problem with _not_ installing Imagick here is that if not all filetypes are supported, then I'm sure we'll get a bug report saying: Previewrgenerator for XYZ is not working" because the user didn't enable Imagick and expect everything to work anyway. Most users don't read documentation, or even the text in the `msg_box`, they just expect everything to work.
> 
> As this adds more complexity it's very important to be more clear on exactly what happens and why.
> 
> Example, if you come from Windows and never saw a console box before, you would be very unfamiliar with everything, also take in to consideration that you don't read the informatic boxes - what would the end result be in 6 months from now when you decide to add support for SVG (since you got better in Nextcloud and now knows what the `config.php` file does) and you forgot to add Imagick, or didn't add it because "it wasn't forced / needed" at the time? Answer: a bug report, a frustrated user, and a worse VM experience because the user will think it was the VMs fault, and not her/his own decision 6 months ago.
> 
> What I'm trying to say, you have to think of all scenarios when designing this stuff. :)

I will try to find out which formats are not supported without imagick and then edit my PR to just allow the formats that are available depending on if you have chosen on installing imagick or not.

szaimen:
So, I tested the different file formats out:
The following is the case (I first enabled all formats):
- if you just install ffmpeg these formats work:
PNG
JPEG
GIF
BMP
MP3
TXT
Markdown
Movie

-if you then install imagick these additional formats work:
Photoshop
SVG
TIFF

-I couln't make the following formats work (also if imagick and libreoffice is installed):
HEIC
MSOffice2003
MSOffice2007
MSOfficeDoc
OpenDocument
PDF
Postscript
StarOffice
Font
XBitmap
Illustrator

So either they are no longer supported, or something is broken in the previewgenerator or somewhere else.
Actually when generating the previews, I always got this error for office documents (when imagick (with and without libreoffice) is installed):
image
That could be the reason why the previews for these formats don't get generated.

So I would then change my script to represent the working files.

enoch85:
> What I wanted to do there is letting the user enter a nextcloud-user for which the cronjob and the preview-generation at the end of the script is running.

All previews should be generated and that's done with the existing `www-data`. I don't see the point with doing it with a second Nextcloud user. Please enlighten me.

szaimen:
> > What I wanted to do there is letting the user enter a nextcloud-user for which the cronjob and the preview-generation at the end of the script is running.
> 
> All previews should be generated and that's done with the existing `www-data`. I don't see the point with doing it with a second Nextcloud user. Please enlighten me.

Sry for not beeing clear.
I didn't mean a system-user but a nextcloud-user (from the instance).
Because in the preview generation command you can define a user for which you want to run the command and cronjob:
https://github.com/rullzer/PreviewGenerator#previewgenerate-all-user-id

This can especially be useful, if you use groupfolders or the external storage app to mount folders to many nextcloud-users with a lot of files in it. Because then you mount all folders for one puppet-user and the folders then get scanned only one time and not for each user again which isn't really efficient.

But if you don't need to define a nextcloud-user for the cronjob, you can just skip this step.

szaimen:
> > > What I wanted to do there is letting the user enter a nextcloud-user for which the cronjob and the preview-generation at the end of the script is running.
> > 
> > 
> > All previews should be generated and that's done with the existing `www-data`. I don't see the point with doing it with a second Nextcloud user. Please enlighten me.
> 
> Sry for not beeing clear.
> I didn't mean a system-user but a nextcloud-user (from the instance).
> Because in the preview generation command you can define a user for which you want to run the command and cronjob:
> https://github.com/rullzer/PreviewGenerator#previewgenerate-all-user-id
> 
> This can especially be useful, if you use groupfolders or the external storage app to mount a few folders to many nextcloud-users with a lot of files in it. Because then you mount all folders to one puppet-user and the cronjob then runs for this single user and not for all users, which would result in these folders getting scanned a lot of times each time you run the cronjob.
> 
> So would be really nice to get this into the script, because if you don't need to define a nextcloud-user for the cronjob, you can just skip this last step.

I think I found a way to doing it right in my last commit.

apps/previewgenerator.sh
```diff
@@ -16,67 +16,218 @@ debug_mode
...
+    occ_command config:app:set previewgenerator widthSizes  --value="256 384"
+    occ_command config:app:set previewgenerator heightSizes --value="256"
+    occ_command config:system:set preview_max_x --value="2048"
+    occ_command config:system:set preview_max_y --value="2048"
+    occ_command config:system:set jpeg_quality --value="60"
+    occ_command config:app:set preview jpeg_quality --value="60"
     
-    # Add crotab
-    crontab -u www-data -l | { cat; echo "@daily php -f $NCPATH/occ preview:pre-generate >> /var/log/previewgenerator.log"; } | crontab -u www-data -
+    msg_box "In the last step you can define a nextcloud-user for which you want to run the preview-generation. "
```
enoch85: Please remove this option. If the user makes a typo, the pre-generate will never happen. Also, as it's a cronjob, it will run in the background and no one will notice as it's in the middle of the night.

If you *really* think this option is useful and want to add this feature, please make `$nextcloud_user` `www-data` per default, and rename the variable to `cronjob_user`. So that if nothing is chosen, use www-data.

Something like: "The default user for generating previews are www-data. There's no reason to change that expect in special edge cases. Please only change this if you know what you are doing!"

Adding complexity = more likely to get bugs. Keep it simple. :)

apps/previewgenerator.sh
```diff
@@ -16,67 +16,218 @@ debug_mode
...
-    # Install needed dependencies
-    install_if_not ffmpeg
-    install_if_not libreoffice
-    install_if_not php-imagick
-    install_if_not libmagickcore-6.q16-3-extra
-    
+
     # Pre generate everything
-    occ_command preview:generate-all
+    occ_command preview:generate-all $nextcloud_user
```
enoch85: We want to pre-generate for all users, not only a single user.

Please remove the specific user.

apps/previewgenerator.sh
```diff
@@ -16,67 +16,218 @@ debug_mode
...
+    occ_command config:system:set enable_previews --value=true --type=boolean
+    install_if_not ffmpeg
 fi
 
-# Install preview generator
-install_and_enable_app previewgenerator
+msg_box "In the next step you can choose to install a package called imagick to speed up the generation of previews and get support for more filetypes. 
 
-# Run the first preview generation and add crontab
-if [ -d "$NC_APPS_PATH/previewgenerator" ]
```
enoch85: Why was this check removed?

szaimen:
My PR is ready for review again. Travis is finally fixed.

apps/previewgenerator.sh
enoch85: Also, if you go for the second option, add a check with `occ_command user:list | grep $user` to check if the user exists. If not, then ask for input again.

Thanks.

apps/previewgenerator.sh
szaimen: I copied this to line 200, because I think there is the only command for which you really need the previewgenerator enabled.
If you want, I can reverse this change.

apps/previewgenerator.sh
```diff
@@ -16,67 +16,218 @@ debug_mode
...
+    "Movie" OFF \
+    "Photoshop" OFF \
+    "SVG" OFF \
+    "TIFF" OFF 2>results
     
+    while read -r -u 9 choice
+    do
+        case $choice in
+            PNG)
+                clear
```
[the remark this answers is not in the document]
szaimen: I removed all except for the first one.

apps/previewgenerator.sh
```diff
@@ -16,67 +16,218 @@ debug_mode
...
+    "MarkDown" ON \
+    "MP3" ON \
+    "TXT" ON \
+    "Movie" OFF 2>results
+
+    while read -r -u 9 choice
+    do
+        case $choice in
+            PNG)
+                clear
```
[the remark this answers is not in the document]
szaimen: I removed all except for the first one.

apps/previewgenerator.sh
szaimen: The thing is, that I don't want to change the cronjob-user but define a nextcloud-user for which I want to run the command as I explained here: https://github.com/nextcloud/vm/pull/1002#issuecomment-569841096
But I added the check. Thanks for the advice.

apps/previewgenerator.sh
szaimen: If a user is not defined, the previews get generated for all users.

apps/previewgenerator.sh
```diff
@@ -16,67 +16,209 @@ debug_mode
...
+    occ_command config:app:set preview jpeg_quality --value="60"
     
-    # Add crotab
-    crontab -u www-data -l | { cat; echo "@daily php -f $NCPATH/occ preview:pre-generate >> /var/log/previewgenerator.log"; } | crontab -u www-data -
+    msg_box "In the last step you can define a nextcloud-user for which you want to run the preview-generation. "
+    if [[ "yes" == $(ask_yes_or_no "So do you want to choose a nextcloud-user?") ]]
+    then
+        nextcloud_user=$(whiptail --inputbox "Enter the nextcloud-user for which you want to run the preview-generation" 10 30 3>&1 1>&2 2>&3)
+        export nextcloud_user
+        if [[ "yes" == $(ask_yes_or_no "Is this correct? $nextcloud_user") ]]
```
szaimen: @enoch85 btw: do we need this check? (Because what should happen if you press no?)

apps/previewgenerator.sh
```diff
@@ -16,67 +16,209 @@ debug_mode
...
+    done 9< results
+    rm -f results
+    clear
+fi
+
+# Enable additional previews (most likely not working; remove the # to enable the specific preview)
+#occ_command config:system:set preview_libreoffice_path --value="/usr/bin/libreoffice"
+#occ_command config:system:set enabledPreviewProviders 4 --value="OC\\Preview\\XBitmap"
+#occ_command config:system:set enabledPreviewProviders 8 --value="OC\\Preview\\Illustrator"
+#occ_command config:system:set enabledPreviewProviders 10 --value="OC\\Preview\\MSOffice2003"
```
enoch85: Doesn't MSOffice and PDF work if you install the libreoffice packages (which you now seems to have removed)?

apps/previewgenerator.sh
enoch85: It's not what you want, it's what everyone wants. :) This VM is downloaded several thousand times. Having a setting that can be confusing (e.g. why even run with a certain user? Not everyone have the same setup as you.) is asking for issue reports when users say: "I installed preview generator, but it doesn't work".

Please change so that `www-data` is default, add text to explain what happens if you change it, like you explained for me. www-data is default for every user, having a single user is *very much an edge case* which involves group folders and so on.

apps/previewgenerator.sh
enoch85: Better safe than sorry. :)

enoch85: I will improve everything once I have time. Just commenting the obvious now.

apps/previewgenerator.sh
szaimen: No, I have tested this out and it hasn't worked also when libreoffice was installed
but look here for details:
https://github.com/nextcloud/vm/pull/1002#issuecomment-569828067

enoch85: Aah ok, thanks for confirming. 

Maybe remove those lines and refer to the Nextcloud documentation instead to keep it clean.

apps/previewgenerator.sh
```diff
@@ -110,6 +112,10 @@ then
     clear      
 
 else
+    # remove imagick
+    apt purge php-imagick -y
```
enoch85: This is not needed as it's not installed in the VM. Good thinking though.

apps/previewgenerator.sh
szaimen: actually I don't know what you mean. The system user is actually set to www-data
and if not set the preview generation runs for all nextcloud-users as default.

look at line 243:
`crontab -u www-data -l | { cat; echo "0 4 * * * php -f $NCPATH/occ preview:pre-generate $nextcloud_user >> /var/log/previewgenerator.log"; } | crontab -u www-data -`

so if the $nextcloud_user variable is empty, nothing gets added. So then it runs for all nextcloud-users.

szaimen: I have tried to improve the explanation of this a bit. and make sure that it runs for all nextcloud-users, if no user is set up

apps/previewgenerator.sh
szaimen: if you run the script a second time it could happen though.

apps/previewgenerator.sh
szaimen: I moved it back to the start of the script.
So: Done.

apps/previewgenerator.sh
szaimen: I removed them and added the comment to the specific section (currently line 79 and 154)

apps/previewgenerator.sh
szaimen: If it is better for you, I could move this part into the if-else-then and completely separate it, depending on if you choose to select a nextcloud user or not.

szaimen: I've done this now.
Maybe it is a bit more clear now.

apps/previewgenerator.sh
enoch85: Then please check if the package exists first. Thanks.

szaimen: Have added the check.
So should be done.

enoch85:
I will take some time to test this now. If all is good I will merge this today if you don't have any other additions?

enoch85:
BUG: sed: -e expression #1, char 2: extra characters after command

szaimen:
> I will take some time to test this now. If all is good I will merge this today if you don't have any other additions?

No, I am quiet satisfied with the options and flow the new script brings with it.
So go ahead and merge it when you are satisfied too and if there are no bugs left. 👍

enoch85:
Have you tested this yourself? I'm asking because the whiptail doesn't seem to work, and never was.

What do you say about enabling all previews instead of choosing? I mean, why wouldn't you want to enable everything?

enoch85:
There are multiple bugs, I don't have more time now. 

Please have a look yourself.

szaimen:
> Have you tested this yourself? I'm asking because the whiptail doesn't seem to work, and never was.

whiptails are working now.

> What do you say about enabling all previews instead of choosing? I mean, why wouldn't you want to enable everything?

e.g. I don't like txt previews. They don't make much sense imo.
The same for markdown and mp3
... so having the choice is better imho.

apps/previewgenerator.sh
```diff
@@ -16,67 +16,258 @@ debug_mode
...
+        
+        # install needed dependency for movies
+        install_if_not ffmpeg
+        
+        # reset the preview formats
+        occ_command config:system:delete "enabledPreviewProviders"
+        
+        # reset the cronjob
+        print_text_in_color "$ICyan" "Resetting the cronjob for the preview-generation"
+        crontab -u www-data -l | grep -v 'preview:generate-all'  | crontab -u www-data -
```
szaimen: @enoch85 this doesn't remove anything in the moment.
do you have another idea what to change to make it work?

apps/previewgenerator.sh
```diff
@@ -16,67 +16,258 @@ debug_mode
...
+occ_command config:app:set previewgenerator widthSizes  --value="256 384"
+occ_command config:app:set previewgenerator heightSizes --value="256"
+occ_command config:system:set preview_max_x --value="2048"
+occ_command config:system:set preview_max_y --value="2048"
+occ_command config:system:set jpeg_quality --value="60"
+occ_command config:app:set preview jpeg_quality --value="60"
+
+msg_box "In the last step you can define a specific Nextcloud user for which will be the user that runs the preview-generation. 
+
+The default behavoiur (just hit [ENTER]) is to run with the system user 'www-data' which will generate previews for all users.
```
szaimen: hitting enter deosn't work here.
So I could remove that line?

apps/previewgenerator.sh
```diff
@@ -16,67 +16,258 @@ debug_mode
...
+    
     # Pre generate everything
     occ_command preview:generate-all
+else
+    while true
+    do
+    PREVIEW_USER=$(whiptail --inputbox "Enter the Nextcloud user for which you want to run the preview-generation" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    if ! occ_command user:list | grep "$PREVIEW_USER" | awk '{print $3}'
+        then
+            msg_box "It seems like the user you entered ($PREVIEW_USER) doesn't exist, please try again."
```
szaimen: the check somehow doesn't get triggered. 
do you have another idea how to do it?

apps/previewgenerator.sh
enoch85: No please leave it. I'll test everything more later.

apps/previewgenerator.sh
enoch85: Yeah, that's one of the bugs. I'll think about it. If you remove `awk print` you get another error.

apps/previewgenerator.sh
enoch85: When I tested it, it removed everything. Using `sed` isn't a clean solution imo.

apps/previewgenerator.sh
enoch85: The most "secure" option would be to use `read` as we do in the other scrips.

apps/previewgenerator.sh
szaimen: Hm...

apps/previewgenerator.sh
enoch85: Enter choose the default - `www-data`.

apps/previewgenerator.sh
enoch85: It works as expected.

apps/previewgenerator.sh
enoch85: This is fixed.

enoch85:
OK, I'm happy about this now in general. Maybe some smaller fixes, like wording and similar, but code wise I think it looks good. :+1:

apps/previewgenerator.sh
szaimen: Okay

szaimen:
> OK, I'm happy about this now in general. Maybe some smaller fixes, like wording and similar, but code wise I think it looks good.

Very Nice! 🎉

enoch85:
Nice work @szaimen! :+1: :rocket: 

Please test this thoroughly, if you don't find anything, I'll merge tomorrow.

apps/previewgenerator.sh
```diff
@@ -16,67 +16,256 @@ debug_mode
...
+occ_command config:system:set preview_max_x --value="2048"
+occ_command config:system:set preview_max_y --value="2048"
+occ_command config:system:set jpeg_quality --value="60"
+occ_command config:app:set preview jpeg_quality --value="60"
+
+msg_box "In the last step you can define a specific Nextcloud user for which will be the user that runs the preview-generation. 
+
+The default behaviour (just hit [ENTER]) is to run with the system user 'www-data' which will generate previews for all users. 
+
+If you on the other hand choose to use a specific user, previews will ONLY be generated for that specific user."
```
szaimen: I still think it is a bit confusing to mention the www-data user here since it is just important to run it for all nextcloud-users or only for a single one imho.
So maybe: 
`The befault behaviour (just hit [ENTER]) ist to generate previews for all users.`

szaimen:
> Nice work @szaimen! 👍 🚀
> 
> Please test this thoroughly, if you don't find anything, I'll merge tomorrow.

Thank you @enoch85!

Everything seems to work in my testing! So ready to merge imo. 🚀
But before, we could think one last time about this topic:
https://github.com/nextcloud/vm/pull/1002#discussion_r362350088

apps/previewgenerator.sh
szaimen: We could also think here about adding a bit more Nextcloud user 
So that it is really clear that we mean Nextcloud users here (is just mentioned at the beginning)

szaimen:
Maybe we could also mention somewhere that the script can run more than once to change some settings  later on?

apps/previewgenerator.sh
```diff
@@ -16,67 +16,256 @@ debug_mode
...
+then
+    msg_box "At least PHP 7.X is supported, please upgrade your PHP version: https://shop.hanssonit.se/product/upgrade-php-version-including-dependencies/"
+    exit
+fi
 
-You can choose to cancel installing this in the next step."
-if [[ "no" == $(ask_yes_or_no "Do you still want to continue?") ]]
+# Encryption may not be enabled
+if [ -n "$(occ_command app:list | grep encryption | awk '{print $3}')" ]
 then
```
szaimen: Is this working for you?
I haven't tested it myself.
And what about end to end encryption?
Shouldn't that get checked here, too?

szaimen: I guess my mistake. Should work with both, since they both have "encryption" in their name.

enoch85:
It works as expected, merging!

PR #1008, opened 2019-12-31
Update reference to line number of RAM check

dbolton:


enoch85:
Thanks alot! Welcome as a contributor to @nextcloud/vm!

Issue #1006, opened 2019-12-30
Let us discuss the remember_login_cookie_lifetime


szaimen:
I would like to discuss whether this should be the default in the VM:
https://github.com/nextcloud/vm/blob/9c73939ac395f0091e7ca33c73311eb869dfffcf/nextcloud_install_production.sh#L412

Maybe it could get transfered to the startup script? Because this is probably something you want to configure and therefore would fit inside the startup-script imho.

Or if you don't want to move it to the startup-script, maybe it could be available as an apps-script?

szaimen:
I thought about this a bit more and I think it would fit perfectly here: 
https://github.com/nextcloud/vm/blob/9c73939ac395f0091e7ca33c73311eb869dfffcf/nextcloud-startup-script.sh#L375
because it isn't really an additional app, but should be configurable during the initial setup.
Then if the value doesn't get change by the user during this step, it would be the default value (1800).

enoch85:
You could prepare a PR where the default is 1800 seconds, with a `msg_box` that describes what it does, and a value for changing it to preferred value.

FYI, the reason it's not 0 is that it makes the web UI upload stop as it's forcefully logging out the session after that period of time.

Try different scenarios first, then make a PR with your findings.

Thanks a lot!

szaimen:
Do you agree, that it fits here:
https://github.com/nextcloud/vm/blob/9c73939ac395f0091e7ca33c73311eb869dfffcf/nextcloud-startup-script.sh#L375
Or where should it be located inside the start up script?

PR #1012, opened 2020-01-02
Configure the cookie lifetime during the startup script

szaimen:
fix https://github.com/nextcloud/vm/issues/1006


szaimen:
@enoch85 Travis is fixed. 
Please give me some feedback on this.
(I have also already tested the part that I added)
But not in a complete new installation when running the startup script first...

szaimen:
I moved the script to additional apps, because I think it fits better there.

apps/cookielifetime.sh
```diff
@@ -0,0 +1,61 @@
...
+# Check for errors + debug code and abort if something isn't right
+# 1 = ON
+# 0 = OFF
+DEBUG=0
+debug_mode
+
+# Check if root
+root_check
+
+COOKIE_LIFETIME=$(whiptail --radiolist  "Configure after what time (in seconds) after every Login every Nextcloud user gets logged out in the Browser\nSelect one with the [ARROW] Keys and the [SPACE] key and confirm by pressing [ENTER]" "$WT_HEIGHT" "$WT_WIDTH" 4 \
```
szaimen: @enoch85 would you say a --menu fits better here?

apps/cookielifetime.sh
```diff
@@ -0,0 +1,61 @@
...
+"1800s" "half an hour" ON \
+"7200s" "two hours" OFF \
+"43200s" "half a day" OFF \
+"172800s" "two days" OFF \
+"604800s" "one week" OFF \
+"2419200s" "four weeks" OFF \
```
szaimen: what do you think about showing the seconds here? shall I remove them?

apps/cookielifetime.sh
enoch85: Haven't had time to look at this yet. 

Do what you think is best, but be consequent with the other style. :+1:

apps/cookielifetime.sh
```diff
@@ -0,0 +1,61 @@
...
+elif [ "$COOKIE_LIFETIME" == "172800s" ]
+then 
+    occ_command config:system:set remember_login_cookie_lifetime --value="172800"
+elif [ "$COOKIE_LIFETIME" == "604800s" ]
+then 
+    occ_command config:system:set remember_login_cookie_lifetime --value="604800"
+elif [ "$COOKIE_LIFETIME" == "2419200s" ]
+then 
+    occ_command config:system:set remember_login_cookie_lifetime --value="2419200"
+elif [ "$COOKIE_LIFETIME" == "Custom" ]
```
szaimen: what would you say about the formatting here? is it readable?

apps/cookielifetime.sh
```diff
@@ -0,0 +1,61 @@
...
+then 
+    while true
+    do
+        COOKIE_LIFETIME=$(whiptail --inputbox "Please enter the Cookie Lifetime in seconds, so e.g. 1800 for half an hour or 3600 for an hour\nIt is not recommended to set it to less than hafl an hour!" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+        COOKIE_LIFETIME=${COOKIE_LIFETIME//[!0-9]/}
+        if [ "$COOKIE_LIFETIME" -lt "1800" ]
```
szaimen: what would you say should be the lowest value here?

nextcloud-startup-script.sh
```diff
@@ -493,11 +494,16 @@ do
             clear
             run_app_script talk
         ;;
-
+	
+        CookieLifetime)
+            clear
+            run_app_script cookielifetime
```
enoch85: Put it in static instead. This is no app. :)

apps/cookielifetime.sh
enoch85: The lowest value should be the default set before this PR came alive.

People will get issues with uploads on slow connections of large files otherwise.

enoch85: Maybe add in the info that the user will get forcefully logged out even if a upload is in progress?

apps/cookielifetime.sh
szaimen: okay, then I will leave it this way.
(because in previewgenerator we have a radiolist.

apps/cookielifetime.sh
enoch85: Make sure the user puts a **number** between 1800 and ---> infinity.

If not a number, msg_box in a while loop. Thanks!

If a value is wrong in the config file Nextcloud will freak out. ;)

nextcloud-startup-script.sh
szaimen: is moved over.

apps/cookielifetime.sh
szaimen: I have already done sure that it is not below 1800 seconds, which is the default value before this PR came alive. Also I already have made sure that it is always a number (the algorithm deletes everything, which isn't 0-9)

szaimen: > Maybe add in the info that the user will get forcefully logged out even if a upload is in progress?

Does this happen each time you change the value?
Or just if the time is up?

nextcloud-startup-script.sh
```diff
@@ -376,7 +376,8 @@ clear
 whiptail --title "Extra configurations" --checklist --separate-output "Choose what you want to configure\nSelect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Security" "(Add extra security based on this http://goo.gl/gEJHi7)" OFF \
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
-"Automatic updates" "(Automatically update your server every week on Sundays)" OFF 2>results
+"Automatic updates" "(Automatically update your server every week on Sundays)" OFF \
+"CookieLifetime" "(Configure after what time every Nextcloud user gets logged out in the Browser)" OFF 2>results
```
szaimen: Does it fit here good? (below Security, Static IP and Automatic updates.)

enoch85: Yes, that's a better place for it.

szaimen: okay 👍

apps/cookielifetime.sh
enoch85: In the while loop, add a `msg_box` which explain what's expected.

Thanks!

apps/cookielifetime.sh
enoch85: Looks good to me, have to test this first. Maybe I change my mind later :)

apps/cookielifetime.sh
szaimen: @enoch85 any thoughts here?

apps/cookielifetime.sh
szaimen: I added a description here. 👍

static/cookielifetime.sh
```diff
@@ -0,0 +1,65 @@
...
+    occ_command config:system:set remember_login_cookie_lifetime --value="604800"
+elif [ "$COOKIE_LIFETIME" == "2419200s" ]
+then 
+    occ_command config:system:set remember_login_cookie_lifetime --value="2419200"
+elif [ "$COOKIE_LIFETIME" == "Custom" ]
+then 
+    while true
+    do
+        COOKIE_LIFETIME=$(whiptail --inputbox "Configure the logout time (in seconds) which will forcefully logout the Nextcloud user from the web browser when the timeout is reached.\nPlease enter the Cookie Lifetime in seconds, so e.g. 1800 for half an hour or 3600 for an hour\nIt is not recommended to set it to less than half an hour!" "$WT_HEIGHT" "$WT_WIDTH" 1800 3>&1 1>&2 2>&3)
+        COOKIE_LIFETIME=${COOKIE_LIFETIME//[!0-9]/}
```
enoch85: If you type "hi" here, the script bugs out.

That's what I meant when I said that we should add a check that only allow numbers.

szaimen: Oh yes, you are right. I will fix that later.

szaimen: `if [ "$COOKIE_LIFETIME" == "" ]`

apps/cookielifetime.sh
szaimen: @enoch85 any thoughts on this?

static/cookielifetime.sh
szaimen: I have fixed this. is working now.

apps/cookielifetime.sh
szaimen: everything in this review should be done. 👍

szaimen:
This is ready again and should work reliably now.

enoch85:
Just answered all my 100's of emails. Will have a look at this now. :)

static/cookielifetime.sh
```diff
@@ -51,7 +51,7 @@ then
     do
         COOKIE_LIFETIME=$(whiptail --inputbox "Configure the logout time (in seconds) which will forcefully logout the Nextcloud user from the web browser when the timeout is reached.\nPlease enter the Cookie Lifetime in seconds, so e.g. 1800 for half an hour or 3600 for an hour\nIt is not recommended to set it to less than half an hour!" "$WT_HEIGHT" "$WT_WIDTH" 1800 3>&1 1>&2 2>&3)
         COOKIE_LIFETIME=${COOKIE_LIFETIME//[!0-9]/}
-        if [ "$COOKIE_LIFETIME" -lt "1800" ]
+        if [ "$COOKIE_LIFETIME" < "1800" -o "$COOKIE_LIFETIME" == "" ]
```
enoch85: is the `-o` instead of `||` as in "or" ?

In that case, please use `||` instead. Easier to read.

Thanks!

szaimen: I already had to fix this to fix travis...

szaimen:
one second before mergin pls

enoch85:
OK, I'm done. Ready to mrege?

szaimen:
> OK, I'm done. Ready to mrege?

just one thing I wanted to add

enoch85:
Remove it please. :)

szaimen:
okay, then I am ready 👍

szaimen:
actually your test doesn't work... i have tested it

enoch85:
Woop woop! 

> actually your test doesn't work... i have tested it

Test again ;) The function didn't exist in this master before the merge.

szaimen:
> Woop woop!
> 
> > actually your test doesn't work... i have tested it
> 
> Test again ;) The function didn't exist in this master before the merge.

ah yes you are probably right

enoch85:
Great job! :rocket:

szaimen:
> Great job! 🚀

the test works, but I am stuck here: 
image
Because I cannot enter a number again.
Seems like doing this in two loops isn't working...

enoch85:
Hmm, back to the drawing board.

PR #1015, opened 2020-01-06
Refactor the way TLS certs are generated + make all options (standalone, dns) available everywhere

liao20081228:
Hi Daniel ：

Yesterday I raised an issue in the repository: my  ISP blocked port detection caused port verification and certificate verification to fail. After that, I tried to solve it and succeeded.

Thanks.


enoch85:
Thanks for your PR! :rocket: 

I will have a look at this when I have time.

enoch85:
I read through it quickly now, and I think it would be better to somehow use the same code for both the regular Let's Encrypt script and the `$SUDBOMAIN` function. That way, changes would only happen in one file, and not multiple places = easier to maintain.

I'll think of a solution. Maybe you can do the same?

What this PR is about essentially is making DNS (and the other tries) available for the scripts that needs TLS.

liao20081228:
> I read through it quickly now, and I think it would be better to somehow use the same code for both the regular Let's Encrypt script and the `$SUDBOMAIN` function. That way, changes would only happen in one file, and not multiple places = easier to maintain.
> 
> I'll think of a solution. Maybe you can do the same?
> 
> What this PR is about essentially is making DNS (and the other tries) available for the scripts that needs TLS.



> I read through it quickly now, and I think it would be better to somehow use the same code for both the regular Let's Encrypt script and the `$SUDBOMAIN` function. That way, changes would only happen in one file, and not multiple places = easier to maintain.
> 
> I'll think of a solution. Maybe you can do the same?
> 
> What this PR is about essentially is making DNregular Let's Encrypt script aS (and the other tries) available for the scripts that needs TLS.


I got it. But what is the regular Let's Encrypt script? is it active_ssl.sh?

enoch85:
> But what is the regular Let's Encrypt script? is it active_ssl.sh?

Yes, that's correct. My first thought was to just call that script for SUBDOMAINS as well, but it wouldn't work as the Apache config is different. 

So maybe something like:
1. One script for main domain (activate-ssl.sh - name should be cha)
2. One script for subdomain (activate-subdomain.sh)
3. Make the generation of the TLS a `function`, with a variable that can be set depending on which script that is run. Like `if $1 -eq $subdomain; then`

OR:

1. Make an `if fucntion` in the main script, and only set configs if the it's the TLS cert for the main domain. Harder to do, but much cleaner IMO. If functions are changed everything happens in one script **only**.

The Apache config is already set in the app scripts, since they are different from each other. So the only thing needed really would be the TLS part, and the test (test-new-config.sh) but with values that match the app files.

liao20081228:
> > But what is the regular Let's Encrypt script? is it active_ssl.sh?
> 
> Yes, that's correct. My first thought was to just call that script for SUBDOMAINS as well, but it wouldn't work as the Apache config is different.
> 
> So maybe something like:
> 
> 1. One script for main domain (activate-ssl.sh - name should be cha)
> 2. One script for subdomain (activate-subdomain.sh)
> 3. Make the generation of the TLS a `function`, with a variable that can be set depending on which script that is run. Like `if $1 -eq $subdomain; then`
> 
> OR:
> 
> 1. Make an `if fucntion` in the main script, and only set configs if the it's the TLS cert for the main domain. Harder to do, but much cleaner IMO. If functions are changed everything happens in one script **only**.
> 
> The Apache config is already set in the app scripts, since they are different from each other. So the only thing needed really would be the TLS part, and the test (test-new-config.sh) but with values that match the app files.

 My idea is to place this part code， which use to get a TLS cert，  into a new special file named "....../lets-encrypt/genarate_cert.sh". Then delete them in acctive_ssl.sh and other scripts instead of  "curl -sL...." or a function.

enoch85:
> My idea is to place this part code， which use to get a TLS cert， into a new special file named "....../lets-encrypt/genarate_cert.sh". Then delete them in acctive_ssl.sh and other scripts instead of "curl -sL...." or a function.

Make a POC and then we'll discuss. :+1: 

Thanks!

lib.sh
```diff
@@ -434,27 +434,48 @@ fi
...
 
 # Check if port is open # check_open_port 443 domain.example.com
 check_open_port() {
 print_text_in_color "$ICyan" "Checking if port ${1} is open with https://ports.yougetsignal.com..."
-install_if_not curl
+install_if_not curl 
 # WAN Adress
 if check_command curl -s -H 'Cache-Control: no-cache' 'https://ports.yougetsignal.com/check-port.php' --data "remoteAddress=${WANIP4}&portNumber=${1}" | grep -q "is open on"
-then
+then
```
enoch85: Please remove spaces like this. Try to keep the code clean. 

Thanks! :+1:

lets-encrypt/genarate_cert.sh
```diff
@@ -0,0 +1,117 @@
...
+then
+    printf "%b" "${ICyan}It seems like no certs were generated, we will do 1 more try.\n${Color_Off}"
+    any_key "Press any key to continue..."
+#elif [ "$method" == "tls-alpn-01" ]
+#then
+#    printf "%b" "${ICyan}It seems like no certs were generated, we will do 1 more try.\n${Color_Off}"
+#    any_key "Press any key to continue..."
+elif [ "$1" == "dns" ]
+then
+    printf "%b" "${IRed}It seems like no certs were generated, please check your DNS and try again.\n${Color_Off}"
```
enoch85: use `print_text_in_color` instead. Please change in the original script as well if you find any occurrences like this.

lets-encrypt/genarate_cert.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+# Methods
```
enoch85: You need to add the default header lines here. Thanks!

```
#!/bin/bash

# T&M Hansson IT AB © - 2019, https://www.hanssonit.se/

true
# shellcheck source=lib.sh

# Check for errors + debug code and abort if something isn't right
# 1 = ON
# 0 = OFF
DEBUG=0
debug_mode

# Check if root
root_check
```

lets-encrypt/genarate_cert.sh
liao20081228: > use `print_text_in_color` instead. Please change in the original script as well if you find any occurrences like this.

I am sorry，most of code is copied from original scripts 。I will do it as your comments when I complete all work， I will notify you 。

lib.sh
liao20081228: Where are the spaces you said?Do you mean the "then" on a separate line and not fllow “if”？

enoch85: No, there are spaces in the formatting. 

Open the file in a console and you'll see.

liao20081228:
@enoch85  Hi，Daniel：All the work has been done，But there is a problem that the whole Code no uniform format：Whether the first line of code in the function should be indented？ And the “do” and  “then” should place at the end of the line or the start？I found a lot of two styles of code in the source code, even in the same function. they make me tangled.

lib.sh
```diff
@@ -222,9 +223,9 @@ domain_check_200() {
     else
         print_text_in_color "$IRed" "DNS lookup failed with nslookup."
         print_text_in_color "$IRed" "Please check your DNS settings! Maybe the domain isn't propagated?"
-	print_text_in_color "$ICyan" "Please check https://www.whatsmydns.net/#A/${1} if the IP seems correct."
+        print_text_in_color "$ICyan" "Please check https://www.whatsmydns.net/#A/${1} if the IP seems correct."
         nslookup "${1}" $DNS1
-        return 1
+        exit 1
```
enoch85: Are there any special reason you changed this?

Usually we send return instead of exit to answer a question in a function that is used. Exit will exit the whole thing, and that's not what we want if the user are running the first setup script for example.

lib.sh
```diff
@@ -430,52 +431,28 @@ else
...
 elif check_command curl -s -H 'Cache-Control: no-cache' 'https://ports.yougetsignal.com/check-port.php' --data "remoteAddress=${2}&portNumber=${1}" | grep -q "is open on"
-then  
+then
     print_text_in_color "$IGreen" "Port ${1} is open on ${2}!"
-else  
+else
     msg_box "it seems like you port $1 is not open.\nThis can happen when your ISP or upstream GATEWAY banned port scan/detecting.\nPlease check your port status and make sure it is open"
-    if [[ "no" == $(ask_yes_or_no "Make sure you port $1 is open?") ]]
-    then  
+    if [[ "no" == $(ask_yes_or_no "Are you sure you port $1 is open?") ]]
```
enoch85: Nice catch!

enoch85:
The genreal style is:

if bla bla
then
[4 spaces] if bla bla
[4 spaces] then
[4 spaces] [4 spaces] do stuff
[4 spaces] [4 spaces] fi
[4 spaces] fi

So always use 4 spaces in `if arguments` and similar. 

Thanks!

lib.sh
liao20081228: should the process stop immediately， but when a domain can't be accessed。I make some tests that I type the error domain， some strange things appeared.

enoch85: Hmm, I don't understand?

liao20081228: > Hmm, I don't understand?

the function, domain_check_200(), is used to check if a domain can be accessed by using nslookup. when the command failed, it indicates that user types an error domain. So I think the process should be stopped by "exit" instead of "return".  if using return, the process will continue to execute   "dig....", and the screen will show: "DNS seems correct when checking with dig!".

liao20081228: > Hmm, I don't understand?
The reason I said above is incorrect。My real reason is as follows：
The function, domain_check_200(), is used to check if a domain can be accessed by using `nslookup`. when the command failed, the function will be return directly don't execute the following code  that is :
```
 DIG="$(dig +short "${1}" @resolver1.opendns.com)"                                                                                                                                                                                                                 
 if [[ "$DIG" = "$WANIP4" ]]
 then
        print_text_in_color "$IGreen" "DNS seems correct when checking with dig!"
 elif [[ "$DIG" != "$WANIP4" ]]
... ...
.... ...
```

This causes the script that called the function to continue executing，and don‘t know the domain name is wrong.

enoch85: The reason why I want it to continue is that the checks are independent. Sometimes it works with DIG, and not with NSLOOKUP. That's why the two exist.

I also think that the whole thing in generate-cert.sh should be in `lib.sh` as there's where all the functions exists. Not a separate script. Please change.

apps/onlyoffice.sh
```diff
@@ -6,6 +6,7 @@
 true
 # shellcheck source=lib.sh
 NC_UPDATE=1 && OO_INSTALL=1 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
```
enoch85: remove this line please.

lets-encrypt/activate-ssl.sh
```diff
@@ -102,6 +102,7 @@ ssl_conf="/etc/apache2/sites-available/"$domain.conf""
 # DHPARAM
 DHPARAMS="$CERTFILES/$domain/dhparam.pem"
 
+
```
enoch85: remove this line please

enoch85:
One thing I think you forgot here is that we test if the TLS generation was OK with `test-new-config.sh`. Right now it seems like you aren't using the structure for the checks that were implemented before.

But maybe you have a better idea?

lets-encrypt/activate-ssl.sh
```diff
@@ -202,99 +202,5 @@ then
...
-
-This script is developed on GitHub, feel free to contribute:
-https://github.com/nextcloud/vm
-
-The script will now do some cleanup and revert the settings."
-
-# Cleanup
-apt remove certbot -y
-apt autoremove -y
-clear
```
enoch85: In my opinion, everything removed here should be put into a function instead "generate_cert" or whatever. 

Right now it seems like you divided everything into smaller pieces. Is there any thought behind that decision? Why is it better?

liao20081228:
> One thing I think you forgot here is that we test if the TLS generation was OK with `test-new-config.sh`. Right now it seems like you aren't using the structure for the checks that were implemented before.
> 
> But maybe you have a better idea?

I do the check in create_config()  in lib.sh。

lets-encrypt/activate-ssl.sh
liao20081228: Yeah, It could be to implement if they only use for getting TLS. I just think maybe it looks simple and

lib.sh
liao20081228: I will do another test using original script

enoch85:
OK, as we said before: do what you had in mind, then I'll review everything once you think you are done. :+1: I think then it will be easier to get the whole picture.

lib.sh
liao20081228: I have done the test,  Using the original script, no bad thing happen. when `nslookup` failed, the function return. Then,  the script will stop because of `check_open_port`.  

Now, I changed the behavior of `check_open_port`, if one types "yes" to answer the question which is "Are you sure that your port 80/443 is open ".  the script will continue to execute the following code. 

That's why I did that.

enoch85: Well, either the port is closed or it's open. Why add a question that's asking if it's really open?

The check happens externally anyway, and the TLS cert generation will fail if the port is closed. 

Do you see the logic here?

liao20081228: When the ISP banned the port detecting/scan from external, the User can't active SSL. In fact, their port is open,  but no response can be received from https://ports.yougetsignal.com/check-port.php.

enoch85: OK, add a `msg_box` instead, explaining that. Something like:

```
msg_box "It seems like the port [port] is closed. This could be because your ISP has blocked the port, or that the port isn't open.

If you are 100% sure the port is open you can now choose to continue. There are no guarantees that it will work anyway though, since Let's Encrypt depend on that the port is open and accessible from outside your network."

ask_yes_or_no [Are you 100% sure the port is open?] Yes/No
```

Thanks!

liao20081228:
@enoch85   I have done all the work and make a successful test. it seemed like it worked well.

lib.sh
```diff
@@ -225,7 +225,7 @@ domain_check_200() {
         print_text_in_color "$IRed" "Please check your DNS settings! Maybe the domain isn't propagated?"
         print_text_in_color "$ICyan" "Please check https://www.whatsmydns.net/#A/${1} if the IP seems correct."
         nslookup "${1}" $DNS1
-        exit 0
+        retrun 1
```
enoch85: there's a typo here

enoch85:
>  I have done all the work and make a successful test. it seemed like it worked well.

Thanks a lot! Did you test all possible scenarios, or just one scenario where the port is open?

**Test cases:**
1. Port is open, everything is working
2. DNS name is not valid, port is open
3. Port is open, DNS name is not valid
4. Port is open, but the checker thinks it's closed due to block
5. Anything else?

@nextcloud/vm @szaimen do you think you could help out testing this PR please?

enoch85:
My first thoughts just looking at the code:

1. The function should **only** handle TLS. Not `occ_command` for installing apps or similar. It would be better to check `if generate_cert then; install whatever`. The only function `generate_cert` should have, is the adding TLS, nothing else.
2. The indentation is wrong. Please back the first if argument to the first row.
3. `test-new-config.sh` could be developed for Onlyoffice/Collabora as well. Since we are doing this big change, why not do it properly and more generic? What I'm thinking is same way as `$domain` is used in the original code, do the same but with subdomain instead, and make it a generic $variable.
4. [DHPARAMS](https://github.com/nextcloud/vm/blob/master/lets-encrypt/activate-ssl.sh#L103) is not set.
5. [This](https://github.com/liao20081228/vm/blob/master/lib.sh#L440) should always be the same, not matter which TLS that are generated.
6. What happened to all the methods? We use different methods in the original script. It was easier to add new methods before this change. Please add that back.

Thanks for your efforts! Really amazing!

liao20081228:
> > I have done all the work and make a successful test. it seemed like it worked well.
> 
> Thanks a lot! Did you test all possible scenarios, or just one scenario where the port is open?
> 
> **Test cases:**
> 
> 1. Port is open, everything is working
> 2. DNS name is not valid, port is open
> 3. Port is open, DNS name is not valid
> 4. Port is open, but the checker thinks it's closed due to block
> 5. Port is closed, no DNS setup
> 6. Anything else?
> 
> @nextcloud/vm @szaimen do you think you could help out testing this PR please?




> My first thoughts just looking at the code:
> 
> 1. The function should **only** handle TLS. Not `occ_command` for installing apps or similar. It would be better to check `if generate_cert then; install whatever`. The only function `generate_cert` should have, is the adding TLS, nothing else.
> 2. The indentation is wrong. Please back the first if argument to the first row.
> 3. `test-new-config.sh` could be developed for Onlyoffice/Collabora as well. Since we are doing this big change, why not do it properly and more generic? What I'm thinking is same way as `$domain` is used in the original code, do the same but with subdomain instead, and make it a generic $variable.
> 4. [DHPARAMS](https://github.com/nextcloud/vm/blob/master/lets-encrypt/activate-ssl.sh#L103) is not set.
> 5. [This](https://github.com/liao20081228/vm/blob/master/lib.sh#L440) should always be the same, not matter which TLS that are generated.
> 6. What happened to all the methods? We use different methods in the original script. It was easier to add new methods before this change. Please add that back.
> 
> Thanks for your efforts! Really amazing!

I also considered this issue. IThat's exactly what I will do，because  I find the config in acctive_ssl.sh and in other app' scripts are different. it's the work of tomorrow.

As for another method， I find it was commented on. And in the document of certbot, there are two methods(dns or http-01).

enoch85:
> (dns or http-01).

Yes, right now, from my understanding [they are planning TLS-ALPN](https://github.com/certbot/certbot/issues/6724) as well. In any case I think the script should be more generic, and easy to add stuff later if needed. We are not coding for today, we are coding for tomorrow :)

enoch85:
OK, so I did a very quick test with a working domain, and it worked. 

I still think we need to address the points I wrote about before. Code needs to be cleaner, smarter, and better in general.

enoch85:
@liao20081228 Did you mean to close this?

enoch85:
Hmm, seems like you deleted your repo?

Too bad, I think this was shaping up to be quite nice! :D

Please try to restore your repo if possible.

liao20081228:
> @liao20081228 Did you mean to close this?

Yes, I will do another PR, because it has too many commits.

enoch85:
Aha ***pew***. Thought you made a mistake. :)

Issue #1018, opened 2020-01-06
Onlyoffice: if the cpu load spikes up while loading previews in Nextcloud, onlyoffice is not reachable


szaimen:
## Steps to reproduces
1. Install Nextcloud VM
2. install the onlyoffice server with the apps-script
3. open a folder in nextcloud with many pictures -> the preview files get loaded or generated -> the cpu load spikes up to 100%
4. onlyoffice is not reachable.

## Expected behaviour
Onlyoffice shoud work no matter what the rest of the VM is doing

## Actual behaviour
Onlyoffice is not reachable.

## Server configuration
I am running the latest Nextcloud VM

## Ideas how to solve it
Maybe the Nextcloud VM should reserve some ressources to the onlyoffice server to keep it running? (of course preview loading/generating is slower than, but it wouldn't crash the onlyoffice server)

enoch85:
This is probably an error with OnlyOffice. Did you try Collabora? Is it the same thing?

I prefer Collabora over Onlyoffice actually. It seems to me that Collabora is more stable. Almost every issue I get regarding the two, OnlyOffice is the one not working as expected.

szaimen:
I found this:
https://github.com/nextcloud/vm/issues/961#issuecomment-544225015
Disabling the evasive module also fixes the problem for me.
So I would suggest something like 
```
If [ ${Apache2ctl -M | grep evasive} != "" ]
then
    sudo a2dismod evasive
```
before installing onlyoffice

enoch85:
@szaimen That's only valid if you installed the ""Extra Security". But sure, we could add a check for that.

I'm quite busy now, new year, new budget, and a lot of work. Please make a PR and I'll review it when I have time.

Issue #961, opened 2019-10-06
Occasionally getting HTTP Error 403 on OnlyOffice. Happens more often when I haven't used it for a while. Happens more often when used frequently.


BetaLeaf:
Occasionally OnlyOffice will fail to load with HTTP Error 403.
It happens more often when OnlyOffice hasn't been used in a while
It happens more often when OnlyOffice is used frequently and rapidly

I will get either 
`ONLYOFFICE cannot be reached. Please contact admin` [log](https://github.com/nextcloud/vm/files/3694064/cloud.betaleaf.net-1570345027304-failed.cannot.be.reached.log) or 
`Connection is lost. You can still view the document, but it will not be able to download or print until the connection is restored.` [log](https://github.com/nextcloud/vm/files/3694066/cloud.betaleaf.net-1570343486084-failed.screenshot.log). 

When it does work, it loads with some issues 
[log](https://github.com/nextcloud/vm/files/3694067/cloud.betaleaf.net-1570343519603-working.log)

## Server configuration

**Nextcloud server version:** (see your admin page)
17

**Server OS** (Ubuntu server is default)
18.04.3 LTS

**How did you install the VM?** (Scripted install from master **OR** Released version)
Scripted install from master

## Network
**Do you use DHCP?**
No, Static

**Is port 80 and/or 443 open?**
Yes

#### Log file (/mnt/ncdata/nextcloud.log)
```
[nextcloud.log](https://github.com/nextcloud/vm/files/3694052/nextcloud.log)

```

#### Screenshots
![image](https://user-images.githubusercontent.com/7028351/66265375-50c21580-e7e3-11e9-9f7f-8c5be6aa05f0.png)
![image](https://user-images.githubusercontent.com/7028351/66265513-02624600-e7e6-11e9-80ac-febc7692a3fa.png)

[other_vhosts_access.log](https://github.com/nextcloud/vm/files/3694053/other_vhosts_access.log)
[error.log](https://github.com/nextcloud/vm/files/3694054/error.log)

BetaLeaf:
The inconsistency is driving me up a wall, can you help?

enoch85:
Try increasing the RAM. Does that help?

BetaLeaf:
I gave it 12GB. Shouldn't that be enough? That's 3x more than what the installer script says it needs.

enoch85:
Hmm, yeah that should be more than enough. 

[...]

PR #1020, opened 2020-01-07
Fix Onlyoffice

szaimen:
fix https://github.com/nextcloud/vm/issues/1018

### DIFF
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -49,7 +49,7 @@ sudo bash onlyoffice.sh"
 fi
 
 # check if apache2 evasive-mod is enabled and disable it because of compatibility issues
-if [ ${Apache2ctl -M | grep evasive} != "" ]
+if [ "$(Apache2ctl -M | grep evasive)" != "" ]
 then
     msg_box "Please note: if you continue, the ddos protection for webservices will get disabled now, because of compatibility issues with onlyoffice."
     if [[ "no" == $(ask_yes_or_no "So do you want to continue?")  ]]
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -54,10 +54,10 @@ then
     msg_box "We noticed that 'mod_evasive' is installed which is the DDOS protection for webservices. It has comptibility issues with OnlyOffice and you can now choose to disable it."
     if [[ "no" == $(ask_yes_or_no "Do you want to disable DDOS protection?")  ]]
     then
-        print_text_in_color "$Cyan" "Keeping mod_evasive active."
+        print_text_in_color "$ICyan" "Keeping mod_evasive active."
     else
         a2dismod evasive
-        a2dismod mod-evasive
+        # a2dismod mod-evasive # not needed, but existing in the Extra Security script.
         apt purge libapache2-mod-evasive -y
 	systemctl restart apache2
     fi
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -48,15 +48,16 @@ sudo bash onlyoffice.sh"
     exit 1
 fi
 
-# check if apache2 evasive-mod is enabled and disable it because of compatibility issues
-if [ "$(Apache2ctl -M | grep evasive)" != "" ]
+# Check if apache2 evasive-mod is enabled and disable it because of compatibility issues
+if [ "$(apache2ctl -M | grep evasive)" != "" ]
 then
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 Actually, I haven't tested this check yet, but in theory it should work ;)

apps/onlyoffice.sh
```diff
@@ -48,6 +48,18 @@ sudo bash onlyoffice.sh"
     exit 1
 fi
 
+# check if apache2 evasive-mod is enabled and disable it because of compatibility issues
+if [ "$(Apache2ctl -M | grep evasive)" != "" ]
+then
+    msg_box "Please note: if you continue, the ddos protection for webservices will get disabled now, because of compatibility issues with onlyoffice."
+    if [[ "no" == $(ask_yes_or_no "So do you want to continue?")  ]]
+        then
```
enoch85: Skip the "So" here, I noticed you add that on all questions. 

It's enough with just "Do you want to continue?" :)

apps/onlyoffice.sh
```diff
@@ -48,6 +48,18 @@ sudo bash onlyoffice.sh"
...
+            exit 1
```
enoch85: Just add `sleep 0.1` here, if you exit the whole script exits.

apps/onlyoffice.sh
enoch85: I will fix it later, let it be, there are more changes to be made.

enoch85:
@szaimen Please test if this works.

apps/onlyoffice.sh
```diff
@@ -48,6 +48,20 @@ sudo bash onlyoffice.sh"
...
+if [ "$(apache2ctl -M | grep evasive)" != "" ]
+then
+    msg_box "We noticed that 'mod_evasive' is installed which is the DDOS protection for webservices. It has comptibility issues with OnlyOffice and $
+    if [[ "no" == $(ask_yes_or_no "Do you want to disable DDOS protection?")  ]]
+        then
+            sleep 0.1
+        else
+            a2dismod evasive
+	    a2dismod mod-evasive
+            apt -y install libapache2-mod-evasive
```
szaimen: Axtually to my experiemce disabling 'a2dismod evasive' is enough. But don't know if mod-evasive also doesn't work then...

apps/onlyoffice.sh
szaimen: you have already done this 👍

apps/onlyoffice.sh
szaimen: you have already done this ..👍

szaimen:
> @szaimen Please test if this works.

Yes, the check works. (currently line 52)

I have just tested  "a2dismod evasive" and this already makes sure that onlyoffice works. So the other removals aren't really much needed to make onlyoffice work...

apps/onlyoffice.sh
enoch85: Yeah OK, if you say so. :) The module activated in the Extra Security script is called `mod-evasive`. That's why I added it.

apps/onlyoffice.sh
```diff
@@ -58,7 +58,7 @@ then
     else
         a2dismod evasive
         a2dismod mod-evasive
-        apt -y install libapache2-mod-evasive
+        apt purge libapache2-mod-evasive -y
```
enoch85: Haha, what a mistake! :D

enoch85:
> aren't really much needed to make onlyoffice work...

OK, great! only do what's necessary then. :+1:

apps/onlyoffice.sh
enoch85: It "will" not be disabled, the user can "choose" to disable it. There's a big difference.

Please change wording.

enoch85:
OK, let's merge this!

enoch85:
@szaimen FYI https://github.com/nextcloud/vm/issues/961#issuecomment-544281236

@BetaLeaf Do you think this will do the trick? Or do we need to disable `mod_evasive` as well?

szaimen:
@enoch85
As @BetaLeaf has written above that comment:
> Now I am not knowledgeable enough to figure out how to set the server configuration properly, so I just disabled mod_evasive entirely and that seems to solve the issue.
> 
> ```
> sudo a2dismod evasive
> sudo systemctl restart apache2
> ```

That was the reason why I came up with this.

szaimen:
I think 
> sudo a2dismod evasive
and
> sudo a2dismod mod_evasive 
Both disable mod_evasive.

enoch85:
Yeah, my bad. Please ignore.

Issue #1022, opened 2020-01-08
Configure and adding cifs-mounts as optional app-script?


szaimen:
I actually use CIFS-mounts very often to access files of my host machine from the VMs. And this could therefore be something useful for the VM. 

At least for script-testing it would be really nice, if it could get done through an appscript. But of course also if you don't want to store all userfiles inside the Volume. (external storage app). 

Also the Nextcloud-SMBclient seems to work very unreliably in the latest past (I couldn't add an SMB-Share on 17.0.2) and system mount is much more stable, reilable and fast in my experince. 
So could this be a useful addition for the Nextcloud VM?

What do you think about this?
I would like to work on this next if you would accept PRs on this topic.

enoch85:
Do you mean mounting SMB shares in the VM directly as a root file system?

szaimen:
> Do you mean mounting SMB shares in the VM directly as a root file system instead of `/mnt/ncdata`?

Maybe this or just as an additional path e.g. /mnt/data

enoch85:
I'm OK with adding *additional* SMB mounts as an option. I'm not OK with replacing `/mnt/ncdata`.

All mounts should happen in `/mnt/smbshares/` as root for all shares. So if there are multiple shares then it would look like:
`/mnt/smbshares/1`
`/mnt/smbshares/2`
`/mnt/smbshares/3`
and so on.

Thank you very much @szaimen!

szaimen:
Actually if you wuld want to use the smbmount with the external storage app, then you need to use the ww-data user as file owner amd directory owner... So I would prefer that...

enoch85:
> ww-data user as file owner and directory owner... So I would prefer that...

Yes, that's what I expected you to do. :)

PR #1024, opened 2020-01-09
Adding SMB-Shares

szaimen:
fix https://github.com/nextcloud/vm/issues/1022

### DIFF
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -25,17 +25,8 @@ then
     chmod 600 /etc/fstab
 fi
 
-# choose categories
-SMB_MOUNT=$(whiptail --title "SMB-Share" --radiolist  "This script let you manage SMB-Shares to access files from the host-computer or other machines in the local network.\nChoose what you want to do.\nSelect one with the [ARROW] keys and select with the [SPACE] key. Confirm by pressing [ENTER]" "$WT_HEIGHT" "$WT_WIDTH" 4 \
-"add a SMB-Mount" "(and mount/connect it)" ON \
-"mount SMB-Shares" "(connect SMB-Shares)" OFF \
-"show all SMB-Mounts" "" OFF \
-"unmount SMB-Shares" "(disconnect SMB-Shares)" OFF \
-"delete SMB-Mounts" "(and unmount/disconnect them)" OFF 3>&1 1>&2 2>&3)
-
-# add SMB-Mount
-if [ "$SMB_MOUNT" == "add a SMB-Mount" ]
-then
+# functions
+add_mount() {
     # check if mounting slots are available
     if [ -n "$(grep /mnt/smbshares/1 /etc/fstab)" ] && [ -n "$(grep /mnt/smbshares/2 /etc/fstab)" ] && [ -n "$(grep /mnt/smbshares/3 /etc/fstab)" ]
     then
@@ -109,9 +100,9 @@ then
         count=$(( $count + 1))
     done
     run_app_script smbmount 
-# mount smb-shares
-elif [ "$SMB_MOUNT" == "mount SMB-Shares" ]
-then
+}
+
+mount_shares() {
     # check if any smb-share is created
     if [ -z "$(grep /mnt/smbshares /etc/fstab)" ]
     then
@@ -148,9 +139,9 @@ then
         count=$(( count + 1))
     done
     run_app_script smbmount
-# show all smb-mounts
-elif [ "$SMB_MOUNT" == "show all SMB-Mounts" ]
-then
+}
+
+show_all_mounts() {
     # if no entry created nothing to show
     if [ -z "$(grep /mnt/smbshares /etc/fstab)" ]
     then
@@ -181,9 +172,9 @@ then
         count=$(( $count + 1))
     done
     run_app_script smbmount
-# unmount smb-shares
-elif [ "$SMB_MOUNT" == "unmount SMB-Shares" ]
-then
+}
+
+unmount_shares() {
     # check if any smb-shares are available for unmounting
     if [[ ! $(findmnt -M "/mnt/smbshares/1") ]] && [[ ! $(findmnt -M "/mnt/smbshares/2") ]] && [[ ! $(findmnt -M "/mnt/smbshares/3") ]]
     then
@@ -219,9 +210,9 @@ then
         count=$(( $count + 1))
     done
     run_app_script smbmount
-# delete smb-mounts
-elif [ "$SMB_MOUNT" == "delete SMB-Mounts" ]
-then
+}
+
+delete_mounts() {
     # check if any smb-share available
     if [ -z "$(grep /mnt/smbshares /etc/fstab)" ]
     then
@@ -262,6 +253,36 @@ then
         count=$(( $count + 1))
     done
     run_app_script smbmount
+}
+
+# Mainmenu
+SMB_MOUNT=$(whiptail --title "SMB-Share" --radiolist  "This script let you manage SMB-Shares to access files from the host-computer or other machines in the local network.\nChoose what you want to do.\nSelect one with the [ARROW] keys and select with the [SPACE] key. Confirm by pressing [ENTER]" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+"add a SMB-Mount" "(and mount/connect it)" ON \
+"mount SMB-Shares" "(connect SMB-Shares)" OFF \
+"show all SMB-Mounts" "" OFF \
+"unmount SMB-Shares" "(disconnect SMB-Shares)" OFF \
+"delete SMB-Mounts" "(and unmount/disconnect them)" OFF 3>&1 1>&2 2>&3)
+
+# add SMB-Mount
+if [ "$SMB_MOUNT" == "add a SMB-Mount" ]
+then
+    add_mount
+# mount smb-shares
+elif [ "$SMB_MOUNT" == "mount SMB-Shares" ]
+then
+    mount_shares
+# show all smb-mounts
+elif [ "$SMB_MOUNT" == "show all SMB-Mounts" ]
+then
+    show_all_mounts
+# unmount smb-shares
+elif [ "$SMB_MOUNT" == "unmount SMB-Shares" ]
+then
+    unmount_shares
+# delete smb-mounts
+elif [ "$SMB_MOUNT" == "delete SMB-Mounts" ]
+then
+    delete_mounts
 else
     sleep 1
 fi
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -20,9 +20,9 @@ root_check
 install_if_not cifs-utils
 
 # secure fstab
-if [ "$(ls -l /etc/fstab | awk '{print $1}')" != "-rw-------" ]
+if [ "$(stat -c %a /etc/fstab)" != "600" ]
 then
-    chmod 0600 /etc/fstab
+    echo "chmod 0600 /etc/fstab"
 fi
 
 # choose categories
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -98,7 +98,7 @@ then
                 break
             fi
         fi
-        count=$(( count + 1))
+        count=$(( $count + 1))
     done
     run_app_script smbmount
 elif [ "$SMB_MOUNT" == "mount SMB-Shares" ]
@@ -116,7 +116,7 @@ then
         then
             args+=("/mnt/smbshares/$count" "$(grep "/mnt/smbshares/$count" /etc/fstab | awk '{print $1}')" OFF)
         fi
-        count=$(( count + 1))
+        count=$(( $count + 1))
     done
     selected_options=$("${args[@]}" 3>&1 1>&2 2>&3)
     count=1
@@ -153,13 +153,13 @@ then
     fi
     args=(whiptail --title "unmount SMB-Shares" --checklist "This option let you unmount SMB-Shares to disconnect network-shares from the host-computer or other machines in the local network.\nChoose what you want to do.\nSelect or unselect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4)
     count=1
-    while  [ count -le 3 ]
+    while  [ $count -le 3 ]
     do
         if [[ $(findmnt -M "/mnt/smbshares/$count") ]]
         then
             args+=("/mnt/smbshares/$count" "$(grep "/mnt/smbshares/$count" /etc/fstab | awk '{print $1}')" OFF)
         fi
-        count=$(( count + 1))
+        count=$(( $count + 1))
     done
     selected_options=$("${args[@]}" 3>&1 1>&2 2>&3)
     count=1
@@ -175,7 +175,7 @@ then
                 msg_box "Your unmount of /mnt/smbshares/$count was successfull!"
             fi
         fi
-        count=$(( count + 1))
+        count=$(( $count + 1))
     done
     run_app_script smbmount
 elif [ "$SMB_MOUNT" == "delete SMB-Mounts" ]
@@ -187,7 +187,7 @@ then
     fi
     args=(whiptail --title "delete SMB-Mounts" --checklist --separate-output "This option let you delete SMB-Shares to disconnect and remove network-shares from the host-computer or other machines in the local network.\nChoose what you want to do.\nSelect or unselect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4)
     count=1
-    while  [ count -le 3 ]
+    while  [ $count -le 3 ]
     do
         if [ "$(grep "/mnt/smbshares/$count" /etc/fstab)" != "" ]
         then
@@ -213,7 +213,7 @@ then
                 msg_box "Your deletion of /mnt/smbshares/$count was successfull!"
             fi
         fi
-        count=$(( count + 1))
+        count=$(( $count + 1))
     done
     run_app_script smbmount
 else
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -109,18 +109,15 @@ then
         run_app_script smbmount
     fi
     args=(whiptail --title "mount SMB-Shares" --checklist --separate-output "This option let you mount SMB-Shares to connect network-shares from the host-computer or other machines in the local network.\nChoose what you want to do.\nIf nothing is shown, then there is nothing to mount.\nSelect or unselect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4)
-    if [[ ! $(findmnt -M "/mnt/smbshares/1") ]] && [ "$(grep /mnt/smbshares/1 /etc/fstab)" != "" ]
-    then
-        args+=("/mnt/smbshares/1" "$(grep /mnt/smbshares/1 /etc/fstab | awk '{print $1}')" OFF)
-    fi
-    if [[ ! $(findmnt -M "/mnt/smbshares/2") ]]  && [ "$(grep /mnt/smbshares/2 /etc/fstab)" != "" ]
-    then
-        args+=("/mnt/smbshares/2" "$(grep /mnt/smbshares/2 /etc/fstab | awk '{print $1}')" OFF)
-    fi
-    if [[ ! $(findmnt -M "/mnt/smbshares/3") ]]  && [ "$(grep /mnt/smbshares/3 /etc/fstab)" != "" ]
-    then
-        args+=("/mnt/smbshares/3" "$(grep /mnt/smbshares/3 /etc/fstab | awk '{print $1}')" OFF)
-    fi
+    count=1
+    while  [ $count -le 3 ]
+    do
+        if [[ ! $(findmnt -M "/mnt/smbshares/$count") ]] && [ "$(grep "/mnt/smbshares/$count" /etc/fstab)" != "" ]
+        then
+            args+=("/mnt/smbshares/$count" "$(grep "/mnt/smbshares/$count" /etc/fstab | awk '{print $1}')" OFF)
+        fi
+        count=$(( $count + 1))
+    done
     selected_options=$("${args[@]}" 3>&1 1>&2 2>&3)
     if [[ $selected_options == *"/mnt/smbshares/1"* ]]
     then
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -84,7 +84,7 @@ then
             occ_command app:enable files_external
             MOUNT_ID=$(occ_command files_external:create SMB1 local null::null -c datadir=/mnt/smbshares/1)
             MOUNT_ID=${MOUNT_ID//[!0-9]/}
-            occ_command files_external:applicable --add-group=admin $MOUNT_ID -q
+            occ_command files_external:applicable --add-group=admin "$MOUNT_ID" -q
             msg_box "Your mount was successfull, congratulations!\n It is accessible in your root directory in /mnt/smbshares/1.\nYou can now use the Nextcloud external storage app to access files there."
         fi
     elif [ "$(grep /mnt/smbshares/2 /etc/fstab)" == "" ]
@@ -100,7 +100,7 @@ then
             occ_command app:enable files_external
             MOUNT_ID=$(occ_command files_external:create SMB2 local null::null -c datadir=/mnt/smbshares/2)
             MOUNT_ID=${MOUNT_ID//[!0-9]/}
-            occ_command files_external:applicable --add-group=admin $MOUNT_ID -q
+            occ_command files_external:applicable --add-group=admin "$MOUNT_ID" -q
             msg_box "Your mount was successfull, congratulations!\n It is accessible in your root directory in /mnt/smbshares/2.\nYou can now use the Nextcloud external storage app to access files there."
         fi
     elif [ "$(grep /mnt/smbshares/3 /etc/fstab)" == "" ]
@@ -116,7 +116,7 @@ then
             occ_command app:enable files_external
             MOUNT_ID=$(occ_command files_external:create SMB3 local null::null -c datadir=/mnt/smbshares/3)
             MOUNT_ID=${MOUNT_ID//[!0-9]/}
-            occ_command files_external:applicable --add-group=admin $MOUNT_ID -q
+            occ_command files_external:applicable --add-group=admin "$MOUNT_ID" -q
             msg_box "Your mount was successfull, congratulations!\n It is accessible in your root directory in /mnt/smbshares/3.\nYou can now use the Nextcloud external storage app to access files there."
         fi
     fi
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# T&M Hansson IT AB © - 2019, https://www.hanssonit.se/
+# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
 
 # shellcheck disable=2034,2059
 true
@@ -37,7 +37,7 @@ SMB_MOUNT=$(whiptail --title "SMB-Share" --radiolist  "This script let you manag
 if [ "$SMB_MOUNT" == "add a SMB-Mount" ]
 then
     # check if mounting slots are available
-    if [ -n "$(grep /mnt/smbshares/1 /root/fstabtest)" ] && [ -n "$(grep /mnt/smbshares/2 /root/fstabtest)" ] && [ -n "$(grep /mnt/smbshares/3 /root/fstabtest)" ]
+    if [ -n "$(grep /mnt/smbshares/1 /etc/fstab)" ] && [ -n "$(grep /mnt/smbshares/2 /etc/fstab)" ] && [ -n "$(grep /mnt/smbshares/3 /etc/fstab)" ]
     then
         msg_box "No mounting slots available. Please delete one SMB-Mount."
         run_app_script smbmount
@@ -92,7 +92,7 @@ then
             then
                 # if not remove this line from fstab
                 msg_box "It seems like the mount wasn't successful. It will get deleted now. Please try again."
-                sed -i "\/mnt\/smbshares\/$count/d" /etc/fstab
+                sed -i "/\/mnt\/smbshares\/$count/d" /etc/fstab
                 break
             else
                 # Install and enable files_extrnal
@@ -113,7 +113,7 @@ then
 elif [ "$SMB_MOUNT" == "mount SMB-Shares" ]
 then
     # check if any smb-share is created
-    if [ "$(grep /mnt/smbshares /etc/fstab)" == "" ]
+    if [ -z "$(grep /mnt/smbshares /etc/fstab)" ]
     then
         msg_box "It seems like you have not created any SMB-Share."
         run_app_script smbmount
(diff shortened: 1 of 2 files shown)


szaimen:
@enoch85 This is already woking. But I have to find out, if it is possible to create entrys to --radiolist or --checklist based on conditions.

And could you help me how I can make the code more slick?
In the moment, there are 3 slots available for mounting smb-shares. How could that be solved to allow an unlimited number? (so maybe based on a rising number or something like that?)

And how can I fix travis? this doesn't make sense:
image

szaimen:
todo:

- [ ] for unmount list only mounted smb-shares
- [ ] for delete list only shares with fstab-entry
- [ ] for mount list only unmounted smb-shares

apps/smbmount.sh
```diff
@@ -20,9 +20,9 @@ root_check
 install_if_not cifs-utils
 
 # secure fstab
-if [ "$(ls -l /etc/fstab | awk '{print $1}')" != "-rw-------" ]
+if [ "$(stat -c %a /etc/fstab)" != "600" ]
 then
-    chmod 0600 /etc/fstab
+    echo "chmod 0600 /etc/fstab"
```
enoch85: err, my testing was merged. great job @enoch85! :D

enoch85:
Travis fixed at least. I don't have time to look at this now, sorry. 

I want to merge the TLS refactoring first.

enoch85:
Just a pro tip; try not to repeat the same code twice. Make a function instead and repeat that function. It's easier to read and understand.

Thanks for working on this!

apps/smbmount.sh
```diff
@@ -289,6 +289,7 @@ then
             msg_box "Your deletion of /mnt/smbshares/3 was successfull!"
         fi
     fi
+    run_app_script smbmount
```
enoch85: I don't understand this. Are you running the script itself when everything is done?

szaimen: > I don't understand this. Are you running the script itself when everything is done?

Yes. The idea is to have a kind of mainmenu for managing smb-mounts.
From there you can mount, unmount, add list and delete smb-shares.
exiting only works from the mainmenu.

enoch85: Hmm... OK....

Maybe I'm just tired now, but what would happen is:

1.  Run the script
2. Make your choices
3. Run the script again inside the same script that is already running 
4. (endless loop)

It doesn't make sense to me.

szaimen: no, because if you exit from the main menu, you exit the whole script.
The main menu is just a whiptile list where you can choose an option. You should definitely try it yourself. I know the code doesn't look that nice, and i have to refactor a lot of things but it works quiet well!

szaimen: image
this is the main menu

szaimen: actually the best part of this script imo is that new smbshares get automatically mounted to the Nextcloud-admin group, so that you are instantly able to access files in that share or just manage it to make it available to other users or just edit the path.

szaimen: > Hmm... OK....
> 
> Maybe I'm just tired now, but what would happen is:
> 
> 1. Run the script
> 2. Make your choices
> 3. Run the script again inside the same script that is already running
> 4. (endless loop)
> 
> It doesn't make sense to me.

okay, you are right. do you have another idea how to go back to the main menu after done with a part? or should it just exit the script?

szaimen:
> Just a pro tip; try not to repeat the same code twice. Make a function instead and repeat that function. It's easier to read and understand.

I will try that later on. Do you have some examples how this could be done?

apps/smbmount.sh
szaimen: > > Hmm... OK....
> > Maybe I'm just tired now, but what would happen is:
> > 
> > 1. Run the script
> > 2. Make your choices
> > 3. Run the script again inside the same script that is already running
> > 4. (endless loop)
> > 
> > It doesn't make sense to me.
> 
> okay, you are right. do you have another idea how to go back to the main menu after done with a part? or should it just exit the script? or do I have to put every option in different files to be able to do that? (probably an advantege: maybe the script gets slicker then.)

I found a possibility: use exec to start the same script again and end the last execution.

szaimen:
> > Just a pro tip; try not to repeat the same code twice. Make a function instead and repeat that function. It's easier to read and understand.
> 
> I will try that later on. Do you have some examples how this could be done?

Or what about this:
Because I am not experienced in writing functions, maybe you could write the functions like you have done in the cookielifetime script and I will try to implement them?

apps/smbmount.sh
szaimen: But exec would probably also exit the startup-script. Wouldn't it?

enoch85: > But exec would probably also exit the startup-script. Wouldn't it?

That's the thing with developing, you need to test test test test. :)

I just want to merge the TLS PR first. Will give this my attention once that's done.

One obvious thing I saw is that you can make a lot of repeating code functions instead. Teyon't need to be put in `lib.sh` as they only are used here.

```
mount_in_files_external_app() {
occ_command app:enable files_external
            MOUNT_ID=$(occ_command files_external:create $1 local null::null -c datadir=$2)
            MOUNT_ID=${MOUNT_ID//[!0-9]/}
            occ_command files_external:applicable --add-group=admin "$MOUNT_ID" -q
            msg_box "Your mount was successfull, congratulations!\n It is accessible in your root directory in $2.\nYou can now use the Nextcloud external storage app to access files there."
}

mount_in_files_external_app SMB1 /mnt/smbshares/1

```

It's simple really, just make repeating code a function and call $1 for the first line after the function, and $2 for the second word in the function, then $3, $4 and so on if it's a larger function.

szaimen: thanks for the example!
But I think loops fit better here since every function is just needed here.
So I've put everything possible into loops.
but it would make sense think about putting some of the remaining code into functions.

szaimen:
```
In apps/smbmount.sh line 101:
        count=$(( $count + 1))
                  ^----^ SC2004: $/${} is unnecessary on arithmetic variables.
```
@enoch85 I currently don't know how to fix that.
Do you have an idea?

enoch85:
Revert back to the way it was and I'll show you the magic with functions. ;)

The loops you created, great! But not easy to read, and produces more lines than needed.

enoch85:
Or, I can show you here:
**Function**
```
mount_in_files_external_app() {
occ_command app:enable files_external
            MOUNT_ID=$(occ_command files_external:create $1 local null::null -c datadir=$2)
            MOUNT_ID=${MOUNT_ID//[!0-9]/}
            occ_command files_external:applicable --add-group=admin "$MOUNT_ID" -q
            msg_box "Your mount was successfull, congratulations!\n It is accessible in your root directory in $2.\nYou can now use the Nextcloud external storage app to access files there."
}
```
**Run the function how many times you want**
```
mount_in_files_external_app SMB1 /mnt/smbshares/1
mount_in_files_external_app SMB2 /mnt/smbshares/2
mount_in_files_external_app SMB3 /mnt/smbshares/3
mount_in_files_external_app SMB4 /mnt/smbshares/4
mount_in_files_external_app SMB5 /mnt/smbshares/5
```

apps/smbmount.sh
szaimen: > But exec would probably also exit the startup-script. Wouldn't it?

@enoch85 Since that would probably be a problem, I could maybe solve it by putting every option into a new file and put the main script into a loop. What do you think about this?

szaimen:
> Or, I can show you here:
> **Function**
> 
> ```
> mount_in_files_external_app() {
> occ_command app:enable files_external
>             MOUNT_ID=$(occ_command files_external:create $1 local null::null -c datadir=$2)
>             MOUNT_ID=${MOUNT_ID//[!0-9]/}
>             occ_command files_external:applicable --add-group=admin "$MOUNT_ID" -q
>             msg_box "Your mount was successfull, congratulations!\n It is accessible in your root directory in $2.\nYou can now use the Nextcloud external storage app to access files there."
> }
> ```
> 
> **Run the function how many times you want**
> 
> ```
> mount_in_files_external_app SMB1 /mnt/smbshares/1
> mount_in_files_external_app SMB2 /mnt/smbshares/2
> mount_in_files_external_app SMB3 /mnt/smbshares/3
> mount_in_files_external_app SMB4 /mnt/smbshares/4
> mount_in_files_external_app SMB5 /mnt/smbshares/5
> ```
> 
> If you really want a loop, then loop the function, or put the loop inside the function.

@enoch85 Thanks for the example. 

Since e.g. the whole part of mounting, testing and add external storage got repeated, I thought looping would be the best. And why are comments not a way to compensate a missing function in this case? Because if not looping here, then mounting, testing, add external storage would get manually repeated by myself, which isn't that efficient imo (also if functions are used). Theoratically when using loops, I could also change the number of available mount-slots...

My problem with functions is, that I am currently not able to understand which parts of the script make sense to put in a function. 
Because taking your example and putting it into a function also doesn't make a lot of sense to me, if it is only used here... (and only written once inside a loop)

I would just put all loops that I created into a function but thats probably not what you meant. (e.g. I would put line 79-116 into a function. Another example would be 250-269).

enoch85:
The first code you did looked perfectly fine to me. Why complicate it more than necessary?

A rule of thumb, if a script section is used more than once (which it seems to be here) then a function is better since it means less lines of actual code.

Some functions in `lib.sh` are only used in one script for example. The idea behind it is that when it's time for maintaining or changing something, you don't want to change in a number of places -you also don't want 1000 lines of code which can be done in maybe 500. 

Main functions used in several different scripts (or which maybe will be in the future) are put in `lib.sh`. Special functions (only used for one script)are put in the same script.

What I'm trying to say here: don't repeat code.

> My problem with functions is, that I am currently not able to understand which parts of the script make sense to put in a function.

Maybe that's also why you don't see the benefits of it?

I think like this: Is a loop of 100 lines of code, better than one function that can be repeated several times, only generating 30 lines of code? If both do the same thing I'd vote for the function being used.

> Because taking your example and putting it into a function also doesn't make a lot of sense to me, if it is only used here... (and only written once inside a loop)

It makes sense, because it makes the code easier to read. You only have to change stuff in one place and not 3 (in this case). You also don't use as many lines as you would without using a function.

> Or Do you mean putting parts like the following into functions although they are only used here?

Yes, that's one example. 3 lines of code just became one line the second time it's used.

---------

I will read through it now and see if there are any obvious stuff that can be changed.

szaimen:
> The first code you did looked perfectly fine to me. Why complicate it more than necessary?
> 
> A rule of thumb, if a script section is used more than once (which it seems to be here) then a function is better since it means less lines of actual code.
> 
> Some functions in `lib.sh` are only used in one script for example. The idea behind it is that when it's time for maintaining or changing something, you don't want to change in a number of places -you also don't want 1000 lines of code which can be done in maybe 500.
> 
> Main functions used in several different scripts (or which maybe will be in the future) are put in `lib.sh`. Special functions (only used for one script)are put in the same script.
> 
> What I'm trying to say here: don't repeat code.
> 
> > My problem with functions is, that I am currently not able to understand which parts of the script make sense to put in a function.
> 
> Maybe that's also why you don't see the benefits of it?
> 
> I think like this: Is a loop of 100 lines of code, better than one function that can be repeated several times, only generating 30 lines of code? If both do the same thing I'd vote for the function being used.
> 
> > Because taking your example and putting it into a function also doesn't make a lot of sense to me, if it is only used here... (and only written once inside a loop)
> 
> It makes sense, because it makes the code easier to read. You only have to change stuff in one place and not 3 (in this case). You also don't use as many lines as you would without using a function.
> 
> > Or Do you mean putting parts like the following into functions although they are only used here?
> 
> Yes, that's one example. 3 lines of code just became one line the second time it's used.
> 
> I will read through it now and see if there are any obvious stuff that can be changed.

Thanks for the clarification! Now it makes sense.
If you want, I can revert the loops. Afterwards we could immplement the functions?

enoch85:
I've read through it now and made some quick changes.

I'd say, make all the whiptail options functions instead. That way you can rerun them as you please, and there is one less if statment to be made for each choice.

```
add_mount() {
code of that option
}

mount_shares() {
code of that option
}

show_all_mounts() {
code of that option
}

unmount_shares() {
code of that option
}

delete_mounts() {
code of that option
}
```

Also, I saw that you invoke the script itself multiple times. I think that will cause issues and zombie processes in the underlying OS. Each time the `run_app_script smbmount` is called a process for that script is started. I never saw you ending it anywhere? Running the script several times logically means that there will be as many processes.

Making the options functions instead, you could cleanly exit the process when done, and go to the next. It's late now, but something like that should work.

enoch85:
Also, I can see now that I missed some spots..

Here's one example: `ìf [ $(grep /mnt/smbshares /etc/fstab)" == "" ]`
Use this instead: ` ìf [ -z $(grep /mnt/smbshares /etc/fstab)" ]`

Here's one example: `ìf [ $(grep /mnt/smbshares /etc/fstab)" != "" ]`
Use this instead: ` ìf [ -n $(grep /mnt/smbshares /etc/fstab)" ]`

More stuff: https://ryanstutorials.net/bash-scripting-tutorial/bash-if-statements.php

Please also change copyright to 2020 while you're at it. Thanks! :)

szaimen:
> Also, I can see now that I missed some spots..
> 
> Here's one example: `ìf [ $(grep /mnt/smbshares /etc/fstab)" == "" ]`
> Use this instead: ` ìf [ -z $(grep /mnt/smbshares /etc/fstab)" ]`
> 
> Here's one example: `ìf [ $(grep /mnt/smbshares /etc/fstab)" != "" ]`
> Use this instead: ` ìf [ -n $(grep /mnt/smbshares /etc/fstab)" ]`
> 
> More stuff: https://ryanstutorials.net/bash-scripting-tutorial/bash-if-statements.php
> 
> Please also change copyright to 2020 while you're at it. Thanks! :)

should be all done now.

szaimen:
> delete_mounts

This is done.
I have just changed the run_app_script to return and put the whole main menu into a while loop. Now it works like I wanted it to work. ;)

szaimen:
Now we should fix travis next imo.

apps/smbmount.sh
szaimen: This is done.

szaimen:
Actually, what travis suggests doesn't work. So I don't know what to change here to make it work.

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+    SMB_PASSWORD=$(whiptail --inputbox "Please enter the password of the SMB-user $SMB_USER" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    if [[ "no" == $(ask_yes_or_no "Is this correct? $SMB_PASSWORD") ]]
+    then
+        msg_box "It seems like your weren't satisfied by the password for the SMB-User you entered. Please try again."
+    else
+        break
+    fi
+done
+# write everything to /etc/fstab, mount and connect external storage
+count=1
```
szaimen: Maybe this is the cause why the suggested change doesn't work.
It should be count="1"
And then probably
count=$($count+1) works

szaimen:
```
In apps/smbmount.sh line 146:
if [ -z "$(grep /mnt/smbshares /etc/fstab)" ]
     ^-- SC2143: Use ! grep -q instead of comparing output with [ -z .. ].
In apps/smbmount.sh line 156:
    if [ -n "$(grep "/mnt/smbshares/$count" /etc/fstab)" ]
         ^-- SC2143: Use grep -q instead of comparing output with [ -n .. ]
```
Ive tried ! grep -q and grep -q and both are returning nothing. So this couldn't have worked. 
Do you know another possibility of doing this?

apps/smbmount.sh
```diff
@@ -28,7 +28,7 @@ fi
 # functions
 add_mount() {
 # check if mounting slots are available
-if [ -n "$(grep /mnt/smbshares/1 /etc/fstab)" ] && [ -n "$(grep /mnt/smbshares/2 /etc/fstab)" ] && [ -n "$(grep /mnt/smbshares/3 /etc/fstab)" ]
+if [ grep -q /mnt/smbshares/1 /etc/fstab ] && [ grep -q /mnt/smbshares/2 /etc/fstab) ] && [ grep -q /mnt/smbshares/3 /etc/fstab ]
```
enoch85: I don't think you need the [ ]here as it's no if statment with a switch like -n -z -f or similar.

apps/smbmount.sh
szaimen: this is fixed

apps/smbmount.sh
szaimen: thanks for the advice. Travis is fixed finally!

szaimen:
@enoch85 Travis is fixed finally! 
Where does this script fit inside the startup script? Is it an app or more a static script?

enoch85:
> Where does this script fit inside the startup script? Is it an app or more a static script?

Great!

I'd say it's an app script.

The TLS PR is now merged (finally) and I will take some time this week to look at this PR.

Great job! :+1:

szaimen:
> > Where does this script fit inside the startup script? Is it an app or more a static script?
> 
> I'd say it's an app script.

okay, I've added it to this section inside the startup script.

szaimen:
Hm... Since this commit:  https://github.com/nextcloud/vm/pull/1024/commits/45399fa844677650ee804739b1cd20ff82c13f10
Travis won't run. But I am sure I haven't broken anything.
Maybe it' because that branch isn't up-to-date?

nextcloud-startup-script.sh
```diff
@@ -440,7 +440,8 @@ whiptail --title "Which apps do you want to install?" --checklist --separate-out
 "FullTextSearch" "(Elasticsearch for Nextcloud [2GB RAM])   " OFF \
 "PreviewGenerator" "(Pre-generate previews)   " OFF \
 "LDAP" "(Windows Active directory)   " OFF \
-"Talk" "(Nextcloud Video calls and chat)   " OFF 2>results
+"Talk" "(Nextcloud Video calls and chat)   " OFF \
+"SMB" "(Connect to SMB-Shares from your local network)   " OFF 2>results
```
enoch85: Ok, good call, since it's actually using the SMB app. :)

enoch85:
> Maybe it' because that branch isn't up-to-date?

Travis is being Travis. Simple as that. :)

enoch85:
Btw, as I don't have any SMB shares to test against, I'm fine with your word on that this is OK and solid as rock. :)

In other words, test test test, tell me when it's 100% and I'll do some formatting stuff, you test again, and if it works we merge.

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+# secure fstab
+if [ "$(stat -c %a /etc/fstab)" != "600" ]
+then
+    chmod 600 /etc/fstab
+fi
+
+# functions
+add_mount() {
+# check if mounting slots are available
+if grep -q /mnt/smbshares/1 /etc/fstab && grep -q /mnt/smbshares/2 /etc/fstab && grep -q /mnt/smbshares/3 /etc/fstab
```
enoch85: Please make `/mnt/smbshares` a variable if we decide to change the name later on: `$SMBSHARES`.

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+then
+    msg_box "No mounting slots available. Please delete one SMB-Mount."
```
enoch85: Change wording please:

"All three slots are occupied. No mounting slots available. Please delete one of the SMB-mounts."

Please also change it *everywhere* to SMB-mounts, SMB-server and so on - no capital letter after -.

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+    return
+fi
+# enter smb-server and share name
```
enoch85: Use capital letter on the first character in the comments. I'm picky I know. :)

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+while true
+do
+    SERVER_SHARE_NAME=$(whiptail --inputbox "Please Enter the Server and Share-Name like this:\n//Server/Share" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
```
enoch85: Please enter the server and Share-name

enoch85: Please also check if there are more occurrences like that. Be consistent please.

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+        break
+    fi
+done
+#enter the password of the smb-user
+while true
+do
+    SMB_PASSWORD=$(whiptail --inputbox "Please enter the password of the SMB-user $SMB_USER" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    if [[ "no" == $(ask_yes_or_no "Is this correct? $SMB_PASSWORD") ]]
+    then
+        msg_box "It seems like your weren't satisfied by the password for the SMB-User you entered. Please try again."
```
enoch85: SMB-user

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+    else
+        break
+    fi
+done
+# write everything to /etc/fstab, mount and connect external storage
+count=1
+while  [ $count -le 3 ]
+do
+    # check which mounting slot is available
+    if ! grep -q "/mnt/smbshares/$count" /etc/fstab
```
enoch85: /mnt/smbshares/**"**$count"

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+    then 
+        # write to /etc/fstab and mount
+        echo "$SERVER_SHARE_NAME /mnt/smbshares/$count cifs username=$SMB_USER,password=$SMB_PASSWORD,vers=3,uid=33,gid=33,file_mode=0770,dir_mode=0770,nounix,noserverino 0 0" >> /etc/fstab
+        mkdir -p /mnt/smbshares/$count
+        mount /mnt/smbshares/$count
```
enoch85: /mnt/smbshares/**"**$count"

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+        # check if mounting was successful
+        if [[ ! $(findmnt -M "/mnt/smbshares/$count") ]]
```
enoch85: /mnt/smbshares/**"**$count"

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+    count=$((count+1))
+done
+return
+}
+
+mount_shares() {
+# check if any smb-share is created
+if ! grep -q /mnt/smbshares /etc/fstab
+then
+    msg_box "It seems like you have not created any SMB-Share."
```
enoch85: SMB-share

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+    return
+fi
+args=(whiptail --title "mount SMB-Shares" --checklist --separate-output "This option let you mount SMB-Shares to connect network-shares from the host-computer or other machines in the local network.\nChoose what you want to do.\nIf nothing is shown, then there is nothing to mount.\nSelect or unselect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4)
```
enoch85: SMB-shares

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+
+show_all_mounts() {
+# if no entry created nothing to show
+if ! grep -q /mnt/smbshares /etc/fstab
+then
+    msg_box "You haven't created any SMB-Mount. So nothing to show."
+    return
+fi
+# find out which smb-shares are available
+args=(whiptail --title "list SMB-Shares" --checklist "This option let you show detailed information about your SMB-Shares.\nChoose what you want to show.\nSelect or unselect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4)
```
enoch85: SMB-shares

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+    count=$((count+1))
+done
+return
+}
+
+show_all_mounts() {
+# if no entry created nothing to show
+if ! grep -q /mnt/smbshares /etc/fstab
+then
+    msg_box "You haven't created any SMB-Mount. So nothing to show."
```
enoch85: SMB-mount

apps/smbmount.sh
enoch85: And so on...

enoch85:
Please fix the rest according to the review, I only read through half the script.

szaimen:
> Please fix the rest according to the review, I only read through half the script.

I do that when I have time. 👍

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+done
+# write everything to /etc/fstab, mount and connect external storage
+count=1
+while  [ $count -le 3 ]
+do
+    # check which mounting slot is available
+    if ! grep -q "/mnt/smbshares/$count" /etc/fstab
+    then 
+        # write to /etc/fstab and mount
+        echo "$SERVER_SHARE_NAME /mnt/smbshares/$count cifs username=$SMB_USER,password=$SMB_PASSWORD,vers=3,uid=33,gid=33,file_mode=0770,dir_mode=0770,nounix,noserverino 0 0" >> /etc/fstab
```
szaimen: @enoch85 is the uid and gid of www-data always 33?

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+        mkdir -p /mnt/smbshares/$count
+        mount /mnt/smbshares/$count
+        # check if mounting was successful
+        if [[ ! $(findmnt -M "/mnt/smbshares/$count") ]]
+        then
+            # if not remove this line from fstab
+            msg_box "It seems like the mount wasn't successful. It will get deleted now. Please try again."
```
szaimen: as a note to myself: add a hint here, that maybe you have to add servermaxprotocol to the smb.conf file of the linux smb server: https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html#SERVERMAXPROTOCOL

apps/smbmount.sh
szaimen: if I would made a variable from /mnt/smbshares, it would then be: "$SMBSHARES/$count", wouldn't it?

apps/smbmount.sh
szaimen: here the same like above

apps/smbmount.sh
szaimen: here the same like above

szaimen:
> Btw, as I don't have any SMB shares to test against, I'm fine with your word on that this is OK and solid as rock. :)
> 
> In other words, test test test, tell me when it's 100% and I'll do some formatting stuff, you test again, and if it works we merge.

You could of course easily create your own (e.g. by using any PC with Linux or Windows) but OK.

enoch85:
> You could of course easily create your own SMB-Server (e.g. by using any PC with Linux or Windows) but OK.

Yeah sure, I could do that, but it would take longer time to merge this PR, and usually I let the Author of the PR for new scripts be maintainers for the script in question. So all the bugs that happen (if any) are up to you to fix.

apps/smbmount.sh
enoch85: I'm not sure actually, but make it a $var, it's safer.

apps/smbmount.sh
```diff
@@ -0,0 +1,288 @@
...
+then
+    msg_box "You haven't created any SMB-Mount. So nothing to show."
+    return
+fi
+# find out which smb-shares are available
+args=(whiptail --title "list SMB-Shares" --checklist "This option let you show detailed information about your SMB-Shares.\nChoose what you want to show.\nSelect or unselect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4)
+count=1
+while  [ $count -le 3 ]
+do
+    if grep -q "/mnt/smbshares/$count" /etc/fstab
```
szaimen: @enoch85 if I would made a variable from /mnt/smbshares, it would then be: "$SMBSHARES/$count", wouldn't it?

apps/smbmount.sh
szaimen: is done

apps/smbmount.sh
szaimen: should be done

apps/smbmount.sh
szaimen: should be done

apps/smbmount.sh
szaimen: should be done

apps/smbmount.sh
szaimen: should be done

apps/smbmount.sh
szaimen: should be done

apps/smbmount.sh
szaimen: should be done

apps/smbmount.sh
szaimen: should be done

apps/smbmount.sh
szaimen: this is done

apps/smbmount.sh
szaimen: this is done.

apps/smbmount.sh
szaimen: this should be done

apps/smbmount.sh
szaimen: this should be done

apps/smbmount.sh
szaimen: this should be done

apps/smbmount.sh
szaimen: this should be done

apps/smbmount.sh
szaimen: I have done this. But I had to make to variables, because sed didn't like the formatting.

szaimen:
@enoch85 I am ready. Everything seems to work and I have resolved your reviews.

apps/smbmount.sh
```diff
@@ -0,0 +1,296 @@
...
+        # Check if mounting was successful
+        if [[ ! $(findmnt -M "$SMBSHARES/$count") ]]
+        then
+            # If not remove this line from fstab
+            msg_box "It seems like the mount wasn't successful. It will get deleted now. Please try again.\nAs a hint: you might fix the connection problem by enabling SMB3 on your SMB-server."
+            sed -i "/$SMBSHARES_SED\/$count/d" /etc/fstab
+            break
+        else
+            # Install and enable files_external
+            install_and_enable_app files_external
```
szaimen: actually the only thing, I find a bit odd is, that this function enables the external storage app each time or at least it says that to the user...
Maybe a check would be better, if the app is already enabled and only enable it if it is disabled.

szaimen: and btw: files_external doesn't has to get installed first (since it is shipped together with Nextcloud). so that check is unnecessary here.

enoch85: Please read the function.

There's no need for a check as we want to install it if its not installed, and enable it if its not enabled. That's what the function does. It will always say that it enables it even if its already enabled since that's how the `occ` command works in Nextcloud.

Adding a check is just cosmetic in my opinion as the only change would be that we don't show the "enables text" to the user. Also prone to more failure, because what if the `occ` command displays the info differently in coming versions? Then the check would fail anyway.

apps/smbmount.sh
```diff
@@ -0,0 +1,296 @@
...
+# Check if mounting slots are available
+if grep -q "$SMBSHARES/1" /etc/fstab && grep -q "$SMBSHARES/2" /etc/fstab && grep -q "$SMBSHARES/3" /etc/fstab
+then
+    msg_box "All three slots are occupied. No mounting slots available. Please delete one of the SMB-mounts."
+    return
+fi
+# Enter SMB-server and Share-name
+while true
+do
+    SERVER_SHARE_NAME=$(whiptail --inputbox "Please Enter the server and Share-name like this:\n//Server/Share" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
```
enoch85: should enter be with capital "E"?

apps/smbmount.sh
szaimen: OK

apps/smbmount.sh
szaimen: No, but that's actually something I haven't check. I will now read the whole script again for spelling mistakes.

szaimen: This is hopefully done. Please tell me, if you find more.

PR #1025, opened 2020-01-10
Refactor the way TLS certs are generated + make all options (standalone, dns, tls) available everywhere

liao20081228:
@enoch85 

 I have done all the tests in the case the port was blocked.  It works well.


lets-encrypt/activate-ssl.sh
```diff
@@ -294,7 +231,10 @@ https://github.com/nextcloud/vm
...
 
-# Cleanup
-apt remove certbot -y
-apt autoremove -y
-clear
+    # Cleanup
+    apt remove certbot -y
+    apt autoremove -y
+    clear
+fi
```
enoch85: In my opinion, the cleanup should happen in the function in lib.sh if it fails. That way we don't repeat the code in all files that use the function.

lib.sh
```diff
@@ -430,11 +430,41 @@ else
...
+    print_text_in_color "${ICyan}" "try to generate certs and validate them with $f method."
+    eval  ff=$(echo "\$$f")
+    if eval "$ff"
+    then
+        return 0
+    elif [ "$f" != "${methods[@]:(-1)}" ]
+    then
+        print_text_in_color "${ICyan}" "It seems like no certs were generated when trying to validate them with the $f method. We will do more tries."
+        any_key "Press any key to continue..."
+    else
```
enoch85: PERFECT!

lib.sh
```diff
@@ -947,7 +983,7 @@ done
 }
 
 print_text_in_color() {
-	printf "%b%s%b\n" "$1" "$2" "$Color_Off"
+printf "%b%s%b\n" "$1" "$2" "$Color_Off"
```
enoch85: If you want to to indentation on other stuff at the same time, sure! Why not. :)

Many devs, many styles. I have tried to keep it clean, but as you said, there's no structure in some of it. 
```
function() {
[should start at the first line, or what do you think? Intended 4 spaces here as well?
}
```
```
if arguments
    [as I posted before]
fi
```

lib.sh
```diff
@@ -430,11 +430,41 @@ else
...
+#http-01
+local  standalone="certbot certonly --standalone --pre-hook 'service apache2 stop' --post-hook 'service apache2 start' $default_le"
+#tls-alpn-01
+local  tls_alpn_01="certbot certonly --preferred-challenges tls-alpn-01 $default_le"
+#dns
+local  dns="certbot certonly --manual --manual-public-ip-logging-ok --preferred-challenges dns $default_le"
+local  methods=(standalone dns)
+for f in "${methods[@]}"
+do
+    print_text_in_color "${ICyan}" "try to generate certs and validate them with $f method."
```
enoch85: `"s|try|Trying|g"`

lib.sh
```diff
@@ -430,11 +430,41 @@ else
...
+        return 0
+    elif [ "$f" != "${methods[@]:(-1)}" ]
+    then
+        print_text_in_color "${ICyan}" "It seems like no certs were generated when trying to validate them with the $f method. We will do more tries."
+        any_key "Press any key to continue..."
+    else
+        print_text_in_color "${ICyan}" "It seems like no certs were generated when trying to validate them with the $f method. We have tried all the methods."
+        print_text_in_color "${IRed}" "Can't generate certs and validate them.  Please check your DNS and try again."
+        any_key "Press any key to continue..."
+        return 1;
```
enoch85: Put the cleanup here maybe?

enoch85:
Really nice work so far! Thanks a lot!

lib.sh
```diff
@@ -430,11 +430,41 @@ else
 fi
 }
 
-# Let's Encrypt for subdomains
-le_subdomain() {
+#generate certs and auto-configure
+# https://certbot.eff.org/docs/using.html#certbot-command-line-options
+generate_cert() {
+if [[ "no" == $(ask_yes_or_no "Do you need to add the upgrade-insecure-requests(UIR) and Strict-Transport-Security(HSTS) header to every HTTP response?") ]]
```
enoch85: Why ask? 99% of all the users doesn't even know what that meaning means. 

It's added due to security. The more secure the better.

liao20081228: The onlyoffice and collabora have no need of HSTS and UIR.

lib.sh
liao20081228: when failure happens in the progress of installing onlyoffice， the script of automatic updating of nextcloud‘s certs will fail because of uninstall of certbot

lib.sh
enoch85: Then make it an if statement instead. 

```
if [ -z $SUBDOMAIN ]
then
    add HSTS and UIR
fi
```

lib.sh
enoch85: Aah, true. I'm tired now. Thanks!

lib.sh
liao20081228: In a few functions, the first code in function has  indention，but most of them have no indention.

I approve that the coding style should be consistent with the majority of the code.

enoch85:
@liao20081228 Are you done with this so that I can start making my own changes?

liao20081228:
> @liao20081228 Are you done with this so that I can start making my own changes?

Yeah

liao20081228:
By the way, I still think that the return value of domain_check_200() should be caught and handled. when the domain is not unreachable， all operation that depends on a domain name should be stopped. This is just my personal opinion. 
Thanks.

liao20081228:
@enoch85 
Yeah, the work that I wanna does have been completed. You continue with the changes you want. 
 
Thanks.

enoch85:
> By the way, I still think that the return value of domain_check_200() should be caught and handled. when the domain name is not unreachable， all operation that depends on a domain name should be stopped. This is just my personal opinion.

Actually I think the `domain_check_200` is working really great, just as intended. I just refactored it some months ago. In some cases regarding the functions I handle the `return` value in the original script, based on what's most suiting.

But what did you have in mind? Is the change out of scope for this PR, please make a new one and we can discuss.

Thanks!

> Yeah, the work that I wanna does have been completed


Really nice, awesome work! Thank you soo much! :rocket:

test.sh
```diff
@@ -0,0 +1,10 @@
+. ./lib.sh
+
+msg_box "It seems like the port ${1} is closed. This could
+happend when your ISP has blocked the port, or that the port
+isn't open.
+
+If you are 100% sure the port ${1} is open you can now choose
+to continue. There are no guarantees that it will work anyway
```
enoch85: Do we call this test somewhere or is it only a test for this PR?

liao20081228: > Do we call this test somewhere or is it only a test for this PR?

Oh, my god! It was careless of me. That is a file I used for a test.

enoch85:
@liao20081228 I haven't tested my changes, but please fix travis and then I'll test later tonight.

Please also test all the scripts from your end.

Thanks!

enoch85:
@liao20081228 

I just tested and everything seems to work as expected. I can't see to figureout how to fix the travis tests though.

```
In lib.sh line 460:

    eval  ff=$(echo "\$$f")

             ^------------^ SC2046: Quote this to prevent word splitting.

             ^------------^ SC2116: Useless echo? Instead of 'cmd $(echo foo)', just use 'cmd foo'.

In lib.sh line 461:

    if eval "$ff"

             ^-^ SC2154: ff is referenced but not assigned.

In lib.sh line 464:

    elif [ "$f" != "${methods[@]:(-1)}" ]

                   ^------------------^ SC2198: Arrays don't work as operands in [ ]. Use a loop (or concatenate with * instead of @).
```

liao20081228:
>I just tested and everything seems to work as expected. I can't seem to figure out how to fix the travis tests though.

This is where I am confused. It seems that he cannot recognize this syntax, but bash works correctly

liao20081228:
@enoch85 
 I have changed the implementation of `for` loop，but the Travis-ci always is pending. I cannot get the result for these changes.

lib.sh
```diff
@@ -440,34 +440,45 @@ fi
 # https://certbot.eff.org/docs/using.html#certbot-command-line-options
 generate_cert() {
 uir_hsts=""
+disable_000=""
 if [ -z "$SUBDOMAIN" ]
 then
     uir_hsts="--uir --hsts"
 fi
+if [ -e "$SITES_AVAILABLE"/../sites-enabled/000-default.conf ]
```
enoch85: Why did you add this? Just adds complexity when there's no need imo.

lib.sh
```diff
@@ -440,34 +440,45 @@ fi
...
 #dns
 local  dns="certbot certonly --manual --manual-public-ip-logging-ok --preferred-challenges dns $default_le"
 local  methods=(standalone dns)
-for f in "${methods[@]}"
+
+for f in ${methods[*]}
 do
     print_text_in_color "${ICyan}" "Trying to generate certs and validate them with $f method."
-    eval  ff=$(echo "\$$f")
+    ff="\$$f"
```
enoch85: I don't know if this is the bug, but current status is:

```
+ methods=(standalone dns)
+ local methods
+ for f in ${methods[*]}
+ print_text_in_color '\e[0;96m' 'Trying to generate certs and validate them with standalone method.'
+ printf '%b%s%b\n' '\e[0;96m' 'Trying to generate certs and validate them with standalone method.' '\e[0m'
Trying to generate certs and validate them with standalone method.
+ ff='$standalone'
+ eval '$standalone'
++ certbot certonly --standalone --pre-hook ''\''service' apache2 'stop'\''' --post-hook ''\''service' apache2 'start'\''' --rsa-key-size 4096 --renew-by-default --no-eff-email --agree-tos --uir --hsts --server https://acme-v02.api.letsencrypt.org/directory -d cloud.ellahansson.se
usage: 
  certbot [SUBCOMMAND] [options] [-d DOMAIN] [-d DOMAIN] ...

Certbot can obtain and install HTTPS/TLS/SSL certificates.  By default,
it will attempt to use a webserver both for obtaining and installing the
certificate. 
certbot: error: unrecognized arguments: apache2 stop' apache2 start'
```

Standalone doesn't work.

liao20081228: I will do a test for this.

lib.sh
enoch85: Or maybe this does something that I missed?

liao20081228: In previous versions，  I simply disabled this default site， and reload apache2. When failure to generate certs，This site has not been re-enabled.

enoch85:
@liao20081228 I don't know if you are familiar with https://github.com/koalaman/shellcheck/wiki/SC2003 or that you can check the Travis before your commit with https://www.shellcheck.net/

I'm happy to merge this once the bugs are fixed and travis checks are green. Would be nice if that could be done today.

Thanks!

lib.sh
enoch85: We use certain files instead of 000-default.conf: $TLS_CONF and $HTTP_CONF.

Here we do that logic in the test script already: https://github.com/nextcloud/vm/blob/master/lets-encrypt/test-new-config.sh#L15-L24

Maybe move that logic to the function instead?

lib.sh
liao20081228: @enoch85 
 
I have done a test, it works well with dns and standalone.  Maybe Your code is not up to date.

liao20081228: @enoch85 

The reason is that the variable does not expand correctly。I have fixed it.

liao20081228:
@enoch85 

Hi,
I have fixed the bug that the nesting variable couldn't expand correctly and do a test successfully.  

I also made the code past the Travis-ci. 

Everything looks fine. Maybe you couldn't do more tests ，because my ports always are blocked by my ISP.
Thanks！

enoch85:
Now tested on a fresh install with the merged code in master. Works perfectly.

Thanks you very much for doing this @liao20081228!

PR #1027, opened 2020-01-10
virtual kernels not needed in the home sme server

enoch85:
Investigate if Collabora / OnlyOffice needs some extra kernels though, and maybe for UTF8 as in the comment.


enoch85:
This is fixed in 20.04

PR #1030, opened 2020-01-13
notify users in gui

enoch85:


enoch85:
cc @szaimen

szaimen:
> I guess we have to test the function with admin user ids that contain spaces. (If that is possible)
> If not possible it should work.

szaimen:
notify_user_gui() {
USER=$(occ_command user:list | awk '{print $3}';)
for user in $USER
do
    if [[ $(occ_command user:info "$user")  == *"- admin"* ]]
    then
        args+="$user"
    fi
done
NCADMIN="${args[*]}"
for admin in $NCADMIN
do
    occ_command notification:generate -l "$2" "$admin" "$1"
done
}
That should fix the function bugs, that travis shows.

enoch85:
> I think it is possible. So we should test that case.

I just tested. Users with spaces doesn't work.

enoch85:
Right now I'm laborating with this:

```

notify_user_gui() {
USER=$(occ_command user:list | awk '{print $3}';)
print_text_in_color "$ICyan" "Looping through users..."
for user in $USER
do
    if occ_command user:info "$user" | grep -q "admin";
    then
        admin_users+=("$user" )
    fi
done

print_text_in_color "$ICyan" "Posting to admins..."
for admin in "${admin_users[*]}"
do
    echo "$2 $admin $1"
    #occ_command notification:generate -l "$2" "$admin" "$1"
done
}

notify_user_gui \
"Please setup SMTP" \
"Please remember to setup SMTP to be able to send shared links, user notifications and more via email. Please go here and start setting it up: https://your-nextcloud/settings/admi$
```

For loops aren't my strongest side, and right now it's posting all users on the same line, without looping through them. Need some advice here.

szaimen:
> Right now I'm laborating with this:
> 
> ```
> . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
> 
> notify_user_gui() {
> USER=$(occ_command user:list | awk '{print $3}';)
> print_text_in_color "$ICyan" "Looping through users..."
> for user in $USER
> do
>     if occ_command user:info "$user" | grep -q "admin";
>     then
>         admin_users+=("$user" )
>     fi
> done
> 
> print_text_in_color "$ICyan" "Posting to admins..."
> for admin in "${admin_users[*]}"
> do
>     echo "$2 $admin $1"
>     #occ_command notification:generate -l "$2" "$admin" "$1"
> done
> }
> 
> notify_user_gui \
> "Please setup SMTP" \
> "Please remember to setup SMTP to be able to send shared links, user notifications and more via email. Please go here and start setting it up: https://your-nextcloud/settings/admi$
> ```
> 
> For loops aren't my strongest side, and right now it's posting all users on the same line, without looping through them. Need some advice here.

I will look for a solution later on 👍

in theory it should work like this:
print the user list to a file -> build a while loop (with the length of lines of that txt file) that seperates each line into an index of an array with "" surrounding each user -> do the for loop for getting the user:info -> do the for loop for notification; then all users should work regardless of if there are spaces in the user id or not.

enoch85:
It's fixed now, a simple mistake as always.

Will test some more and merge soon.

szaimen:
> It's fixed now, a simple mistake as always.
> 
> Will test some more and merge soon.

very cool! 😄👍

lib.sh
```diff
@@ -1080,6 +1080,31 @@ case "${1}" in
...
+# "Subject"
+# "Message"
+#
+# occ_command notification:generate -l "$2" "$admin" "$1"
+notify_user_gui() {
+USER=$(occ_command user:list | awk '{print $2}' | cut -d ":" -f1;)
+print_text_in_color "$ICyan" "Looping through users..."
+for user in $USER
+do
+    if occ_command user:info "$user" | grep -q "admin";
```
szaimen: actually I would grep  for "- admin"
since it is possible theoratically that a user as admin in it, but  "- admin" definitely grep's the group.

lib.sh
```diff
@@ -1080,6 +1080,31 @@ case "${1}" in
...
 }
 
+# Example:
+# notify_user_gui \
+# "Subject" \
+# "Message"
+#
+# occ_command notification:generate -l "$2" "$admin" "$1"
+notify_user_gui() {
+USER=$(occ_command user:list | awk '{print $2}' | cut -d ":" -f1;)
```
szaimen: `cut -d ":" -f1;`
what  is tis addition doing?

enoch85: run `"occ_command user:list | awk '{print $2}"`  and notice that it adds a `:`, the `cut` command deletes the `:`

enoch85:
OK, I'm done. Please test @szaimen :)

enoch85:
No need to test, it's solid.

MERGE!

Thanks for the idea about the `for` loop @szaimen!

szaimen:
> No need to test, it's solid.
> 
> MERGE!
> 
> Thanks for the idea about the `for` loop @szaimen!

You are welcome!

szaimen:
@enoch85 have you retested users with spaces in the userid?
in my testing now it didn't work...

szaimen:
image
with these users, test test didn't work because awk $2 cuts the second element, that is only the first part of the id in this case for test test

enoch85:
It doesn't fetch the user name, it fetches the user id. It's not possible to have usernames with spaces afaik.

szaimen:
> It doesn't fetch the user name, it fetches the user id. It's not possible to have usernames/id's with spaces afaik.

but it is :/
image

enoch85:
That situation will never happen as it's not possible to create a user called `test test`, you would need to make it `test_test` which is one word.

szaimen:
image
it is not possible?

szaimen:
maybe a bug?

enoch85:
No, read again. :)

Username: 1 word
Display Name: several words.

We check for username in the script, not display name.

szaimen:
ah okay, now i see it!

szaimen:
but what do you see here:
image

enoch85:
OK, run `nextcloud_occ user:create test test -g admin` Please post the result.

szaimen:
image

enoch85:
Please remove the old user first.

szaimen:
image

enoch85:
Whut! Now I would call that a bug! It shouldn't be possible to add usernames with spaces in my opinion.

I mean if you do federation for example. The impossible username would look like: `testing username with spaces@yourdomain.com` How would that even be able to click or link in a web browser.

Users that does that anyway will not benefit from this PR.

szaimen:
> Whut! Now I would call that a bug! It shouldn't be possible to add usernames with spaces in my opinion.
> 
> I mean if you do federation for example. The impossible username would look like: `testing username with spaces@yourdomain.com` How would that even be able to click or link in a web browser.
> 
> Users that does that anyway will not benefit from this PR.

Yes, you are right. Maybe this should get reported in the server repo?

enoch85:
> Maybe this should get reported in the server repo?

Most likely yes. If I'm wrong then at least one would get an explanation.

If you have time, please file a bug, I have a support session in 7 minutes.

enoch85:
Well, here's the result when you try to login with that user, so yes it's a bug:

Screenshot from 2020-01-13 18:56:15

szaimen:
> If you have time, please file a bug

I've done that 👍

szaimen:
BTW: this would be a really awesome addition to this feature (since I am not logged in to my admin-account all the time: just use the regular account as my normal account and because of that cannot get the admin notifications.)
https://github.com/nextcloud/notifications/issues/314

enoch85:
We could make a new function with the same stuff that would send to all users, but.

Imagine this: You have 500 users, do you really think that all 500 users should get a notification? At least I don't think so. :) It would take quite some time to run the function as well, because it loops through all users.

szaimen:
No, you are right. Maybe instead we could let the admin configure manually one or more users where to send notifications to?
This could then get added to the array in the notification function. 
What do you think about that idea?

enoch85:
The thing is, this VM is "setup and forget". You and me are maybe willing to maintain it, but 99% of the users actually don't care. It should "just work". So adding new features where we expect the end-user to maintain stuff is out of scope for this VM if you ask me. 

We should instead build something that doesn't need maintaining and "just works", hence the update script, the app scripts, automated updates and so on.

> I would just wonder then, where to configure this option...

I wonder when that option is configured, will it ever be touched again? Probably not. It would happen in the first setup and the user would forget about it, then send an issue report when it's broken because the users that it was setup for no longer exists - or in general - users that were setup 2 years ago (when the VM was setup for the first time) aren't the ones who should get the notifications anymore...

szaimen:
Okay. Then we leave it just as it is 👍

szaimen:
BTW, I recently discussed the function with my brother and he suggested to do it like this:
```
notify_user_gui() {
USER=$(occ_command user:list | awk '{print $2}' | cut -d ":" -f1;)
print_text_in_color "$ICyan" "Posting notification to users that are admins, this might take a while..."
for user in $USER
do
    if occ_command user:info "$user" | grep -q "\- admin";
    then
        print_text_in_color "$ICyan" "Posting to: $user"
        occ_command notification:generate -l "$2" "$admin" "$1"
    fi
done
}
```
Then you don't need two `for` loops and therefore it should be faster and use less memory.

enoch85:
Make a new PR and we'll discuss it.

Thanks!

PR #1033, opened 2020-01-14
Remove IP version preference code

staticsafe:
There is no discernible and good reason that the user's system IP version preference should be modified from the default and especially not without notifying the user that such a drastic change was made to their system.

In addition, if this was added to solve some sort of networking issue with the user's system, that is outside the scope of a script like this.

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -2,9 +2,6 @@
 
 # T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
 
-# Prefer IPv4
-sed -i "s|#precedence ::ffff:0:0/96  100|precedence ::ffff:0:0/96  100|g" /etc/gai.conf
-
 # Install curl if not existing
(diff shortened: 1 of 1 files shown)


enoch85:
Hi,

Thanks for your PR!

As you can read in the README the installation happens in two sets of scripts. The second script [sets it back to the original](https://github.com/nextcloud/vm/blob/master/nextcloud-startup-script.sh#L679) setting at the end. It's just a preference, nothing hard coded.

The script is developed to work with IPv4 and that's why I added that at the top. It has nothing to do with any Dev-environment. 

Sorry, but I will not merge this PR.

staticsafe:
My bad, I missed the README and the second script. Could you please explain why the script needs to work with IPv4 by changing gai.conf (even temporarily for runtime duration)?

enoch85:
> Could you please explain why the script needs to work with IPv4 by changing gai.conf 

It has to do with the functions regarding the network, checks and so forth. I don't remember 100% to be honest, but I remember that when I removed it, weird stuff started to happen.

PR #1034, opened 2020-01-15
improve the notification function

szaimen:
> If you do it like this, you don't need two for loops and no array and therefore it should be faster and use less memory.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -1088,21 +1088,14 @@ esac
 # occ_command notification:generate -l "$2" "$admin" "$1"
 notify_user_gui() {
 USER=$(occ_command user:list | awk '{print $2}' | cut -d ":" -f1;)
-print_text_in_color "$ICyan" "Looping through users, this might take a while..."
+print_text_in_color "$ICyan" "Posting notification to users that are admins, this might take a while..."
 for user in $USER
 do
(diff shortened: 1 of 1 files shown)


enoch85:
Please make it:

```
notify_user_gui() {
USERS=$(occ_command user:list | awk '{print $2}' | cut -d ":" -f1;)
print_text_in_color "$ICyan" "Posting notification to users that are admins, this might take a while..."
for user in $USERS
do
    if occ_command user:info "$user" | grep -q "\- admin";
    then
        print_text_in_color "$IGreen" "Posting '$1' to: $user"
        occ_command notification:generate -l "$2" "$user" "$1"
    fi
done
}
```

szaimen:
> Please make it:
> 
> ```
> notify_user_gui() {
> CHECK_USERS=$(occ_command user:list | awk '{print $2}' | cut -d ":" -f1;)
> print_text_in_color "$ICyan" "Posting notification to users that are admins, this might take a while..."
> for admin in $CHECK_USERS
> do
>     if occ_command user:info "$admin" | grep -q "\- admin";
>     then
>         print_text_in_color "$IGreen" "Posting '$1' to: $admin"
>         occ_command notification:generate -l "$2" "$admin" "$1"
>     fi
> done
> }
> ```

This should be done 👍

Issue #1029, opened 2020-01-11
Disabling the web-updater by default?


szaimen:
What about disabling the web-updater by default?
Since the web-updater didn't work the last time I've tried it with Nextcloud VM, what about disabling it by default?
From the Nextcloud manual:
![image](https://user-images.githubusercontent.com/42591237/72211333-d7b11900-34c9-11ea-8fcf-2fd8a5d479b6.png)

enoch85:
Hmm, never thought of that. Wouldn't that make the update notification disabled as well?

I would like to have the VMs own notify system. It's possible, but some code is missing in the server still. Maybe next project for you? https://github.com/nextcloud/vm/pull/886/files

szaimen:
I've tested the config option and it doesn't disable the web updater.
But I found this: https://github.com/nextcloud/server/issues/18585

> Hmm, never thought of that. Wouldn't that make the update notification disabled as well

Probably not.

> I would like to have the VMs own notify system. It's possible, but some code is missing in the server still. Maybe next project for you? https://github.com/nextcloud/vm/pull/886/files

Maybe after smbshare is merged.

szaimen:
This is probably how the Netflix snap handles this:
https://github.com/nextcloud/nextcloud-snap/commit/31a623cc5d77f9ddbc1473076c027d13c0f6baeb

enoch85:
@oparoz @kyrofa Hmm, how would you handle that in bash? Just create `ncpath/htdocs/` and then put the same files as you did in there?

kyrofa:
Can you just `occ app:disable`? We do that [with theming](https://github.com/nextcloud/nextcloud-snap/blob/master/src/nextcloud/fixes/fresh-install/1_disable-theming.sh#L9), for example.

szaimen:
Unfortunately not, since it's not a normal app:
If I do `occ app:list`, there is no updater app listed but it is definitely installed and working in the Nextcloud VM.

kyrofa:
Ah. Effectively all the `organize` does is `rm -rf` the app. Might as well just do that.

szaimen:
Ah okay, I get it know 👍
So this would probably not work for the Nextcloud VM since you would loose the update-notifications. We just wanted to disable the web-updater.
But thank you for your fast and helpful reply, @kyrofa!

szaimen:
@enoch85 here another idea:
What about blocking the url http://cloud.domain.tld/updater ?
[...]

PR #1035, opened 2020-01-16
Notify admins about updates in GUI

enoch85:
Fix for https://github.com/nextcloud/vm/issues/1029

Screenshot from 2020-01-16 14:53:21

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -390,7 +390,7 @@ crontab -u www-data -l | { cat; echo "*/5  *  *  *  * php -f $NCPATH/cron.php > 
 # Run the updatenotification on a schelude ##TODO disable Nextclouds updater here 
 download_static_script updatenotification.sh
 check_command chmod +x "$SCRIPTS"/updatenotification.sh
-crontab -u root -l | { cat; echo "*  *  *  *  */3 bash $SCRIPTS/cron.php > /dev/null 2>&1"; } | crontab -u root -
+crontab -u root -l | { cat; echo "05  12  *  *  */3 bash $SCRIPTS/cron.php > /dev/null 2>&1"; } | crontab -u root -
 
 # Change values in php.ini (increase max file size)
 # max_execution_time
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -387,6 +387,11 @@ echo
 # Prepare cron.php to be run every 15 minutes
 crontab -u www-data -l | { cat; echo "*/5  *  *  *  * php -f $NCPATH/cron.php > /dev/null 2>&1"; } | crontab -u www-data -
 
+# Run the updatenotification on a schelude ##TODO disable Nextclouds updater here 
+download_static_script updatenotification.sh
+check_command chmod +x "$SCRIPTS"/updatenotification.sh
+crontab -u root -l | { cat; echo "*  *  *  *  */3 bash $SCRIPTS/cron.php > /dev/null 2>&1"; } | crontab -u root -
+
 # Change values in php.ini (increase max file size)
 # max_execution_time
 sed -i "s|max_execution_time =.*|max_execution_time = 3500|g" "$PHP_INI"
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -392,7 +392,7 @@ occ_command config:system:set upgrade.disable-web --value="true"
 print_text_in_color "$ICyan" "Configuring update notifications specific for this server..."
 download_static_script updatenotification.sh
 check_command chmod +x "$SCRIPTS"/updatenotification.sh
-crontab -u root -l | { cat; echo "05  12  *  *  */3 bash $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
+crontab -u root -l | { cat; echo "59 $AUT_UPDATES_TIME * * 6 $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
 
 # Change values in php.ini (increase max file size)
 # max_execution_time
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -387,10 +387,12 @@ echo
 # Prepare cron.php to be run every 15 minutes
 crontab -u www-data -l | { cat; echo "*/5  *  *  *  * php -f $NCPATH/cron.php > /dev/null 2>&1"; } | crontab -u www-data -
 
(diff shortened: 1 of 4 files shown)


enoch85:
cc @szaimen

enoch85:
This would run weekly on Wednesdays

nextcloud_install_production.sh
```diff
@@ -387,6 +387,11 @@ echo
 # Prepare cron.php to be run every 15 minutes
 crontab -u www-data -l | { cat; echo "*/5  *  *  *  * php -f $NCPATH/cron.php > /dev/null 2>&1"; } | crontab -u www-data -
 
+# Run the updatenotification on a schelude ##TODO disable Nextclouds updater here 
+download_static_script updatenotification.sh
+check_command chmod +x "$SCRIPTS"/updatenotification.sh
+crontab -u root -l | { cat; echo "05  12  *  *  */3 bash $SCRIPTS/cron.php > /dev/null 2>&1"; } | crontab -u root -
```
enoch85: need to check if */3 is correct, maybe just 3 is enough

enoch85: This would run every 3:d day, which is fine with me.

enoch85:
@kyrofa So this would be enough you think?

szaimen:
Some thoughts:

- If I understand the update-notification app  correctly, it also checks for app-updates. So could we also check for app updates? Or isn't that something important?
- If a user has installed automatic updates during the startup-script, then he would get informed on a wednesday, that there is a new update available and that he should run the update script, although the update would get installed during the next sunday automatically?
- And if the script only checks for nextcloud updates, and informs just if nextcloud-updates are available, (and if it also updates apps) then you would get app-updates just around every 4 weeks or so if a new nextcloud-update is available, what wouldn't be ideal imo.
- as an idea: maybe we should seperate app-updates from the main script for that usecase? That can run automatically regardless of whether you have enabled automatic nextcloud updates...

enoch85:
>    what about removing the upgrade.disable-web app completely? Then no user would get the bad idea to reenable the app again.

I'm not sure actually... If all notifications that are made in the GUI are sent by email, then I would like to keep it that way. If it doesn't, then it doesn't matter anyway, and in that case I'd say delete the app.

>    If I understand the update-notification app correctly, it also checks for app-updates. So could we also check for app updates? Or isn't that something important?

I don't know tbh. App-updates are made with the VM script anyway - on each run. I think it would be strange if it didn't show which apps that have updates in the App-store. To me that would be sufficient. I'm also not sure if it's possible to check for app updates with `occ`?

 >   If a user has installed automatic updates during the startup-script, then he would get informed in the worst situation directly after the run of the nextcloud update script, that there is a new update available and that he should run the update script, although the update would get installed during the next sunday automatically?

In a race-condition, yes. So, if the update are released on Sunday @ 10:15 (as an example) and the Automatic is scheduled for 18:15, then the user would get a notification at 12:05 that there's an update available. 

Maybe add some text that users with automatic updates enabled don't have to bother? Or even better, add a notification about that Nextcloud automatically has been updated in the function for automatic updates?

 >   And if the script only checks for nextcloud updates, and informs just if nextcloud-updates are available, (and if it also updates apps) then you would get app-updates just around every 4 weeks or so if a new nextcloud-update is available, what wouldn't be ideal imo.

You would still be able to go to the GUI and manually update all the apps in the App-store. The notification doesn't "enable" app-updates, it just informs.

 >   as an idea: maybe we should seperate app-updates from the main script for that usecase? That can run automatically regardless of whether you have enabled automatic nextcloud updates...

In my opinion there are no need for checking for app updates, but if it's easy to implement, why not.

szaimen:
> > what about removing the upgrade.disable-web app completely? Then no user would get the bad idea to reenable the app again.
> 
> I'm not sure actually... If all notifications that are made in the GUI are sent by email, then I would like to keep it that way. If it doesn't, then it doesn't matter anyway, and in that case I'd say delete the app.

Sry, I have made a mistake here:
disabling the upgrade.disable-web is good,
what is missing here is: disabling the updatenotification app by: 
`occ app:disable updatenotification`
and we should discuss whether removing this app completely would be good.
This is probably what kyrofa meant with this comment: https://github.com/nextcloud/vm/issues/1029#issuecomment-574834586

enoch85:
> occ app:disable updatenotification

Would that disable emails as well?

szaimen:
how app updates could work is the following:
maybe the app-update-script could run automatically, and then afterwards infrom the admin which apps got upgraded?
Or is that the wrong way because the admin cannot longer decide?
As you have stated before: it is afaik not possible to check via occ which apps have updates available.

szaimen:
This is an example output when updating all:

szaimen:
> > occ app:disable updatenotification
> 
> Would that disable emails as well?

no, it shouldn't.
I have run the snap for years with disabled update notifications and email notifications always worked

enoch85:
Check picture in first post. :D

szaimen:
> Check picture in first post. :D

looks good. But I would just send a notification to the admins (when automatic updates on) if nextcloud got upgraded or apps got upgraded. two notifications for an available update seems too much to me.

enoch85:
> so you could grep the app names and inform the admins which apps got upgraded...

Let's not over complicate stuff here. :)

I feel satisfied with this as is. Anything else you'd like to add?

szaimen:
@enoch85 please check my post above:
https://github.com/nextcloud/vm/pull/1035#issuecomment-575163862

szaimen:
i would also like to see something like this:
send a notification that app updates got installed if [ $(occ_command app:update --all) != "" ]

enoch85:
> end a notification that app updates got installed if [ $(occ_command app:update --all) != "" ]

We update all apps in the update script, if the user wants to update them manually in the App Store, then fine. What I think is important is letting the user know about the Nextcloud update.

enoch85:
> Also since the upgrade job is running only once a week, but the notification is happening every 3 days, it could happen that you get 6 messages for this available upgrade before it gets automatically installed.

Good point, maybe only check once a week then, right after automatic updates are run.

szaimen:
> > end a notification that app updates got installed if [ $(occ_command app:update --all) != "" ]
> 
> We update all apps in the update script, if the user wants to update them manually in the App Store, then fine. What I think is important is letting the user know about the Nextcloud update.

Yes, but I think it would be useful to let the user know that app updates got installed. So if it runs once a week that you get each time a notification if app updates where installed and only an addition if also nextcloud was updated.

enoch85:
Is this enough? or do you still want to show which apps that got updated?

enoch85:
I think we can merge this and fix app updates notifications in another PR if you feel like it.

Issue #1037, opened 2020-01-16
Notify about app updates 


enoch85:
@szaimen If you feel like it, please create a PR for app updates as well.

As we discussed in https://github.com/nextcloud/vm/pull/1035

szaimen:
As a note to myself:
occ app:disable updatenotification
send a notification that app updates got installed if [ $(occ_command app:update --all) != "" ]

enoch85:
`if [ -z stuff ]` checks if a $var is empty btw, `if [ -n stuff ]` checks if it's not empty.

PR #1038, opened 2020-01-16
Notify about app updates and remove update notification

szaimen:
fix https://github.com/nextcloud/vm/issues/1037

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -389,6 +389,8 @@ crontab -u www-data -l | { cat; echo "*/5  *  *  *  * php -f $NCPATH/cron.php > 
 
 # Run the updatenotification on a schelude
 occ_command config:system:set upgrade.disable-web --value="true"
+occ_command app:disable updatenotification
+rm -r $NCPATH/apps/updatenotification
 print_text_in_color "$ICyan" "Configuring update notifications specific for this server..."
 download_static_script updatenotification.sh
 check_command chmod +x "$SCRIPTS"/updatenotification.sh
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -318,11 +318,13 @@ then
     UPDATED_APPS=$(sed '1~2d' Updated_apps | awk '{print $1}')
     for apps in $UPDATED_APPS
     do
-        CHANGE_FORMATTING+="$apps,"
+        CHANGE_FORMATTING+="$apps"
     done
+    CHANGE_FORMATTING=${CHANGE_FORMATTING[*]}
+    CHANGE_FORMATTING=${CHANGE_FORMATTING// /", "}
     notify_admin_gui \
     "App Updates got installed" \
-    "The following apps where updated: ${CHANGE_FORMATTING[*]}"
+    "The following apps where updated: $CHANGE_FORMATTING"
 fi
 
 # Nextcloud 13 is required.
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -309,7 +309,20 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
-    occ_command app:update --all
+    occ_command app:update --all >> Updated_Apps
+fi
+
+# Check which apps got updated
+if [ -n $(occ_command app:update --all) ]
+then
+    UPDATED_APPS=$(sed '1~2d' Updated_apps | awk '{print $1}')
+    for apps in $UPDATED_APPS
+    do
+        CHANGE_FORMATTING+="$apps,"
+    done
+    notify_admin_gui \
+    "App Updates got installed" \
+    "The following apps where updated: ${CHANGE_FORMATTING[*]}"
 fi
 
 # Nextcloud 13 is required.
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -316,6 +316,7 @@ fi
 # Check which apps got updated
 if [ -n "$UPDATED_APPS" ]
 then
+    UPDATED_APPS=${UPDATED_APPS:0:-2}
     notify_admin_gui \
     "App Updates got installed" \
     "The following apps were updated: $UPDATED_APPS"
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -322,7 +322,7 @@ then
     "You've got app updates!" \
     "$UPDATED_APPS"
 else
-    print_text_in_color "$IGreen" "It seems like everything is up-to-date!"
+    print_text_in_color "$IGreen" "Your apps are already up to date!"
 fi
 
 # Nextcloud 13 is required.
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -309,17 +309,17 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
-    UPDATED_APPS=$(occ_command app:update --all | sed '1~2d' | awk '{print $1}' | tr '\n' " ")
-    UPDATED_APPS=${UPDATED_APPS// /", "}
+    # occ command can not be used due to the check_command() function.
+    UPDATED_APPS="$(sudo -u www-data php $NCPATH/occ app:update --all)"
 fi
 
 # Check which apps got updated
 if [ -n "$UPDATED_APPS" ]
 then
-    UPDATED_APPS=${UPDATED_APPS:0:-2}
+    print_text_in_color "$IGreen" "$UPDATED_APPS"
     notify_admin_gui \
-    "App Updates got installed" \
-    "The following apps were updated: $UPDATED_APPS"
+    "You have app updates!" \
+    "$UPDATED_APPS"
 fi
 
 # Nextcloud 13 is required.
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -309,22 +309,16 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
-    occ_command app:update --all >> Updated_Apps
+    UPDATED_APPS=$(occ_command app:update | sed '1~2d' | awk '{print $1}' | tr '\n' " ")
+    UPDATED_APPS=${UPDATED_APPS// /", "}
 fi
 
 # Check which apps got updated
-if [ -n $(occ_command app:update --all) ]
+if [[ "$UPDATED_APPS" == @([a-z]) ]]
 then
-    UPDATED_APPS=$(sed '1~2d' Updated_apps | awk '{print $1}')
-    for apps in $UPDATED_APPS
-    do
-        CHANGE_FORMATTING+="$apps"
-    done
-    CHANGE_FORMATTING=${CHANGE_FORMATTING[*]}
-    CHANGE_FORMATTING=${CHANGE_FORMATTING// /", "}
     notify_admin_gui \
     "App Updates got installed" \
-    "The following apps where updated: $CHANGE_FORMATTING"
+    "The following apps where updated: $UPDATED_APPS"
 fi
 
 # Nextcloud 13 is required.
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -309,6 +309,7 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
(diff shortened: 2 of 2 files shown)


nextcloud_update.sh
```diff
@@ -309,7 +309,22 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
-    occ_command app:update --all
+    occ_command app:update --all >> Updated_Apps
```
enoch85: One `>` would be enough here as it only is valid one time.

nextcloud_install_production.sh
```diff
@@ -389,6 +389,8 @@ crontab -u www-data -l | { cat; echo "*/5  *  *  *  * php -f $NCPATH/cron.php >
 
 # Run the updatenotification on a schelude
 occ_command config:system:set upgrade.disable-web --value="true"
+occ_command app:disable updatenotification
+rm -r $NCPATH/apps/updatenotification
```
enoch85: Are you 100% this doesn't disable email notifications when there's a new notification in the GUI?

nextcloud_update.sh
```diff
@@ -309,7 +309,22 @@ fi
...
+fi
+
+# Check which apps got updated
+if [ -n $(occ_command app:update --all) ]
```
enoch85: Better to put everything in one place, so execute `occ_command app:update --all >> Updated_Apps` here instead - in the first `if` statement

nextcloud_update.sh
```diff
@@ -309,7 +309,22 @@ fi
...
+then
+    UPDATED_APPS=$(sed '1~2d' Updated_apps | awk '{print $1}')
+    for apps in $UPDATED_APPS
+    do
+        CHANGE_FORMATTING+="$apps"
+    done
+    CHANGE_FORMATTING=${CHANGE_FORMATTING[*]}
+    CHANGE_FORMATTING=${CHANGE_FORMATTING// /", "}
```
enoch85: Please explain what this does exactly, maybe there's a better way.

enoch85:
```
Found linux image: /boot/vmlinuz-4.15.0-74-generic
Found initrd image: /boot/initrd.img-4.15.0-74-generic
Found linux image: /boot/vmlinuz-4.15.0-72-generic
Found initrd image: /boot/initrd.img-4.15.0-72-generic
done
sed: can't read Updated_apps: No such file or directory
nextcloud_update.sh: line 326: notify_admin_gui: command not found
Checking for new Nextcloud version...
You already run the latest version! (17.0.2)
```

szaimen:
> ```
> Found linux image: /boot/vmlinuz-4.15.0-74-generic
> Found initrd image: /boot/initrd.img-4.15.0-74-generic
> Found linux image: /boot/vmlinuz-4.15.0-72-generic
> Found initrd image: /boot/initrd.img-4.15.0-72-generic
> done
> sed: can't read Updated_apps: No such file or directory
> nextcloud_update.sh: line 326: notify_admin_gui: command not found
> Checking for new Nextcloud version...
> You already run the latest version! (17.0.2)
> ```

I am working on it

nextcloud_update.sh
szaimen: i changed it so not anymore relevant

nextcloud_update.sh
szaimen: this is done

nextcloud_update.sh
szaimen: i will do it on the new algorithm

nextcloud_update.sh
```diff
@@ -309,7 +309,16 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
-    occ_command app:update --all
+    UPDATED_APPS=$(occ_command app:update | sed '1~2d' | awk '{print $1}' | tr '\n' " ")
+    UPDATED_APPS=${UPDATED_APPS// /", "}
```
szaimen: What this does is, it deletes every second line from the output, prints the first element every line and substitues the linebreaks through spaces. Afterwards all spaces get substituted by ", " which should result in the following formatting:
app1, app2, app3

nextcloud_update.sh
```diff
@@ -309,7 +309,16 @@ fi
...
+fi
+
+# Check which apps got updated
+if [[ "$UPDATED_APPS" == @([a-z]) ]]
```
szaimen: this checks whether updated_apps contains a a-z character, which should always be the case if apps got updated.

nextcloud_install_production.sh
szaimen: As I stated here: https://github.com/nextcloud/vm/pull/1035#issuecomment-575159638

> > > occ app:disable updatenotification
> > 
> > 
> > Would that disable emails as well?
> 
> no, it shouldn't.
> I have run the snap for years with disabled update notifications and email notifications always worked
> 
> to be more specific:
> activity notifications with email always worked with the snap
> but if there are email notifications sent for nextcloud or app updates, then these would probably don't work.

nextcloud_update.sh
szaimen: This unfortunately doesn't work yet. We probably have to append the output of occ_command to a file. Do you have an idea how to get all of this working for a file?
afterwards the result has to get put into a variable so that we can check whether we can send the notification

szaimen: Basically what I wanted to do was getting the app names from the output and posting them to the admins.
But if that is too complicated maybe it is enough to post to the admins, that app updates got installed. What do you think?

szaimen: Please forget what I have said. It is working

nextcloud_update.sh
```diff
@@ -309,7 +309,17 @@ fi
...
 then
-    occ_command app:update --all
+    UPDATED_APPS=$(occ_command app:update --all | sed '1~2d' | awk '{print $1}' | tr '\n' " ")
+    UPDATED_APPS=${UPDATED_APPS// /", "}
+fi
+
+# Check which apps got updated
+if [ -n "$UPDATED_APPS" ]
+then
+    UPDATED_APPS=${UPDATED_APPS:0:-2}
```
szaimen: this removes the latest two characters, in the string, which is needed because else you had a comma at the end which wouldn't be right. It needs to be here, because if the string is empty, then you get an error.

szaimen:
@enoch85 this should be ready for review.

enoch85:
I haven't tested but the final fix looks cleaner.

The only thing I'm worried about here is that what happens if the text of the updatd apps change wording, then the hard coded `sed '1~2d' | awk '{print $1}' | tr '\n' " ")` would fail, resulting in that no apps got updated and the script would stop completley, as `occ_command` contains `check_command.`

I don't have an exact number, but I'd guess that around 100 000 users use this script. It's also run in automatic updaes each week. Of all the scripts we have, this is the most sensitive.

szaimen:
> I haven't tested but the final fix looks cleaner.
> 
> The only thing I'm worried about here is that what happens if the text of the updatd apps change wording, then the hard coded `sed '1~2d' | awk '{print $1}' | tr '\n' " ")` would fail, resulting in that no apps got updated and the script would stop completley, as `occ_command` contains `check_command.`
> 
> I don't have an exact number, but I'd guess that around 100 000 users use this script. It's also run in automatic updaes each week. Of all the scripts we have, this is the most sensitive.

I have tested it and it works.
Are you sure, that this would fail, if the output would change, since the only command that should get checked by check_command is `app:update --all ` or am I wrong with that?
Maybe could you explain what check_command exactly does? I don't understand it completely.

szaimen:
@enoch85 any thoughts on my comment above?

enoch85:
> Maybe could you explain what check_command exactly does?

`if ! "$@";` simply means that if something in @ (everything) after `check_command` is executed fails, then it will exit with an error. That means everything, so: `check_command "function $1 text and everything else"` is included.

> I have tested it and it works. 

I believe you. :) Problem is just because it works now, it might fail later, since it's based on hardcoded checks.

I will test this and then consider merging.

szaimen:
> > Maybe could you explain what check_command exactly does?
> 
> `if ! "$@";` simply means that if something in @ (everything) after `check_command` is executed fails, then it will exit with an error. That means everything, so: `check_command "function $1 text and everything else"` is included.

Thanks for explaining! I understand it now better!

> 
> > I have tested it and it works.
> 
> I believe you. :) Problem is just because it works now, it might fail later, since it's based on hardcoded checks.
> 
> I will test this and then consider merging.

Thanks, I am excited to see the result :)

enoch85:
Something like this wold post the whole output instead. Safer in my opinion:

```
# Update all Nextcloud apps
if [ "${CURRENTVERSION%%.*}" -ge "15" ]
then
    UPDATED_APPS=$(occ_command app:update --all)
fi

# Check which apps got updated
if [ -n "$UPDATED_APPS" ]
then
    UPDATED_APPS=${UPDATED_APPS:0:-2}
    notify_admin_gui \
    "App updates got installed" \
    "$UPDATED_APPS"
    print_text_in_color "$IGreen" "$UPDATED_APPS"
fi
```

It doesn't work in current state, but something to work on at least. It would remove the risk of failure due to new wording, and it would also post useful info such as if an app couldn't be updated, No hardcoded stuff either, just the "log" straight up and down.

szaimen:
> # Update all Nextcloud apps
> if [ "${CURRENTVERSION%%.*}" -ge "15" ]
> then
>     # occ command can not be used due to the check_command() function.
>     UPDATED_APPS="$(sudo -u www-data php $NCPATH/occ app:update --all)"
> fi
> 
> # Check which apps got updated
> if [ -n "$UPDATED_APPS" ]
> then
>     notify_admin_gui \
>     "You have app updates!" \
>     "$UPDATED_APPS"
>     print_text_in_color "$IGreen" "$UPDATED_APPS"
> fi

What would you like to see changing here?
I am okay with that and it works quite well in my testing...
I would just print_text_in_color before notifying all admins about the updated apps.

enoch85:
What about this:

```
if [ "${CURRENTVERSION%%.*}" -ge "15" ]
then
    print_text_in_color "$ICyan" "Trying to automatically update all Nextcloud apps..."
    # occ command can not be used due to the check_command() function.
    UPDATED_APPS="$(sudo -u www-data php $NCPATH/occ app:update --all)"
fi

# Check which apps got updated
if [ -n "$UPDATED_APPS" ]
then
    print_text_in_color "$IGreen" "$UPDATED_APPS"
    notify_admin_gui \
    "You've got app updates!" \
    "$UPDATED_APPS"
else
    print_text_in_color "$IGreen" "It seems like everything is up-to-date!"
fi
```

enoch85:
Screenshot_20200118-020836

enoch85:
I haven't disabled and removed the updatenotificatio app. So I get it on mobile as well, which is preferable. 

Please remove the changes in `nextcloud_install_production.sh` Thanks.

enoch85:
OK, time for bed. Later!

szaimen:
> I haven't disabled and removed the updatenotificatio app. So I get it on mobile as well, which is preferable.
> 
> Please remove the changes in `nextcloud_install_production.sh` Thanks.

have you tested the same with disabled upgrade notification app?
If you don't get the updates there, then I would remove the changes to the upgrade notification app.

enoch85:
Mobile notifications still work even after disabling the updatenotification app.

enoch85:
OK, let's merge this then. :)

enoch85:
@szaimen Just saw that Nextcloud 18 is released. Anything more you want to add to that release, or are we done with this now?

szaimen:
Maybe putting at least the apps main menu into a new script? 
https://github.com/nextcloud/vm/issues/1041

enoch85:
> @enoch85 Maybe putting the apps main menu into a new script?
> #1041

Nah, I don's see the benefits of that, one more thing that could go wrong (fetching the script), let's try to keep the point-of-failures as small as possible.

I will hold of until tomorrow for a new release, then if no bugs are reported - NC 18 here we come!

enoch85:
@oparoz @jospoortvliet Can you please invite @szaimen to @nextcloud/vm?

nextcloud_install_production.sh
```diff
@@ -389,6 +389,8 @@ crontab -u www-data -l | { cat; echo "*/5  *  *  *  * php -f $NCPATH/cron.php >
 
 # Run the updatenotification on a schelude
 occ_command config:system:set upgrade.disable-web --value="true"
+occ_command app:disable updatenotification
```
szaimen: @enoch85 this is unneccessary, because it gets restored at least after every major update.
we could also remove it after every update, but maybe it is enough to just get disabled. What do you think?

Issue #1043, opened 2020-01-18
VM Update Failure


ChuckNozzle:
## Steps to reproduce
1. Issued command: sudo bash /var/scripts/update.sh
2. Slow internet connectivity resulted in longer than normal download.
3. Update was left to continue, and on return, the update had failed.

## Expected behaviour
Removal of old Nextcloud instance and installation of new instance.

## Actual behaviour
Old instance was removed, but it seems that the newly downloaded instance had either not downloaded correctly, or was corrupted, possibly due to slow internet connectivity.
Image attached showing what was on screen.

**Nextcloud server version:** Was in process of upgrading from 17.? to 18.0

**Server OS:** Ubuntu 18.04

**How did you install the VM?** Released Version

## Network
**Do you use DHCP?** Yes. Static Lease set in Router

**Is port 80 and/or 443 open?** Yes

## Logs / Screenshots
![image](https://user-images.githubusercontent.com/29411770/72664775-99a47f80-3a0a-11ea-9155-d490fa4773d8.png)

enoch85:
@szaimen could you please have a look? I'm abroad and right now without a laptop, and this seems quite urgent.

Thanks!

szaimen:
@enoch85 I would like to but I don't know if I am skilled enough for that.
BTW it seems like the download was corrupted, so are you comparing the checksums in the update script? This should prevent such problems.

@ChuckNozzle do you have an external backup from you instance and can restore it in the meantime?

ChuckNozzle:
I don’t have a backup for the VM, but I have been considering deploying a fresh VM for a while now.
If I can transfer the database and apps etc, then I am happy to do that, I just need to know what I need to transfer to the new VM.
The main reason for reporting this was the fact that the update didn’t pick up the corruption, and then to find out how to migrate to a new VM.

szaimen:
@ChuckNozzle concerning the backup&restore/migration: the only thing I can currently guide you to is https://github.com/nextcloud/vm/issues/999

For everything else, @enoch85 should answer, or you ask in the help forum: https://help.nextcloud.com/c/support/appliances-docker-snappy-vm

ChuckNozzle:
Had a look, and that seems to be the info I need.
Thanks!

enoch85:
It sure seems like the slow link caused a corruption. Probably due to heavy load on the Nextcloud servers, in combination with the slow link.

I made a PR to at least try to avoid it. Please test if it works @szaimen

[...]

PR #1044, opened 2020-01-18
Possible fix for #1043

enoch85:
Fixes https://github.com/nextcloud/vm/issues/1043

1. Abort if the extraction fails'
2. Check that network is OK before downloading.

Needs serious testing.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -703,6 +703,7 @@ fi
 }
 
 download_verify_nextcloud_stable() {
+# Check the current Nextcloud version
 while [ -z "$NCVERSION" ]
 do
     print_text_in_color "$ICyan" "Fetching the latest Nextcloud version..."
@@ -710,6 +711,7 @@ do
     STABLEVERSION="nextcloud-$NCVERSION"
     print_text_in_color "$IGreen" "$NCVERSION"
 done
+
 # Download the file
 rm -f "$HTML/$STABLEVERSION.tar.bz2"
 cd $HTML
@@ -722,15 +724,14 @@ else
     exit 1
 fi
 
-# Checksum the downloaded file
+# Checksum of the downloaded file
 print_text_in_color "$ICyan" "Checking SHA256 checksum..."
 mkdir -p "$SHA256_DIR"
-install_if_not coreutils
 curl_to_dir "$NCREPO" "$STABLEVERSION.tar.bz2.sha256" "$SHA256_DIR"
 SHA256SUM="$(cat $SHA256_DIR/$STABLEVERSION.tar.bz2.sha256 | awk '{print$1}')"
-if echo "$SHA256SUM" "$STABLEVERSION.tar.bz2" | sha256sum -c
+if ! echo "$SHA256SUM" "$STABLEVERSION.tar.bz2" | sha256sum -c
 then
-    msg_box "The checksums of $STABLEVERSION.tar.bz2 didn't match, please try again."
+    msg_box "The SHA256 checksums of $STABLEVERSION.tar.bz2 didn't match, please try again."
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


szaimen:
So I have updated a test VM successfull to the new version. But it is not my productive instance and I don't have a slow internet connection. So this might be not representative for slow internet connections and large instances.

lib.sh
```diff
@@ -715,7 +715,13 @@ cd $HTML
 print_text_in_color "$ICyan" "Downloading $STABLEVERSION..."
 curl -fSLO --retry 3 "$NCREPO/$STABLEVERSION.tar.bz2"
 mkdir -p "$GPGDIR"
-curl_to_dir "$NCREPO" "$STABLEVERSION.tar.bz2.asc" "$GPGDIR"
+if network_ok
+then
+    curl_to_dir "$NCREPO" "$STABLEVERSION.tar.bz2.asc" "$GPGDIR"
+else
+    msg_box There seems to be an issue with your network, please try again later.\nThis script will exit."
```
szaimen: I think this should be:
`    msg_box "There seems to be an issue with your network, please try again later.\nThis script will exit."`

szaimen:
BTW: I have had an issue there:
if [ "${CURRENTVERSION%%.*}" -ge "15" ]
I think this check doesn't work but I am not completely sure about this.

szaimen:
And isn't there a possibility to directly compare the checksum of the file on the server against the downloaded file? I think this would be more robust... Or is this already implemented?

szaimen:
@enoch85 actually the update got through successfully but there seems something to be really broken with the new sidebar. Don't know if that has to do with the update script.
sometimes the sidear doesnt load at all or if you want to create a share link it doesnt work.

szaimen:
> @enoch85 actually the update got through successfully but there seems something to be really broken with the new sidebar. Don't know if that has to do with the update script.
> sometimes the sidear doesnt load at all or if you want to create a share link it doesnt work.

that was due to my webbrowser. it works now.

szaimen:
The update on my instance was successfull

lib.sh
enoch85:
```suggestion
    msg_box There seems to be an issue with your network, please try again later.\nThis script will exit."
```

Yes, good catch!

enoch85:
> And isn't there a possibility to directly compare the checksum of the file on the server against the downloaded file? I think this would be more robust... Or is this already implemented?

Yeah that would be the best solution.

I will work on something like that now.

enoch85:
> The update on my productive instance was successfull

Did you use this PRs `curl` command in the top? Otherwise you tested master code.

enoch85:
```
SHA256SUM="$(sha256sum $STABLEVERSION.tar.bz2 | awk '{print $1'})"
SHA256SUM_VERIFY="$(cat $STABLEVERSION.tar.bz2.sha256 | awk '{print $1'})"

if "$SHA256SUM_VERIFY" = "$$SHA256SUM"
then
    msg_box "The checksums of the files didn't match, please try again."
    exit
fi
```

This doesen't work yet, but it's a start.

enoch85:
Please test @szaimen 

Thanks!

szaimen:
@enoch85 so I have tested the new update script and it seems to work. (Updated a test VM successfully from 17 to 18) but initially I had an internal server error after the update and had to manually restart the VM to make it work. Shouldn't be a restart considered after executing the update script?

lib.sh
```diff
@@ -12,6 +12,7 @@ HTML=/var/www
 NCDATA=/mnt/ncdata
 SNAPDIR=/var/snap/spreedme
 GPGDIR=/tmp/gpg
+SHA256_DIR=/tmp/shas56
```
szaimen: I am pretty sure, that this should be 
`SHA256_DIR=/tmp/sha256`
But its not an important change.

enoch85: Please add it to the "some improvements PR"

PR #1046, opened 2020-01-19
Some improvements

### LINKED ISSUE #1045: Add notify_admin_gui to check_command
Also remove `occ_command` from `notify_admin_gui`

szaimen:
Fix #1045
- [x] Add notify_admin_gui to check_command
- [x] remove occ_command from notify_admin_gui
- [x] reenable the app updates admin notifications
- [x] create occ_command_no_check function
- [x] revert changes to the updatenotification app
- [x] improve notify_admin_gui based on kyrofas implementation

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -309,7 +309,7 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
-    occ_command_no_check maintenance:mode --off
+    occ_command maintenance:mode --off
     # Check for upgrades
     print_text_in_color "$ICyan" "Trying to automatically update all Nextcloud apps..."
     UPDATED_APPS="$(occ_command_no_check app:update --all)"
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -309,25 +309,22 @@ fi
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
-    # Make sure maintenance:mode isn't activated (it will fail if it is)
-    occ_command maintenance:mode --off
+    occ_command_no_check maintenance:mode --off
     # Check for upgrades
     print_text_in_color "$ICyan" "Trying to automatically update all Nextcloud apps..."
-    # occ command can not be used due to the check_command() function.
-    sudo -u www-data php $NCPATH/occ app:update --all
-#    UPDATED_APPS="$(sudo -u www-data php $NCPATH/occ app:update --all)"
+    UPDATED_APPS="$(occ_command_no_check app:update --all)"
 fi
 
 # Check which apps got updated
-#if [ -n "$UPDATED_APPS" ]
-#then
-#    print_text_in_color "$IGreen" "$UPDATED_APPS"
-#    notify_admin_gui \
-#    "You've got app updates!" \
-#    "$UPDATED_APPS"
-#else
-#    print_text_in_color "$IGreen" "Your apps are already up to date!"
-#fi
+if [ -n "$UPDATED_APPS" ]
+then
+    print_text_in_color "$IGreen" "$UPDATED_APPS"
+    notify_admin_gui \
+    "You've got app updates!" \
+    "$UPDATED_APPS"
+else
+    print_text_in_color "$IGreen" "Your apps are already up to date!"
+fi
 
 # Nextcloud 13 is required.
 lowest_compatible_nc 13
--- a/lib.sh
+++ b/lib.sh
@@ -1118,14 +1118,14 @@ esac
 #
 # occ_command notification:generate -l "$2" "$admin" "$1"
 notify_admin_gui() {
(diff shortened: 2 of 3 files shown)


szaimen:
@enoch85 this should be ready for review.

nextcloud_update.sh
```diff
@@ -309,25 +309,22 @@ fi
...
 then
-    # Make sure maintenance:mode isn't activated (it will fail if it is)
-    occ_command maintenance:mode --off
+    occ_command_no_check maintenance:mode --off
     # Check for upgrades
     print_text_in_color "$ICyan" "Trying to automatically update all Nextcloud apps..."
-    # occ command can not be used due to the check_command() function.
-    sudo -u www-data php $NCPATH/occ app:update --all
-#    UPDATED_APPS="$(sudo -u www-data php $NCPATH/occ app:update --all)"
+    UPDATED_APPS="$(occ_command_no_check app:update --all)"
```
szaimen: Can't we use here the normal occ_command?

szaimen:
BTW: I am not sure if  `Add notify_admin_gui to check_command` really makes sense

enoch85:
> > Add notify_admin_gui to check_command
> 
> BTW: I am not sure if that really makes sense

If there are several admins, they might want to know whats going on, also we would get more issue reports to fix bugs, which is a good thing.

enoch85:
- [ ] Check if sha256sum is installed before checking sum of the downloaded Nextcloud  tar.

nextcloud_update.sh
enoch85: > occ command can not be used due to the check_command() function.

No :)

szaimen:
We could add a check to notify_admin_gui to check whether the notification app is enabled.
and btw: this is how kyrofa has implemented the admin notifications:
https://github.com/nextcloud/nextcloud-snap/pull/1221/files#diff-d98160f676559fde3039f591da87ca8f

szaimen:
We could add a check to notify

szaimen:
We could add a check to notify

szaimen:
Sry, pressed the wrong button

szaimen:
Actually, I would vote to revert the changes made to the upgrade notification app:
it is stilll quite useful: it shows you which apps aren't compatible with the next version and you can see, if the next major release is already stable:
which isn't the case in the moment, although it is released as stable.
(if you check for updates in the stable channel, it doesn't show you the 18.0 release, only if you switch to beta)

enoch85:
> Actually, I would vote to revert the changes made to the upgrade notification app:
it is stilll quite useful: it shows you which apps aren't compatible with the next major version (and need to be updated) and you can see, if the next major release is already stable:
which isn't the case in the moment with 18.0, although it is released as stable.
(if you check for updates in the stable channel, it doesn't show you the 18.0 release, only if you switch to beta)

Please do what you think is best. I noticed today that I got info about new app updates from the notification app anyway, so maybe adding it in the VM is a bit overkill?

I need to sort the tons of emails I have before I can look at this. Sorry.

szaimen:
> Please do what you think is best. I noticed today that I got info about new app updates from the notification app anyway, so maybe adding it in the VM is a bit overkill?
> 
> I need to sort the tons of emails I have before I can look at this. Sorry.

I will do that later on.
The app updates are from the upgrade-notification app. If it is disabled, you don't get anymore.
I don't think, that it is overkill. If you don't install the app updates manually, they get automatically installed by the update script, if enabled. I think it is good to know, if they got installed.

szaimen:
> > Please do what you think is best. I noticed today that I got info about new app updates from the notification app anyway, so maybe adding it in the VM is a bit overkill?
> > I need to sort the tons of emails I have before I can look at this. Sorry.
> 
> I will do that later on.
> The app updates notifications are from the upgrade-notification app. If it is disabled, you don't get them anymore.
> I don't think, that informing the user via script it is overkill. If you don't install the app updates manually, they get automatically installed by the update script, if enabled. I think it is good to know, if they got installed.

the changes to updatenotification are reverted

szaimen:
> We could add a check to notify_admin_gui to check whether the notification app is enabled.
> and btw: this is how kyrofa has implemented the admin notifications in the snap:
> https://github.com/nextcloud/nextcloud-snap/pull/1221/files#diff-d98160f676559fde3039f591da87ca8f
> I assume he has taken the output format json, because he thinks that this output format will be more stable in the future.

What do you think about that?

enoch85:

@kyrofa's implementation seems good to me, please improve our.

szaimen:
> > What do you think about that?
> 
> @kyrofa's implementation seems good to me, please improve our.

I will do that later on. First have to test, if it works in the VM.

szaimen:
> > 
> > 
> > @kyrofa's implementation seems good to me, please improve our.
> 
> I will do that later on. First have to test, if it works in the VM.

This should be done.

szaimen:
@enoch85 what about your checkpoints here: https://github.com/nextcloud/vm/pull/1046#issuecomment-576055408

Issue #1041, opened 2020-01-18
What about putting all app options into a seperate script?


szaimen:
What about putting the whole section followed by this line https://github.com/nextcloud/vm/blob/e051e3240e7a3026e7e39e05bae148b93ef6e5f1/nextcloud-startup-script.sh#L432 into an app script or main script that can get called by the startup script? (it would end at line 515)

Then you would have the possibility to execute only this script afterwards (when trying to install a new app) and see all available apps in one place. This would be better imo then having to look elsewhere for available apps...

Btw: I would also like to put https://github.com/nextcloud/vm/blob/e051e3240e7a3026e7e39e05bae148b93ef6e5f1/nextcloud-startup-script.sh#L375 into a new script, that could get called by the startup script...

enoch85:
I vote no, please see comment in #1038 

Feel free to close.

szaimen:
The idea was especially having a main menu for installing additional apps later on, but if you don't like it, I am okay with that.

enoch85:
Let me think about it.

szaimen:
I would imagine something like the nextcloud update script:
You call an apps.sh static script, that is inside $SCRIPTS directory, that automatically fetches the main script additional_apps.sh from the repository. The additional_apps.sh script is then the main menu to install additional apps. Then you would not longer need the $SCRIPTS/apps dir.
What do you think?

enoch85:
Hmm. Problem with that is that you need to keep everything backwards compatible. It mostly is already, but yeah...

What you mean is something like the `update.sh` script where you just run `update.sh` to get the latest `nextcloud_update.sh`, but like a built in function instead. So something like:

`sudo bash install_app.sh collabora` or
`sudo bash install_app.sh talk`or
`sudo bash install_app.sh fts`

Right?

Make a PR and we'll discuss, I actually think it's a pretty good idea. :)

szaimen:
Not completely, but I will make a PR and show you what I mean with that.

PR #1051, opened 2020-01-23
Additional apps.sh

szaimen:
fix https://github.com/nextcloud/vm/issues/1041


szaimen:
@enoch85 this is what I meant with https://github.com/nextcloud/vm/issues/1041
Please tell me what you think about it and what you would change.

nextcloud-startup-script.sh
```diff
@@ -625,12 +545,6 @@ mesg n
 
 ROOTNEWPROFILE
 
-# Download all app scripts
-print_text_in_color "$ICyan" "Downloading all the latest app scripts to $SCRIPTS/apps..."
-mkdir -p $SCRIPTS/apps
-cd $SCRIPTS/apps
-check_command curl -s https://codeload.github.com/nextcloud/vm/tar.gz/master | tar -xz --strip=2 vm-master/apps
```
szaimen: If we don't delete these lines, the apps.sh would take the apps, that are already installed and we then shouldn't have to make the scripts backwards compatible.
This could only be a problem if we add a new script...

szaimen: But I would really like this in the current form, since it always would load the newest scripts and you have a main menu for installing apps.

additional_apps.sh
```diff
@@ -0,0 +1,103 @@
...
+        ;;
+	
+        *)
+        ;;
+    esac
+done 11< results
+rm -f results
+clear
+
+exit
```
szaimen: We could of course just put this whole part into the static script apps.sh and then had no issue with backwards compatibility.

szaimen: But someone that runs the script would then not profit from new apps.

enoch85:
Aah, now I get it, you want to put it in a seperate script beacuse the startup script is static and only released when a new release is made. This way everyone would benefit from the latest apps, always. Right?

szaimen:
> Aah, now I get it, you want to put it in a seperate script beacuse the startup script is static and only released when a new release is made. This way everyone would benefit from the latest apps, always. Right?

Yes 👍

enoch85:
I was thinking that we could make a script which `curl`s the script from this repo, then install it and delete it again once the installation was OK.

I just added you to docs.hanssonit.se.

Please document this properly, also make changes for existing PRs we merged by you. Email me if you need any permissions.

Thanks!

szaimen:
> I was thinking that we could make a script which `curl`s the script from this repo, then install it and delete it again once the installation was OK.

I don't understand completely what you mean. If find it quite useful to have the apps.sh in the script-directory and so that I can also call it more then one time and install additional apps. Or is that not what you mean?

szaimen:
> I just added you to docs.hanssonit.se.
> 
> Please document this properly, also make changes for existing PRs we merged by you. Email me if you need any permissions.
> 
> Thanks!

I will do that but maybe you could advice me which PRs needs documentation and where exactly in the docs I should add which PR?
Because I have honestly never written any documentation...

enoch85:
> I will do that but maybe you could advice me which PRs needs documentation and where exactly in the docs I should add which PR?

Imagine for yourself, what information would be useful if you knew nothing about Linux, Ubuntu, sudo, Nextcloud, or running a server?

szaimen:
> Imagine for yourself, what information would be useful if you knew nothing about Linux, Ubuntu, sudo, Nextcloud, or running a server?

Thats very unspecific but I will think about what makes sense to add/change based on my PRs. 👍

szaimen:
> > I was thinking that we could make a script which `curl`s the script from this repo, then install it and delete it again once the installation was OK.
> 
> I don't understand completely what you mean. If find it quite useful to have the apps.sh in the script-directory and so that I can also call it more then one time and install additional apps. Or did you mean something else?

Maybe another possibility would be adding a command here to install asditional apps based on a command
https://github.com/nextcloud/vm/blob/cb9d8bc36a4b53c1d01cd0702c15a61c7dc7f2b4/static/change-root-profile.sh#L48

enoch85:
> > > I was thinking that we could make a script which `curl`s the script from this repo, then install it and delete it again once the installation was OK.
> > 
> > 
> > I don't understand completely what you mean. If find it quite useful to have the apps.sh in the script-directory and so that I can also call it more then one time and install additional apps. Or did you mean something else?
> 
> Maybe another possibility would be adding a command here to install asditional apps based on a command
> 
> https://github.com/nextcloud/vm/blob/cb9d8bc36a4b53c1d01cd0702c15a61c7dc7f2b4/static/change-root-profile.sh#L44

That's what I was talking about, that would remove the need for the `/apps` folder since it always would get the latest scripts.

szaimen:
> That's what I was talking about, that would remove the need for the `/apps` folder since it always would get the latest scripts.

Okay, great! Because this was also my intention 👍

enoch85:
OK, let's merge this.

enoch85:
This was a bad idea now when I think of it some more.

Since there are no checks in place if the app was already installed, the script will either fail hard or overwrite current settings which would make some configs invalid.

The only scripts that can be run several times as I can remember now is Colalbora and OnlyOffice, and CookieLifetime.

This means one of two things:
1. Revert this PR
2. Add checks in each script, or make all the app script possible to be run twice.

szaimen:
I will take the second option 👍

szaimen:
But since it is always fetched from the server we can still correct it 👍

enoch85:
> I will take the second option +1

OK, please check all the scripts, run them several times and change so that they can be run several times no matter if it's configured before or not. If the app in question can't be reinstalled due to critical config, stop the script instead and put a `msg_box` that it's not possible and explain why. Bitwarden is such an example, due to saved passwords for example.

Remember, stability. :)

szaimen:
I will first start with a check if a script was already executed. Maybe I will make them executable more then one time later on.

szaimen:
I would say that the following are already reinstallable:

- collabora
- fulltextsearch
- netdata
- onlyoffice
- previewgenerator
- smbmount
- spreedme
- bitwarden? cannot find the check for if it is already installed...

Am I right?

Issue #1052, opened 2020-01-23
one last addition for the 18.0.0 release?


szaimen:
I would like to add the following option to the startup script to let the user configure, whether he wants to get all shares in his root directory or in a folder.
`occ_command config:system:set share_folder --value="/Shared"`

I would then add this as an option to here: https://github.com/nextcloud/vm/blob/cb9d8bc36a4b53c1d01cd0702c15a61c7dc7f2b4/nextcloud-startup-script.sh#L375

I think this makes a lot of sense, since your root directory gets very fast really cluttered, if you get a lot of shares. (imo this should be the default setting for every nextcloud instance)

enoch85:
Add it to the smbmount script instead, right?

szaimen:
No, this works for all instances. Not only if smbshares is configured.
(All shares that you get will appear in the Shared folder and not in your Nextcloud root directory)

enoch85:
Please make a PR and we'll discuss.

szaimen:
> Please make a PR and we'll discuss.

done 👍

enoch85:
Weäll continue the discussion in the PR instead.

PR #1054, opened 2020-01-23
add the config option "share-folder"

szaimen:
fix https://github.com/nextcloud/vm/issues/1052

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -410,7 +410,7 @@ do
             if [[ "yes" == $(ask_yes_or_no "Do you want to enable this option?") ]]
             then
                 occ_command config:system:set share_folder --value="/Shared"
-                msg_box "All new Nextcloud-shares from other users will be stored in the 'Shared' folder from now on."
+                msg_box "All new Nextcloud-shares from other users will appear in the 'Shared' folder from now on."
             fi
         ;;
 	
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -406,10 +406,10 @@ do
 		
         "Share-folder")
             clear
-            msg_box "This option will make all Nextcloud shares from other users appear in a folder named 'Shared' in the root of Nextcloud.\n\nIf you don't enable this option, all shares will appear in the Nextcloud GUI root folder, which is the default behaviour."
+            msg_box "This option will make all Nextcloud shares from other users appear in a folder named 'Shared' in the Nextcloud GUI.\n\nIf you don't enable this option, all shares will appear directly in the Nextcloud GUI root folder, which is the default behaviour."
             if [[ "yes" == $(ask_yes_or_no "Do you want to enable this option?") ]]
             then
-                #occ_command config:system:set share_folder --value="/Shared"
+                occ_command config:system:set share_folder --value="/Shared"
                 bash "$SECURE"
                 msg_box "All new Nextcloud shares from other users will appear in the 'Shared' folder from now on."
             fi
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -407,7 +407,7 @@ do
         "Share-Folder")
             clear
             occ_command config:system:set share_folder --value="/Shared"
-            msg_box "All new Nextcloud-shares from other users will be stored in the 'Shared' folder by now."
+            msg_box "All new Nextcloud-shares from other users will be stored in the 'Shared' folder from now on."
         ;;
 	
         *)
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -378,7 +378,7 @@ whiptail --title "Extra configurations" --checklist --separate-output "Choose wh
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
 "Automatic updates" "(Automatically update your server every week on Sundays)" OFF \
 "CookieLifetime" "(Configure forced logout timeout for users using the web GUI)" OFF \
-"Share-folder" "(If configured, all Nextcloud-shares from other users will be stored in the 'Shared' folder)" OFF 2>results
+"Share-folder" "(If configured, all Nextcloud shares from other users will be stored in a folder named 'Shared')" OFF 2>results
 
 while read -r -u 9 choice
 do
@@ -406,11 +406,12 @@ do
 		
         "Share-folder")
             clear
-            msg_box "This option will let all Nextcloud-shares, that a Nextcloud-user gets from other Nextcloud-users appear in a folder named 'Shared', that is inside the Nextcloud users root folder.\nIf you don't enable this option, all shares will appear in the Nextcloud users root folder.\nThis folder will be created automatically when the specific user gets his first share.\nPlease note: your root folder must be writeable so that this folder can be automatically created, or you have to create it manually. Otherwise you will not be able to get shares from other Nextcloud-users."
+            msg_box "This option will make all Nextcloud shares from other users appear in a folder named 'Shared' in the root of Nextcloud.\n\nIf you don't enable this option, all shares will appear in the Nextcloud GUI root folder, which is the default behaviour."
             if [[ "yes" == $(ask_yes_or_no "Do you want to enable this option?") ]]
             then
-                occ_command config:system:set share_folder --value="/Shared"
-                msg_box "All new Nextcloud-shares from other users will appear in the 'Shared' folder from now on."
+                #occ_command config:system:set share_folder --value="/Shared"
+                bash "$SECURE"
+                msg_box "All new Nextcloud shares from other users will appear in the 'Shared' folder from now on."
             fi
         ;;
 	
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -406,8 +406,12 @@ do
 		
         "Share-Folder")
             clear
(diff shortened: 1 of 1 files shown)


nextcloud-startup-script.sh
```diff
@@ -402,7 +403,13 @@ do
             clear
             run_static_script cookielifetime
         ;;
-
+		
+        "Share-Folder")
+            clear
+            occ_command config:system:set share_folder --value="/Shared"
+            msg_box "All new Nextcloud-shares from other users will be stored in the 'Shared' folder by now."
```
enoch85: ... users will be stored in the 'Shared' folder from now on

nextcloud-startup-script.sh
```diff
@@ -377,7 +377,8 @@ whiptail --title "Extra configurations" --checklist --separate-output "Choose wh
 "Security" "(Add extra security based on this http://goo.gl/gEJHi7)" OFF \
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
 "Automatic updates" "(Automatically update your server every week on Sundays)" OFF \
-"CookieLifetime" "(Configure forced logout timeout for users using the web GUI)" OFF 2>results
+"CookieLifetime" "(Configure forced logout timeout for users using the web GUI)" OFF \
+"Share-folder" "(If configured, all Nextcloud-shares from other users will be stored in the 'Shared' folder)" OFF 2>results
 
 while read -r -u 9 choice
```
enoch85: no need to change to 10 here?

szaimen: I guess no: but actually I don't know what that number does.

nextcloud-startup-script.sh
szaimen: done 👍

nextcloud-startup-script.sh
enoch85: Me neither to be honest. :)

enoch85: I saw you changed it in other places though, hence the question.

nextcloud-startup-script.sh
```diff
@@ -402,7 +403,13 @@ do
             clear
             run_static_script cookielifetime
         ;;
-
+		
+        "Share-Folder")
+            clear
+            occ_command config:system:set share_folder --value="/Shared"
```
enoch85: Don't we need to create the `/Shared` folder, or is it created automatically?

enoch85:
Please be more clear on which shares are saved where in a second `msg_box` then ask if the user want's to change it with `ask_yes_no`.

nextcloud-startup-script.sh
szaimen: It is created automatically. 👍

nextcloud-startup-script.sh
szaimen: I changed it because I thought it was the max number of options to choose from ;)

szaimen:
> Please be more clear on which shares are saved where in a second `msg_box` then ask if the user want's to change it with `ask_yes_no`.

this should be done 👍

nextcloud-startup-script.sh
szaimen: Maybe we should test if removing the 9 changes anything :D

enoch85:
OK, now I understand it better.

> Otherwise you will not be able to get shares from other Nextcloud-users.    

This makes me worried. Are you talking about `www-data` permissions? n that case, skip that part and run `setup_secure_permissions.sh` instead as a proactive action.

I can do that now.

szaimen:
> OK, now I understand it better.
> 
> > Otherwise you will not be able to get shares from other Nextcloud-users.
> 
> This makes me worried. Are you talking about `www-data` permissions? n that case, skip that part and run `setup_secure_permissions.sh` instead as a proactive action.
> 

No, I am not talking about www-data permissions.
Your Nextcloud root folder is only not writeable, if you use the external storage app to change the root folder of the nextcloud-users and make it not-writeable.

enoch85:
What do you think about the latest commit?

szaimen:
> What do you think about the latest commit?

have you read my comment above?
I hope this explains everything. 
(I would uncomment `#occ_command config:system:set share_folder --value="/Shared"` and remove `bash "$SECURE"` again.)

szaimen:
@enoch85 now it looks already better but we don't need `bash "$SECURE"` here.

enoch85:
> Your Nextcloud root folder is only not writeable

I don't understand that.

If this has something to do with external storage's I absolutely don't think this is suited for the start-up script. It's better suited as a part of smbmount as I mentoned before.

You say "rights" which permissions are needed? Nextcloud is on `www-data`, nothing else.

> I could remove that part, if this is to much information and unnecessary.

I'm confused, imagine a regular user. :)

szaimen:
> > Your Nextcloud root folder is only not writeable
> 
> I don't understand that.
> 
> If this has something to do with external storage's I absolutely don't think this is suited for the start-up script. It's better suited as a part of smbmount as I mentoned before.
> 
> You say "rights" which permissions are needed? Nextcloud is on `www-data`, nothing else.
> 
> > I could remove that part, if this is to much information and unnecessary.
> 
> I'm confused, imagine a regular user. :)

I will try to explain it differently:
- This PR will always work, if the external App isn't enabled!
- If the external App is enabled, this PR will always work, unless you have not created an external storage folder, that is called "/" and have set it to read only.
=> So this is a very custom setting and shouldn't be a problem for this PR.


Are you okay with that explanation?
I currently don't know how to explain it that you understand what I mean.




PS: if you create an external storage folder with the name "/" this folder becomes your new Nextcloud GUI root folder, which in this case wouldn't be writeable. So the "Shared" folder could not be automatically created.
and BTW: external storage is not only a smbshare but a path on the local filessystem or something else.

szaimen:
Sry there was a wording issue in my writing above. I corrected it. Please read it again, if you have already read this.

enoch85:
> If the external App is enabled, this PR will always work, unless you have not created an external storage folder, that is called "/" and have set it to read only.

OK, then we should add a check for that. Else it will fail. 

Actually, I still don't understand the need for this. If the user wants to to this, read the documentation. This seems like an edge case.

szaimen:
> > If the external App is enabled, this PR will always work, unless you have not created an external storage folder, that is called "/" and have set it to read only.
> 
> OK, then we should add a check for that. Else it will fail.
> 
> Actually, I still don't understand the need for this. If the user wants to to this, read the documentation. This seems like an edge case.

I cannot add a check for this since the external storage app is enabled and mounted afterwards.
This is no edge case imo, but quite useful if you want to organize your root folder: it is a default setting in many Nextcloud instances that I know and tried.

Sry that I came up with this edge  error case with the external storage app, I don't think that this should prevent the PR from beeing merged.

But I would definitely remove the `bash "$SCRIPTS"/secure.sh`

enoch85:
So what you are saying is that this will always work *until* they decide to change it themselves in the External Storages app? Aren't their change in GUI reflecting in `config.php`?

Also, this is not about *all* shares. just the shares in the External Storages app?

Also, what is root in this case, the root folder in Nextcloud GUI, or in CLI?

I don't feel confident in merging something that I can't even explain myself. (I don't use External Storages).

enoch85:
Let's do this in 18.0.1 instead. 

I will make a small release now, and the rest of the versions this weekend.

szaimen:
hm :/ but okay
I will prepare a detailed explanation with pictures for that

szaimen:
This is my detailed explanation:
1. for reference: here is the nextcloud storage for the admin user:
image

if the admin user gets a share from another nextcloud user, it appears inside his root folder:
image

3. now the case that you have setuped the Shared folder before:
this is the users root folder again:
image

if this user now gets a share by another nextcloud user, it looks like this:
image
(so the "Shared" folder got created automatically)

if you open this folder, you find the shared file:
image

3. so that the case, that you have setuped an external storage empty root folder:
image

this is your new root folder than (and empty because no file is in it)
image

if you now get a share from another user, the shared folder gets created automatically again:
image

So everything works.

But if you make it read only:
image

and this is now the admins root folder (no shared folder in it)
image

if you now create a share, 
nothing appears hear, because the path is not writeable:
image

But if the Shared folder is already created there or added as additional external storage, the shares are visible again:
image

@enoch85 I hope this helps you in understanding the problem here, but I hope you get that this is really an edge case issue and shouldn't be a problem here.
Please remember: The share_folder has nothing to do with the external storage app!
It is a setting that also works with the default calibration and would never fail!
The problem is as I said before, that you cannot see your shared, files, if the root directory is not writeble and the Shared folder is non-existant.

enoch85:
Now that's a good explanation! Put that in the docs! and then we can link it in the `msg_box`.

I'm fine with merging this, after documentation is done, which in any case will be in 18.0.1.

nextcloud-startup-script.sh
```diff
@@ -402,7 +403,18 @@ choice=$(< results)
             clear
             run_static_script cookielifetime
         ;;
-
+		
+        "Share-folder")
```
enoch85: remove the " " here please.

enoch85: No, don't do it. It's correct.

szaimen:
outrun by https://github.com/nextcloud/vm/pull/1070?

szaimen:
I will close this since its outrun by https://github.com/nextcloud/vm/pull/1070.

Jegbvgiyg (commented):
nextcloud-startup-script.sh[]()

Issue #1065, opened 2020-01-25
Make the option to exit smbmont.sh more obvius


enoch85:
@szaimen Please add an option which is "Exit this script" instead of having to TAB to cancel to exit. Right now that's not obvious at the moment. 

Please also fix the last `whiptail` so that it follows the standard of the rest of the scripts.

Thanks!

szaimen:
I will do that if you are okay, with the new way of doing whiptails: https://github.com/nextcloud/vm/issues/1058#issuecomment-578426862

szaimen:
I would then also fix it in cookielifetime

szaimen:
all should be fixed with https://github.com/nextcloud/vm/pull/1069

Issue #1058, opened 2020-01-24
Fix choice in while loops


enoch85:
https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L165-L166

Fix in all places!

enoch85:
apps/previewgenerator.sh:    done 11< results
apps/previewgenerator.sh:    done 8< results
nextcloud_install_production.sh:choice=$(< results)
nextcloud_install_production.sh:choice=$(< results)
nextcloud_install_production.sh:done 11< results
nextcloud-startup-script.sh:done 9< results
additional_apps.sh:done 11< results

enoch85:
Done in https://github.com/nextcloud/vm/pull/1059

enoch85:
Just for reference:

`whiptails` with `--checklist` must be:

```
while read -r choice <-------
do <-------
    case "$choice" in <-------
        ..................
        ;;
        *)
        ;;
    esac
done < results <-------
rm -f results
```

`whiptails` with `--radiolist` must be:
```
choice=$(< results)  <-------
    case "$choice" in  <-------
    .....................
        ;;
        *)
        ;;
    esac
rm -f results
```

enoch85:
FYI @szaimen

szaimen:
But I guess you could also solve at least the radiolist like in the cookielifetime script
(With if and elif)

enoch85:
Actually, I think this was strange: https://github.com/nextcloud/vm/blob/master/apps/smbmount.sh#L266-L295 and I was about to change it to this https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L160-L179

But I thought maybe you had a plan. Everything is solved now anyway, testing as we speak.

szaimen:
> Actually, I think this was strange: https://github.com/nextcloud/vm/blob/master/apps/smbmount.sh#L266-L295 and I was about to change it to this https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L160-L179
> 
> But I thought maybe you had a plan. Everything is solved now anyway, testing as we speak.

Both methods should work 👍

enoch85:
> Both methods should work +1

I think we should use the same code everywhere. Please change to the standard used in all other places.

Thanks!

enoch85:
`whiptails` are still broken........

Too tired now, have to sleep.

If you have time, please test all of them and try to fix it. 

Thanks!

enoch85:
[...]

PR #1069, opened 2020-01-25
Update smb-mount and cookielifetime with new option and whiptail

szaimen:
fix https://github.com/nextcloud/vm/issues/1065


szaimen:
@enoch85 this should be ready for review

enoch85:
radiolists are done differently in other places: https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L160-L180

How is this better?

enoch85:
Tested CookieLiftime and it works! 

Thanks @szaimen 

Though, I'm still wondering about the change from standard?

szaimen:
> radiolists are done differently in other places: https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L160-L180
> 
> How is this better?

I just like it more: doing it everywhere the same: you have continuity between msg_box -> inputbox -> radiolist -> checklist.

szaimen:
> Tested CookieLiftime and it works!
> 
> Thanks @szaimen
> 
> Though, I'm still wondering about the change from standard?

I hope my comment above explains it.
otherwise here is a more complete view:
https://github.com/nextcloud/vm/issues/1058#issuecomment-578426862

enoch85:
OK, let's merge this, if it works nice, then we'll change in other places as well.

Issue #1063, opened 2020-01-25
Rich workspaces


szaimen:
I have tried the Nextcloud 18 release a bit and I really don't like the rich workspaces, because I think they are really confusing for most people and can spam your file structure with readme files.
So I would like to provde an option during the startup script to disable them, based on the command:
`occ_command config:app:set text workspace_available --value=0`
this will only work starting at 18.0.1: https://github.com/nextcloud/text/issues/586#issuecomment-578387536

szaimen:
Since this would be already the 3rd additional configuration option, maybe we should think about extracting this into an additional whiptail for nextcloud configuration? (now we had 6 options in this one whiptail, 3 are server based things, and the 3 added by me are nextcloud configurations. We could do it like the additional_apps script is now solved?

szaimen:
I will do a PR later on so that we can discuss it.

enoch85:
I'm with you 100%, I feel exactly the same way.

Question is, should we make it default, or an option?

szaimen:
I would disable it by default but this is the Nextcloud VM and as you said before: it should be the Nextcloud that the developer team says that it should be so I would just add the option to disable it.

enoch85:
OK, let's put it in "Extra configuration" as an option, like "Disable top notes in GUI".

szaimen:
> Since this would be already the 3rd additional configuration option, maybe we should think about extracting this into an additional whiptail for nextcloud configuration? (now we had 6 options in this one whiptail, 3 are server based things, and the 3 added by me are nextcloud configurations. We could do it like the additional_apps script is now solved?

What do you think about that idea?

enoch85:
> What do you think about that idea?

Nope, add it in existing whiptail. I don't want to clutter the setup with a lot of different whiptails.

Apps is one thing, Extra configuration is everything else.

szaimen:
I have just created a PR for that :/

szaimen:
I think it makes sense because this shouldn't be only bound to the initial setup, like the server configuration, imo.

szaimen:
Another example why putting all nextclojd configuration options into the startup script is probably not the best way:
If we decide to moce the cookielifetime script later on to another place or rename it, the startup script would fail. But if we just have the curl script linked here, we can change whatever we want later on without having to edit the startup script...

Issue #11, opened 2016-06-22
Check downloaded files with GPG before using them.


ypid:
Refer to:
- https://github.com/owncloud/vm/issues/45
- https://github.com/jchaney/owncloud/pull/12

enoch85:
I saw you made a PR for this before, could you please add one here as well?

enoch85:
Btw, this is not done with packages.

ypid:
> I saw you made a PR for this before, could you please add one here as well?

I think I did not submit a PR for this yet. I will be happy to review your PR :wink:

enoch85:
@ypid Done: https://github.com/nextcloud/vm/commit/c38e0c3c80886a9c29d60919908e5cefe13ba157

ypid:
@enoch85 Thanks very much for your work. That means a lot to me because this
will ensure that tempering with the build process will become more difficult.

However, I am afraid that you completely missed the point of checking the GPG
signature in the first place.

I refer you to this comment: https://github.com/jchaney/owncloud/pull/12#issuecomment-99544307
to which I already referred to in my opening comment in case it is unclear what the use case of GPG is here.

As it seems that the public key of NextCloud is currently not available on public key servers, you can do this little trick:

```Shell
gpg --homedir /tmp/gpg --import nextcloud.asc
gpg --homedir /tmp/gpg --export 28806A878AE423A28372792ED75899B9A724937A | gpg --import -
```

to enforce the correct key fingerprint of the public key.

enoch85:
I used the docs, and according to them my fix is valid: https://docs.nextcloud.org/server/9/admin_manual/installation/source_installation.html?highlight=asc#example-installation-on-ubuntu-14-04-lts-server

ypid:
Your fix is valid but as I said, does only get you very little compared to if you enforce the fingerprint in your build scripts, that will give you (and all users) a whole lot more. In my opinion both the [NextCloud documentation](https://docs.nextcloud.org/server/9/admin_manual/installation/source_installation.html#ubuntu-installation-label) and the [ownCloud documentation](https://doc.owncloud.org/server/9.1/admin_manual/installation/source_installation.html#ubuntu-installation-label) about checking GPG signatures are a bit unmindful and need to be fixed. But note that setting up NextCloud or ownCloud one time as a typical admin vs. writing scripts to build a template VM has a different threat model: https://www.qubes-os.org/news/2016/05/30/build-security/

[...]

PR #1070, opened 2020-01-25
introduce nextcloud_configuration

szaimen:
fix https://github.com/nextcloud/vm/issues/1063
successor of https://github.com/nextcloud/vm/pull/1054?


enoch85:
If anything, please don't change the name of the startup script. It's been the same for years, and is posted in forums, blogs and everywhere.

Changing it would make links and dodumentation invalid.

Thanks!

szaimen:
> If anything, please don't change the name of the startup script. It's been the same for years, and is posted in forums, blogs and everywhere.
> 
> Changing it would make links and documentation invalid.
> 
> Thanks!

Sry, I didn't know that. Have reverted it. 👍

static/change-root-profile.sh
```diff
@@ -44,7 +44,8 @@ ROOT-PROFILE
 # Add Aliases
 {
 echo "alias nextcloud_occ='sudo -u www-data php $NCPATH/occ'"
-echo "alias run_update_nextcloud='bash $SCRIPTS/update.sh'"
+echo "alias nextcloud_update='bash $SCRIPTS/update.sh'"
```
enoch85: the reason why it's called *run*_nextcloud_update is because if you TAB the alias you might make a mistake and run the update instead of nextcloud_occ.

szaimen: maybe would run_nextcloud_update be okay?

enoch85: Hmm, it's been out there for a while now. I'm voting no. People may have it in their current crontabs and such. Already got emails about users that do.

szaimen: It is reverted.

szaimen:
@enoch85 with my comment https://github.com/nextcloud/vm/issues/11#issuecomment-578540328 in your mind, do you think, that this PR is feasible?

szaimen:
@enoch85 Do you have no thoughts on my comment? https://github.com/nextcloud/vm/issues/11#issuecomment-578540328

szaimen:
@enoch85 I thought about this a bit more and I would definitely like to get this into 18.0.1 
Since you seem to don't like this approach, I would change it to be static inside the startup script for this release. Is this an approach you would support? 

We could then think about making it modular later on. (possibly when https://github.com/nextcloud/vm/issues is solved.)

enoch85:
Do you consider this PR to be done and ready for release?

enoch85:
If you are 100% sure on how to solve conflicts, please do that. If you're 99% sure, make a new PR with the same content, and close this one.

Thanks!

static/configuration.sh
```diff
@@ -0,0 +1,28 @@
...
+# Must be root
+root_check
+
+mkdir -p "$SCRIPTS"
+print_text_in_color "$ICyan" "Running the nextcloud configuration script..."
+
+if network_ok
+then
+    # Delete, download, run
+    run_static_script nextcloud_configuration
```
enoch85: `run_static_script` first deletes all copies of the script, then it downloads it, then it runs it, and then deletes it again. 

The whole point here is that everything should be in the setup script already. Maybe I was too quick merging your other PR, not when I think of it, it actually made things ore unstable.

szaimen: yes, I think that the configuration script isn't that important. Since you get additional options, that aren't really neccessary to run nextcloud but that are welcome additions.

szaimen:
> If you are 100% sure on how to solve conflicts, please do that. If you're 99% sure, make a new PR with the same content, and close this one.
> 
> Thanks!

I've created https://github.com/nextcloud/vm/pull/1077. So this can be closed.

Issue #1074, opened 2020-01-28
First bug report regarding SMB script


enoch85:
Message Body:
Hi Daniel, I have successfully installed Nextcloud on my esxi server and can connect remotely using my domain https://nextcloud/mydomain.com.

I am trying to set up an smb share using your smbmount script but I am getting the error advising unable to set up a mount.

I think my error is the UNC //Server/Share. The hostname of my server is nextcloud so I am inputting //nextcloud/Files but no luck.

Is it something obvious I am missing as I can set up an smb share in Ubuntu server the old fashioned way but your script needs a specific input I think.

Chris

@szaimen

szaimen:
@enoch85 does Chris has an github account or an email address so that we can communicate? This seems to be reported over email, so please send me his email address if he allows that.

enoch85:
I asked him to sign up here. If he doesn't I'll send you his email in a PM.

szaimen:
Thanks!

szaimen:
I have already an assumption: he should try the ip-address from his smb-server once instead of the smb-server name "nextcloud"

So e.g. `//ip-address/share`

enoch85:
> If that is the case here, I think I will have to add a check for this.

Add a check for it since it might be an issue anyway. Be progressive. :)

CW450:
Hi all I have signed up for GitHub so can contribute
Chris

CW450:
Daniel,

I tried again after the update and using the IP address it worked (but didn’t work using the UNC name of my pc) and set up two SMB shares which I can see in my Win 10 pc and Nextcloud on both pc and mobile.

Is it by design that I can’t add files via the Windows share and they appear in the Nextcloud SMB? As I can add to my Nextcloud and they appear in my windows share immediately.  They are called SMB1 and SMB2 and not the windows share names, is that by design also?

Chris

szaimen:
> I tried again after the update and using the IP address it worked (but didn’t work using the UNC name of my pc)

Have you tried using the UNC after the update? If yes, it is most likely a network problem.

> Is it by design that I can’t add files via the Windows share and they appear in the Nextcloud SMB?

No, this should work. At least for me it does; the new files should get loaded when you access the directory. 
Can you check if the follwing option is activated?
![image](https://user-images.githubusercontent.com/42591237/73486920-505b1500-43a6-11ea-8754-fe3a4d7c7d19.png)

> They are called SMB1 and SMB2 and not the windows share names, is that by design also?

[...]

PR #1076, opened 2020-01-28
Possible fix for first SMB-bug

szaimen:
Fix https://github.com/nextcloud/vm/issues/1074?

### DIFF
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -46,7 +46,7 @@ fi
 # Enter SMB-server and Share-name
 while true
 do
-    SERVER_SHARE_NAME=$(whiptail --inputbox "Please enter the server and Share-name like this:\n//Server/Share" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 this should be ready. Just if you wonder: I fixed the following travis issue with my solution:
```
In apps/smbmount.sh line 24:
if [ -z "$(grep "^hosts:" /etc/nsswitch.conf | grep wins)" ]
     ^-- SC2143: Use ! grep -q instead of comparing output with [ -z .. ].
```

enoch85:
It's your script. :) Merge. :)

apps/smbmount.sh
```diff
@@ -19,6 +19,13 @@ root_check
 # Install cifs-utils
 install_if_not cifs-utils
 
+# Make sure, that name resolution works
+install_if_not winbind
+if [ "$(grep "^hosts:" /etc/nsswitch.conf | grep wins)" == "" ]
```
enoch85: Isn't it better to use `if [ -z .... ]` here?

enoch85: So that would be something like `if [ -z "$(grep "^hosts:" /etc/nsswitch.conf | grep wins)" ]`

PR #1077, opened 2020-02-03
new nextcloud_configuration.sh

### LINKED ISSUE #1063: Rich workspaces

szaimen:
successor to https://github.com/nextcloud/vm/pull/1070
fix #1063
and #1052

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -241,6 +241,7 @@ download_static_script change_db_pass
 download_static_script nextcloud
 download_static_script update-config
 download_static_script apps
+download_static_script configuration
 download_le_script activate-ssl
 if home_sme_server
 then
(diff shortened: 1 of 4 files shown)


szaimen:
I think this is ready.

nextcloud-startup-script.sh
```diff
@@ -394,14 +394,13 @@ case *"$choice" in
...
-    *"CookieLifetime"*)
-        clear
-        run_static_script cookielifetime
-    ;;&
     *)
     ;;
 esac
 
+# Nextcloud configuration
+bash $SCRIPTS/configuration.sh
```
szaimen: Maybe we could move that inside the startup script if it is not that important?
Could maybe be after lets encrypt and directly before additional_apps?

szaimen: I moved it to where I think it fits best.

enoch85:
OK, I'll merge this. Let's see if it produces any bugs.

Issue #1048, opened 2020-01-20
Onlyoffice out of the box?


szaimen:
Have you seen this on the Nextcloud 18 publication site?
![image](https://user-images.githubusercontent.com/42591237/72716160-0ede9480-3b72-11ea-8201-188cb3af7bf4.png)
So does that mean, that the onlyoffice script is no longer needed?
I have to try it myself later on.
Here is the link: https://nextcloud.com/blog/the-new-standard-in-on-premises-team-collaboration-nextcloud-hub/

szaimen:
I found this: https://apps.nextcloud.com/apps/documentserver_community

enoch85:
I haven't had the time to investigate yet, a lot of stuff happening in my day-to-day life right now.

enoch85:
Had a quick look in my personal NC 18, didn't find any "magic button" for installing OnlyOffice.. :)

szaimen:
The trick is to install this app:
https://apps.nextcloud.com/apps/documentserver_community

enoch85:
Wooow! If it's that easy I'm amazed! Please try and come back with the results.

szaimen:
> Wooow! If it's that easy I'm amazed! Please try and come back with the results.

Yes, I am also very satisfied by that!
I have already tested it on an upgraded Test VM and it works. 
But I am not sure, if that is as reliably as the scripted install since it is an really new app.
We could of course just change the script to `install_and_enable documentserver_community`
What do you think?

szaimen:
I have tried it but I think you have to install onlyoffice first:
so this should do the trick:
```
install_and_enable_app onlyoffice
install_and_enable_app documentserver_community
```
But it is not reliably in the moment, since the documentserver is that big:
https://github.com/nextcloud/server/pull/18976

szaimen:
So I would propose testing this when the bugfix is released (probably with 18.0.1) and thinking about substituting the code in the onlyoffic.sh file then.

oezh:
My onlyoffice stoped working after update from 17 to 18. https://oo.server.com/healthcheck returns True, Saving on Nextcloud Connector config returns Ok with no errors (for both, Docker from onlyoffice.sh and Community Document Server on their respective address). But when going to files and click an office file it gets downloaded instead of opened in onlyoffice. Also, the "create new" (+) file button doesn't show any office mime type file. Seems like missing config or something. Am I missing something?

oezh:
[...]

PR #1081, opened 2020-02-13
add new onlyoffice installation to the script

szaimen:
fix https://github.com/nextcloud/vm/issues/1048

### DIFF
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -23,18 +23,9 @@ root_check
 # Nextcloud 13 is required.
 lowest_compatible_nc 13
 
-## Onlyoffice for NC 18.0.1 onwwards
-# Check for the NC-version
-CURRENTNCVERSION="$(occ_command config:system:get version | sed 's/^\(.*\..*\..*\)\..*/\1/')"
-CURRENTNCVERSION="${CURRENTNCVERSION//./}"
-if [ "$CURRENTNCVERSION" -lt "9999" ]
-then
-    CURRENTNCVERSION="$(sed 's/./&0/3' <<<"$CURRENTNCVERSION")"
-fi
-
 # Check if onlyoffice is already installed
 print_text_in_color "$ICyan" "Checking if Onlyoffice is already installed..."
-if [ "$CURRENTNCVERSION" -ge 18001 ] && ! does_this_docker_exist 'onlyoffice/documentserver'
+if version_gt "$CURRENTVERSION" "18.0.0" && ! does_this_docker_exist 'onlyoffice/documentserver'
 then
     install_if_not jq
     if occ_command_no_check app:list --output=json | jq -e '.enabled | .documentserver_community' > /dev/null
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -59,7 +59,7 @@ then
         print_text_in_color "$ICyan" "Installing OnlyOffice..."
     fi
     install_and_enable_app onlyoffice
-    install_and_enable_app documentserver_community
+    install_and_enable_app documentserver_community && spinner_loading
     msg_box "Onlyoffice was successfully installed."
     exit
 fi
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -27,10 +27,14 @@ lowest_compatible_nc 13
 # Check for the NC-version
 CURRENTNCVERSION="$(occ_command config:system:get version | sed 's/^\(.*\..*\..*\)\..*/\1/')"
 CURRENTNCVERSION="${CURRENTNCVERSION//./}"
+if [ "$CURRENTNCVERSION" -lt "9999" ]
+then
+    CURRENTNCVERSION="$(sed 's/./&0/3' <<<"$CURRENTNCVERSION")"
+fi
 
 # Check if onlyoffice is already installed
 print_text_in_color "$ICyan" "Checking if Onlyoffice or Collabora is already installed..."
-if [ "$CURRENTNCVERSION" -ge 1801 ] && ! does_this_docker_exist 'onlyoffice/documentserver' || ! does_this_docker_exist 'collabora/code'
+if [ "$CURRENTNCVERSION" -ge 18001 ] && ! does_this_docker_exist 'onlyoffice/documentserver' || ! does_this_docker_exist 'collabora/code'
 then
     install_if_not jq
     if occ_command_no_check app:list --output=json | jq -e '.enabled | .documentserver_community' > /dev/null
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -23,6 +23,47 @@ root_check
 # Nextcloud 13 is required.
 lowest_compatible_nc 13
 
+## Onlyoffice for NC 18.0.1 onwwards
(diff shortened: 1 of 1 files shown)


szaimen:
Note to self: 
-really check for collabora here?
-uninstall documentser when installing collabora?

apps/onlyoffice.sh
```diff
@@ -23,6 +23,51 @@ root_check
 # Nextcloud 13 is required.
 lowest_compatible_nc 13
 
+## Onlyoffice for NC 18.0.1 onwwards
```
enoch85: This comes waay to early in the script. 

Need to come after all the checks.

Also, have you tested that this really works?

Nice job btw!

apps/onlyoffice.sh
```diff
@@ -23,6 +23,51 @@ root_check
...
+# Check for the NC-version
+CURRENTNCVERSION="$(occ_command config:system:get version | sed 's/^\(.*\..*\..*\)\..*/\1/')"
+CURRENTNCVERSION="${CURRENTNCVERSION//./}"
```
enoch85: I don't understand this?

apps/onlyoffice.sh
```diff
@@ -23,6 +23,51 @@ root_check
...
+then
+    CURRENTNCVERSION="$(sed 's/./&0/3' <<<"$CURRENTNCVERSION")"
+fi
+
+# Check if onlyoffice is already installed
+print_text_in_color "$ICyan" "Checking if Onlyoffice or Collabora is already installed..."
+if [ "$CURRENTNCVERSION" -ge 18001 ] && ! does_this_docker_exist 'onlyoffice/documentserver' || ! does_this_docker_exist 'collabora/code'
+then
+    install_if_not jq
+    if occ_command_no_check app:list --output=json | jq -e '.enabled | .documentserver_community' > /dev/null
```
enoch85: If you implement a new standard somewhere, please check the rest of the scripts as well and implement the same standard everywhere.

apps/onlyoffice.sh
szaimen: this is the "headline" for the section below. (check first if these steps apply and install it the normal way, if not)
And yes, I have tested it and it works.
Thanks!

apps/onlyoffice.sh
szaimen: this gets the current version from the config file, removes everything after the 4th point (since sometimes you get 18.0.1.4) and deletes all points. So that you get eg. 1801 back.

apps/onlyoffice.sh
```diff
@@ -23,6 +23,51 @@ root_check
 # Nextcloud 13 is required.
 lowest_compatible_nc 13
 
+## Onlyoffice for NC 18.0.1 onwwards
+# Check for the NC-version
+CURRENTNCVERSION="$(occ_command config:system:get version | sed 's/^\(.*\..*\..*\)\..*/\1/')"
+CURRENTNCVERSION="${CURRENTNCVERSION//./}"
+if [ "$CURRENTNCVERSION" -lt "9999" ]
+then
+    CURRENTNCVERSION="$(sed 's/./&0/3' <<<"$CURRENTNCVERSION")"
```
szaimen: if you have a norml version like 1801, you then add here a zero to the 3rd character of 1801 so that you get 18001 back.
Because if later on you would have e.g. 18.0.10, the version would be 18010 which would be greater then the number that is checked later on. So I add a needed zero here.

apps/onlyoffice.sh
```diff
@@ -23,6 +23,51 @@ root_check
...
+fi
+
+# Check if onlyoffice is already installed
+print_text_in_color "$ICyan" "Checking if Onlyoffice or Collabora is already installed..."
+if [ "$CURRENTNCVERSION" -ge 18001 ] && ! does_this_docker_exist 'onlyoffice/documentserver' || ! does_this_docker_exist 'collabora/code'
```
szaimen: This then checks reliably, if the version is 18.0.1 or higher.

apps/onlyoffice.sh
szaimen: okay, I think we already have a few places. so I will do that later on.

apps/onlyoffice.sh
enoch85: Have you checked the current scripts? E.g: https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L186 :)

apps/onlyoffice.sh
enoch85: The only thing we care about here really is 18. Not the 18.X.X.X

apps/onlyoffice.sh
enoch85: Checking for only 18 or higher is what we need.

apps/onlyoffice.sh
szaimen: yes, but I need the whole point release here. and not just the major release version...

apps/onlyoffice.sh
szaimen: no, that is not right. we need 18.0.1 since on 18.0.0 there was a timeout issue when installing this.

apps/onlyoffice.sh
szaimen: no, that is not right. as I stated above.

apps/onlyoffice.sh
szaimen: I will do that in another PR, when this two are merged.
https://github.com/nextcloud/vm/issues/1082

apps/onlyoffice.sh
enoch85: OK, you're right, but we already have a function for checking versions: https://github.com/nextcloud/vm/blob/master/lib.sh#L858-L877

szaimen: Okay, but I don't get how this works. Can you show me an example? Also is the current sub-version provided as an integer? Or how does the comparison of sub-versions work with that function?

enoch85: Actually I don't remember completely either, but it's used when comparing current versions. So I think something like this should work.

I'm testing now on my dev. 

```
if version_gt $CURRENTVERSION 18.0.1
then
    echo "continue"
fi
```

Btw, I would like to release 18.0.1 tomrorow, fyi.

enoch85: Result:

```
 CURRENTNCVERSION=18001
+ version_gt 18.0.1 18.0.1
+ local v1 v2 IFS=.
+ read -ra v1
+ read -ra v2
+ printf -v v1 %03d 18 0 1
+ printf -v v2 %03d 18 0 1
+ [[ 018000001 > 018000001 ]]
+ exit
```

enoch85:
I will have a look at this later, maybe tomorrow.

enoch85:
Btw, you could grep for old code and see how it's used:

```
root@xps13:~/GITHUB/nextcloud# grep -r version_gt
lib.sh:version_gt() {
static/updatenotification.sh:if version_gt "$NCVERSION" "$CURRENTVERSION"
nextcloud_update.sh:if version_gt "$NCVERSION" "$CURRENTVERSION"
```

szaimen:
> Btw, I would like to release 18.0.1 tomrorow, fyi

No problem, since all changes will automatically be available when we merge this ;)

PR #1072, opened 2020-01-26
harden appscripts

szaimen:
fix https://github.com/nextcloud/vm/pull/1051#issuecomment-578543583

I updated all apps.

PR #1084, opened 2020-02-15
change to the Hub method of installing OnlyOffice

enoch85:
@szaimen This is how I would do it.

Doesn't seem to work on my installation though.

### DIFF
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -53,14 +53,14 @@ then
         if [ -f "$CERTFILES/$SUBDOMAIN/cert.pem" ]
         then
             yes no | certbot revoke --cert-path "$CERTFILES/$SUBDOMAIN/cert.pem"
-            rm -rf "${CERTFILES/$SUBDOMAIN:?}/"
+            rm -rf "$CERTFILES"/"$SUBDOMAIN"*
         fi
         # Remove Apache2 config
         if [ -f "$SITES_AVAILABLE/$SUBDOMAIN.conf" ]
         then
             a2dissite "$SUBDOMAIN".conf
             restart_webserver
-            rm "$SITES_AVAILABLE/$SUBDOMAIN.conf"
+            rm -f "$SITES_AVAILABLE/$SUBDOMAIN.conf"
         fi
         # Remove app
         occ_command_no_check app:remove onlyoffice
@@ -125,14 +125,14 @@ then
     if [ -f "$CERTFILES/$SUBDOMAIN/cert.pem" ]
     then
         yes no | certbot revoke --cert-path "$CERTFILES/$SUBDOMAIN/cert.pem"
-        rm -rf rm -rf "${CERTFILES/$SUBDOMAIN:?}/"
+        rm -rf "$CERTFILES"/"$SUBDOMAIN"*
     fi
     # Remove Apache2 config
     if [ -f "$SITES_AVAILABLE/$SUBDOMAIN.conf" ]
     then
         a2dissite "$SUBDOMAIN".conf
         restart_webserver
-        rm "$SITES_AVAILABLE/$SUBDOMAIN.conf"
+        rm -f "$SITES_AVAILABLE/$SUBDOMAIN.conf"
     fi
     # Remove app
     occ_command_no_check app:remove richdocuments
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -27,9 +27,32 @@ cpu_check 2 OnlyOffice
 # Get the latest packages
 apt update -q4 & spinner_loading
 
-# Check if onlyoffice is already installed
-print_text_in_color "$ICyan" "Checking if Onlyoffice is already installed..."
-if version_gt "$CURRENTVERSION" "18.0.1" && ! does_this_docker_exist 'onlyoffice/documentserver'
+# Check if Onlyoffice is installed using the old method
+if does_this_docker_exist 'onlyoffice/documentserver'
+then
+    if version_gt "$CURRENTVERSION" "18.0.1"
+    then
+        print_text_in_color "$ICyan" "Your server is compatible with the new way of installing Onlyoffice. We will now remove the old docker and install the app from Nextcloud instead."
+        # Remove docker image
+        docker_prune_this 'onlyoffice/documentserver'
+        # Revoke LE
+        SUBDOMAIN=$(whiptail --title "T&M Hansson IT - OnlyOffice" --inputbox "Please enter the subdomain you are using for OnlyOffice, eg: office.yourdomain.com" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+        if [ -f "$CERTFILES/$SUBDOMAIN/cert.pem" ]
+        then
+            certbot revoke --cert-path "$CERTFILES/$SUBDOMAIN/cert.pem"
+        fi
+        # Remove Apache2 config
+        if [ -f "$SITES_AVAILABLE/$SUBDOMAIN.conf" ]
+        then
+            a2dissite "$SUBDOMAIN".conf
+            restart_webserver
+            rm "$SITES_AVAILABLE/$SUBDOMAIN.conf"
+        fi
+        # Remove app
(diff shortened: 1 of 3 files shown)


apps/onlyoffice.sh
```diff
@@ -21,31 +18,58 @@ debug_mode
 root_check
 
 # Nextcloud 13 is required.
```
enoch85: need to change comment

szaimen:
I will have a look on this later on. But actually I would like to see https://github.com/nextcloud/vm/pull/1072 getting merged first.

szaimen:
Don't you think, that we should retain the possibility to install onlyoffice the legacy way, since this would be recommended e.g. for NC 18.0.0 ?

apps/onlyoffice.sh
```diff
@@ -21,31 +18,58 @@ debug_mode
...
-
 # Get the latest packages
 apt update -q4 & spinner_loading
 
-# Check if Nextcloud is installed
-print_text_in_color "$ICyan" "Checking if Nextcloud is installed..."
-if ! curl -s https://"${NCDOMAIN//\\/}"/status.php | grep -q 'installed":true'
+# Check if onlyoffice is already installed
+print_text_in_color "$ICyan" "Checking if Onlyoffice is already installed..."
+if version_gt "$CURRENTVERSION" "18.0.1" && ! does_this_docker_exist 'onlyoffice/documentserver'
```
szaimen: `&& ! does_this_docker_exist 'onlyoffice/documentserver'`

This was here to be able to do it the legacy way instead. So probably not needed here, if we get rid of the legacy way.

apps/onlyoffice.sh
```diff
@@ -21,31 +18,58 @@ debug_mode
...
+then
+    install_if_not jq
+    if occ_command_no_check app:list --output=json | jq -e '.enabled | .documentserver_community' > /dev/null
+    then
+        choice=$(whiptail --radiolist "It seems like 'Onlyoffice' is already installed.\nChoose what you want to do.\nSelect by pressing the spacebar and ENTER" "$WT_HEIGHT" "$WT_WIDTH" 4 \
```
szaimen: I think we should decide first, if a whiptail is the way to go here, since we could do it like I have suggested it here: https://github.com/nextcloud/vm/pull/1072#issuecomment-586652048

apps/onlyoffice.sh
```diff
@@ -21,31 +18,58 @@ debug_mode
...
-on your domain to be able to run this script.
-If you use the Nextcloud VM you can use the Let's Encrypt script to get SSL and activate your Nextcloud domain.
-When SSL is activated, run these commands from your terminal:
-sudo curl -sLO $APP/onlyoffice.sh
-sudo bash onlyoffice.sh"
-    exit 1
+msg_box "Sorry, but Nextcloud needs to be run on HTTPS which doesn't seem to be the case here.
+
+You easily activate TLS (HTTPS) by running the Let's Encrypt script found in $SCRIPTS.
+More info here: https://bit.ly/37wRCin
```
szaimen: We don't need https here, to make it work, since it also works without https.
What I wanted to do is to inform the user, that he can reinstall it later on, when ssl is activated to make it work again after https is activated.

apps/onlyoffice.sh
```diff
@@ -21,31 +18,58 @@ debug_mode
...
+
+To run this script again, just exectue 'sudo bash $SCRIPTS/additional_apps.sh' and choose OnlyOffice."
```
szaimen: isn't it better to teach the user:
```
sudo -i
additional apps
```
?

Otherwise, the correct command would be:
'sudo bash $SCRIPTS/apps.sh'

apps/onlyoffice.sh
szaimen: no, that is not right. is better if we just allow https.

szaimen:
> @szaimen This is how I would do it.
> 
> Doesn't seem to work on my installation though.

then it is probably a bad link. Could you check if the link that is filled in the onlyoffice app is called something like:
https://youromain.com/apps/documentserver_community/
?
If not, this is something we could check:
or better: we could let the script insert into the onlyoffice field the following: `$(occ_command_no_check config:system:get overwrite.cli.url)/apps/documentserver_community/`  then it should work always.

apps/onlyoffice.sh
```diff
@@ -21,31 +18,58 @@ debug_mode
...
+    exit
 fi
 
 # Check if apache2 evasive-mod is enabled and disable it because of compatibility issues
```
szaimen: we should test, if that is really needed here. when installing it this way...

enoch85:
> Don't you think, that we should retain the possibility to install onlyoffice the legacy way, since this would be recommended e.g. for NC 18.0.0 ?

No, we push for users to use the latest version of Nextcloud ofc. Keeping stuff backwards compatible would mean a lot more work for us, while the only thing the user has to do is to upgrade.

> This was here to be able to do it the legacy way instead. So probably not needed here, if we get rid of the legacy way.

Yes, because if they use legacy and want the new method, this is how it's done. Hence the priue docker stuff as well. Though you wanted I check for it? I don't care really. The goal here is to get the latest and the greatest.

> https://yourdomain.com/apps/documentserver_community/

Then we should set that in the script, because it didn't set when  I tried (like 10 times). Also, the server I'm testing on is behind a Nginx Reverse Proxy which doesn't seem to work either way.

apps/onlyoffice.sh
```diff
@@ -20,32 +17,60 @@ debug_mode
...
-
 # Get the latest packages
 apt update -q4 & spinner_loading
 
-# Check if Nextcloud is installed
-print_text_in_color "$ICyan" "Checking if Nextcloud is installed..."
-if ! curl -s https://"${NCDOMAIN//\\/}"/status.php | grep -q 'installed":true'
+# Check if onlyoffice is already installed
+print_text_in_color "$ICyan" "Checking if Onlyoffice is already installed..."
+if version_gt "$CURRENTVERSION" "18.0.1" && ! does_this_docker_exist 'onlyoffice/documentserver'
```
szaimen: Actually, if you don't delete the ! before does_this_docker_exist, it will directly exit the if-then here.

apps/onlyoffice.sh
```diff
@@ -85,126 +99,30 @@ then
...
-    echo "exit"
-    } > "$SCRIPTS/dockerprune.sh"
-    chmod a+x "$SCRIPTS/dockerprune.sh"
-    crontab -u root -l | { cat; echo "@weekly $SCRIPTS/dockerprune.sh"; } | crontab -u root -
-    print_text_in_color "$ICyan" "Docker automatic prune job added."
-    service docker restart
-    docker restart onlyoffice
-    print_text_in_color "$IGreen" "OnlyOffice is now successfully installed."
-    any_key "Press any key to continue... "
+    occ_command config:app:set onlyoffice DocumentServerUrl --value=https://"$(occ_command_no_check config:system:get overwrite.cli.url)/"
```
szaimen: that command isn't exaclty correct. it should be:
`occ_command config:app:set onlyoffice DocumentServerUrl --value="$(occ_command_no_check config:system:get overwrite.cli.url)apps/documentserver_community/"`

szaimen:
> No, we push for users to use the latest version of Nextcloud ofc. Keeping stuff backwards compatible would mean a lot more work for us, while the only thing the user has to do is to upgrade.

Then we should probably add a hint, that the user should upgrade via the update script to be able to install onlyoffice.

apps/onlyoffice.sh
enoch85: Nice catch!

enoch85:
OK, I feel confident with this. tested on a clean VPS on Digitalocean and it worked.

Please test the following scenarios:
1. Collabora is previously installed
2. Onlyoffice is previously installed using the old method
3. Onlyoffice is previously installed using the new method
4. Running on Nextcloud 17.0.3
5. Running on Nextcloud 18.0.0
5. Running on Nextcloud 18.0.1

szaimen:
Before testing this, I would work on finishing (and testing) https://github.com/nextcloud/vm/pull/1072

PR #1088, opened 2020-02-18
remove every trace, not just the live version

enoch85:
@szaimen is this a correct use of  a for loop?


szaimen:
If your find command works, correct would probably be:
```
REMOVE_OLD="$(find "$LETSENCRYPTPATH/" -name "$SUBDOMAIN*")"
            for remove in $REMOVE_OLD
            do 
                rm -rf "$LETSENCRYPTPATH/$remove"
            done
```
Or
```
REMOVE_OLD="$(find "$LETSENCRYPTPATH/" -name "$SUBDOMAIN*")"
            for remove in $REMOVE_OLD
            do 
                rm -rf "$remove"
            done
```

PR #1091, opened 2020-02-19
Harden collabora.sh

szaimen:


szaimen:
The script has worked quite nicely in my testing, hence marking as ready for review.

enoch85:
@szaimen 

**Reinstall:**
1. No need to remove the TLS cert as it will be renewed anyway.
2. No need to remove the Apache Vhost as it will be overwritten anyway.

PR #1092, opened 2020-02-19
Harden fail2ban

szaimen:


szaimen:
this also has worked well in my testing, hence marking as ready for review, too.

PR #1093, opened 2020-02-19
Harden fulltextsearch

szaimen:


szaimen:
@enoch85 Fulltextsearch doesn't seem to work after an reinstallation. Do you have and idea why?
If not, we will probably have to delete the option to reinstall the app and just  report to the user, that the app is already installed and doesn't support reinstalling like bitwarden.

enoch85:
It worked before this change, so just copy what I did previously and it will work.

szaimen:
Thanks for the tip, but actually I have copied already what you did previously and the reinstallation and the script at whole works but when I try to search for something in Nextcloud after the reinstallation, it doesn't show any result (although there are files, that should get found e.g. when searching for png or jpg)
Maybe it works in your testings?

enoch85:
@Ark74 Do you approve these changes? Is there anything that can be made better?

Thanks!

Ark74:
Well, if you are making a reinstallation you might wanna clean the ES db too forcing a new index, to avoid issues among installations and possible version changes.
php ~/occ fulltextsearch:reset


In case of removal/uninstall, you might wanna delete the docker volume, not only the app (ES) container.
docker stop

enoch85:
Tested and WORKS! :D

szaimen:
Nice! 🎉

PR #1094, opened 2020-02-19
Harden netdata

szaimen:


szaimen:
this has worked this way in my testing, hence marking as ready for review.

PR #1096, opened 2020-02-19
Harden tmbitwarden

szaimen:


szaimen:
the check has worked in my testing. hence marking as ready for preview.

apps/tmbitwarden.sh
```diff
@@ -18,6 +17,25 @@ debug_mode
 # Check if root
 root_check
 
+# Check if Bitwarden is already installed
+print_text_in_color "$ICyan" "Checking if Bitwarden is already installed..."
+if [ "$(docker ps -a >/dev/null 2>&1 && echo yes || echo no)" == "yes" ]
+then
+    if docker ps -a --format '{{.Names}}' | grep -Eq "bitwarden";
```
enoch85: does `is_this_docker_installed` work here?

enoch85:
Much easier to follow your changes now.

I think this looks good, just want to know if we can use the existing function or not.

apps/tmbitwarden.sh
enoch85: Nope, since it needs to be specific, like 'bitwarden/admin' for example

Issue #1082, opened 2020-02-15
make a function from "check if a specific nextcloud-app is enabled"


szaimen:
as the title says.
ref https://github.com/nextcloud/vm/pull/1081#discussion_r378848659

I will work on this when #1081 and #1072 are merged.

enoch85:
IMO, no need to add logic for that as that's already handled here: https://github.com/nextcloud/vm/blob/master/lib.sh#L680-L711

szaimen:
I would say it is definitely needed and would work like this:

```
is_this_enabled() {
if occ_command app:list --output=json | jq -e '.enabled | .$1' > /dev/null
then
    return 1
else
    return 0
fi
}
```

e.g. we could use it in notify_admin_gui
or now in the PR of new way of installing onlyoffice

enoch85:
OK, please name it `is_app_enabled` and add it to lib.sh, then change all the scripts where this is needed.

Advice, clone this repo and use `grep -r` to search foru occurances like "occ:app enable"

szaimen:
I would like to do this when the other PRs are merged because of conflicts...

szaimen:
### The are the outputs when I grep for:

**no_check:**
```
vm/apps/collabora.sh:if occ_command_no_check app:list --output=json | jq -e '.enabled | .documentserver_community' > /dev/null
vm/apps/onlyoffice.sh:    if occ_command_no_check app:list --output=json | jq -e '.enabled | .documentserver_community' > /dev/null
vm/apps/previewgenerator.sh:if [ -n "$(occ_command app:list | grep encryption | awk '{print $3}')" ]
vm/lib.sh:if ! occ_command_no_check app:list --output=json | jq -e '.enabled | .notifications' > /dev/null
vm/static/nextcloud_configuration.sh:            if ! occ_command app:list --output=json | jq -e '.enabled | .text' > /dev/null
```

**occ app:enable:**
`vm/lib.sh:    sudo -u www-data php ${NCPATH}/occ app:enable "$1"`

**command app:enable:**
```
vm/apps/spreedme.sh:occ_command app:enable spreedme
vm/lib.sh:            occ_command app:enable "$1"
```

szaimen:
I could also provide a function: is_app_installed, that would look like this:
# example: is_app_installed documentserver_community
is_app_installed() {
if [ -d "$NC_APPS_PATH/$1" ]
then
    return 0
else
    return 1
fi
}

we could use that here:
```
[...]

PR #1100, opened 2020-02-21
create function is_app_enabled

szaimen:
fix https://github.com/nextcloud/vm/issues/1082

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -1139,8 +1139,7 @@ esac
 #
 # occ_command_no_check notification:generate -l "$2" "$admin" "$1"
 notify_admin_gui() {
-install_if_not jq
-if ! occ_command_no_check app:list --output=json | jq -e '.enabled | .notifications' > /dev/null
+if ! is_app_enabled notifications
 then
     print_text_in_color "$IGreen" "The notifications app isn't enabled - unable to send notifications"
     return 1
--- a/lib.sh
+++ b/lib.sh
@@ -695,12 +695,6 @@ fi
 }
 
 install_and_enable_app() {
-# Check if app is already installed
-if is_app_enabled "$1"
-then
-    return
-fi
-
 # Download and install $1
 if [ ! -d "$NC_APPS_PATH/$1" ]
 then
@@ -722,11 +716,11 @@ or when a new version of the app is released with the following command:
         if [ -d "$NC_APPS_PATH/$1" ]
         then
             occ_command app:enable "$1"
+            chown -R www-data:www-data "$NC_APPS_PATH"
         fi
     fi
 else
     print_text_in_color "$ICyan" "It seems like $1 is installed already, trying to enable it..."
-    # occ_command not possible here because it uses check_command and will exit if occ_command fails
     occ_command_no_check app:enable "$1"
 fi
 }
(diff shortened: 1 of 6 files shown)


szaimen:
I have tested the function and it works, hence marking as ready for review.

lib.sh
```diff
@@ -683,13 +683,30 @@ calc_wt_size() {
...
+else
+    return 1
+fi
+}
+
 install_and_enable_app() {
+# Check if app is already installed
+if is_app_enabled "$1"
+then
+    return
```
enoch85: What does this addition do?

enoch85: More importantly, why is it necessary?

lib.sh
```diff
@@ -705,14 +722,12 @@ or when a new version of the app is released with the following command:
         if [ -d "$NC_APPS_PATH/$1" ]
         then
             occ_command app:enable "$1"
-            chown -R www-data:www-data "$NC_APPS_PATH"
         fi
     fi
 else
     print_text_in_color "$ICyan" "It seems like $1 is installed already, trying to enable it..."
     # occ_command not possible here because it uses check_command and will exit if occ_command fails
```
enoch85: Maybe change the comment? :)

lib.sh
```diff
@@ -1128,8 +1143,7 @@ esac
 #
 # occ_command_no_check notification:generate -l "$2" "$admin" "$1"
 notify_admin_gui() {
-install_if_not jq
-if ! occ_command_no_check app:list --output=json | jq -e '.enabled | .notifications' > /dev/null
+if ! is_app_enabled notifications
```
enoch85: Maybe it's better to actually enable the app?

enoch85:
@szaimen Please check my comments and answer/fix.

Thanks!

lib.sh
szaimen: I think, that we don't have to inform the user if the app is enabled: then we just skip the whole install process and don't try to enable the app since it is already installed.

lib.sh
szaimen: Shall I just delete that line?

lib.sh
szaimen: I wouldn't say so because there are usecases where you would like to disable the notification app.

lib.sh
enoch85: Sure

lib.sh
enoch85: No, please leave it is as.

lib.sh
```diff
@@ -705,14 +722,12 @@ or when a new version of the app is released with the following command:
         if [ -d "$NC_APPS_PATH/$1" ]
         then
             occ_command app:enable "$1"
-            chown -R www-data:www-data "$NC_APPS_PATH"
```
enoch85: Please don't remove this line, it's there for a reason.

szaimen: okay, in my testing this is not needed (when installing apps through the store), but if you say so...

lib.sh
szaimen: this is reverted

lib.sh
szaimen: it is deleted

lib.sh
szaimen: this is reverted

Issue #1104, opened 2020-02-22
Bug in notify admin


enoch85:
From a customer:

```
+ install_if_not jq
+ dpkg-query -W '-f=${Status}' jq
+ grep -q 'ok installed'
+ occ_command app:list --output=json
+ check_command sudo -u www-data php /var/www/nextcloud/occ app:list --output=json
+ jq -e '.enabled | .notifications'
+ sudo -u www-data php /var/www/nextcloud/occ app:list --output=json
+ return 0
++ occ_command_no_check user:list --output=json
++ jq -r 'keys[]'
++ sudo -u www-data php /var/www/nextcloud/occ user:list --output=json
+ CHECK_USERS='Allgemein PAP
Allgemein Verein
Daniela Batliner
Egon Gstoehl
Elmar Kranz
Karl Gassner
Manfred Meier
Marlies Amann
Martin Buechel
Panja Belzner
Peter Laukas
Praesident AG
Praesident Verein
Rene Wanger
Roland Risch
Siegfried Risch
Thomas Batliner
Thomas Schaedler
Thomas Schaffhauser
Uwe Belzner
Vinzenz Batliner
ncadmin1'
+ print_text_in_color '\e[0;96m' 'Posting notification to users that are admins, this might take a while...'
+ printf '%b%s%b\n' '\e[0;96m' 'Posting notification to users that are admins, this might take a while...' '\e[0m'
Posting notification to users that are admins, this might take a while...
+ for admin in $CHECK_USERS
+ occ_command_no_check user:info --output=json Allgemein
+ sudo -u www-data php /var/www/nextcloud/occ user:info --output=json Allgemein
+ jq -e '.groups | index("admin")'
parse error: Invalid numeric literal at line 1, column 5
+ for admin in $CHECK_USERS
+ occ_command_no_check user:info --output=json PAP
+ jq -e '.groups | index("admin")'
+ sudo -u www-data php /var/www/nextcloud/occ user:info --output=json PAP
parse error: Invalid numeric literal at line 1, column 5
+ for admin in $CHECK_USERS
+ occ_command_no_check user:info --output=json Allgemein
+ sudo -u www-data php /var/www/nextcloud/occ user:info --output=json Allgemein
+ jq -e '.groups | index("admin")'
parse error: Invalid numeric literal at line 1, column 5
+ for admin in $CHECK_USERS
+ occ_command_no_check user:info --output=json Verein
[...]

PR #1105, opened 2020-02-22
make notify_admin_gui failproof

szaimen:
fix https://github.com/nextcloud/vm/issues/1104

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -1144,6 +1144,13 @@ then
 fi
 
 CHECK_USERS=$(occ_command_no_check user:list --output=json | jq -r 'keys[]')
+
+if [[ $CHECK_USERS = *" "* ]]
+then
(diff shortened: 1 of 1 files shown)


szaimen:
This has been already tested by me. And should fix the issue.

lib.sh
```diff
@@ -1144,6 +1144,13 @@ then
 fi
 
 CHECK_USERS=$(occ_command_no_check user:list --output=json | jq -r 'keys[]')
+
+if [[ $CHECK_USERS = *" "* ]]
+then
+    print_text_in_color "$IRed" "There are user existent, that contain spaces - unable to send notifications"
```
enoch85: @szaimen 

Green = Scucess
Cyan = Information
Red = Error

Yes, it's wrong to use spaces in usernames

szaimen: Okay, good to know!

Issue #1102, opened 2020-02-22
Move "lowest_compatible_nc()" check to the Reinstall) option


enoch85:
@szaimen As the title says, this check needs to be move directly under `reinstall)` in all the app scripts.

Otherwise the reinstall option will remove everything, before this check is made, which could cause the script to stop *after* everything is removed which would be the same as `uninstall)`

enoch85:
Here's an example: https://github.com/nextcloud/vm/pull/1093/files#diff-11f2c3c460c2116da5169cd6790f6274R62-R65

szaimen:
We could also move it to the very beginning of the script. What do you think about that?
Then also the deinstallation wouldnt work but we don't care since the version isn't supported...

enoch85:
Deinstallation should work.

I don't remember why I put it there, but I think it was due to the occ_command was introduced in NC 13. 

If we don't use `occ_command` during uninstall, put the check in `Reinstall)`, so that it's possible to uninstall. Otherwise, put it at the top.

Thanks!

szaimen:
Actually, I would like to put the nc check always to the top like in the collabora script (as well as ram and cpu check) because then we don't need it two times in the reinstall and install part of the script...

enoch85:
That's how it was designed from the start. All checks are supposed to be in the start. I dont know why you decided to changes it? 

I was too quick with merging I guess. :)

Please make a PR for it.

Thanks!

szaimen:
I think that had to do with the moving of the "installing the app" -notification, but I don't remember completely...

enoch85:
Thank you!

PR #1107, opened 2020-02-23
put lowest_compatible and ram/gpu_check to the right position

szaimen:
fix https://github.com/nextcloud/vm/issues/1102


szaimen:
I think that should be all. I've already checked the other app scripts.

Issue #1109, opened 2020-02-24
Update notify_admini_gui?


szaimen:
Hi, I've reported the notify_admin_gui error to the nextcloud-snap and kyrofa fixed the issue.
We could do the same: https://github.com/nextcloud/nextcloud-snap/pull/1262

enoch85:
I knew there was a fix for it. :D

Yes, please! :)

PR #1110, opened 2020-02-24
Update notify_admin_gui to support users with spaces

szaimen:
fix https://github.com/nextcloud/vm/issues/1109


szaimen:
I've already tested it and it works.

PR #1113, opened 2020-02-26
check if the apps got installed in the onlyoffice script

szaimen:
I recently upgraded to NC 18 and have run the onlyoffice script again. strangely the onlyoffice app wasn't installed after changing to the hub method of onlyoffice. So I think the check is worth it to track this.

### DIFF
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -157,5 +157,10 @@ then
     msg_box "OnlyOffice was successfully installed."
 fi
 
+if ! is_app_installed onlyoffice || ! is_app_installed documentserver_community
+then
+    msg_box "Something got wrong during the installation. Please report this here: $ISSUES"
+fi
+
 # Just make sure the script exits
 exit


szaimen:
Btw: i've tested enabling mod_evasive after migrating to the documentserver_community. And again: It doesn't work when mod_evasive is enabled. So good to have it disabled.

szaimen:
And a second BTW:
We should think about removing the domain from trusted domains if the docker gets pruned, shouldn't we?

enoch85:
> We should think about removing the domain from trusted domains if the docker gets pruned, shouldn't we?

It's impossible to know which order the trusted domains are set. We can't guess here as it could break the systems we run this on.

It doesn't hurt to have it there, it's just cosmetic imo.

szaimen:
I mean we could get all trusted domains and check against the domain, that we got from the user during the uninstallation of docker-onlyoffice but maybe that is a bit over the top?

enoch85:
Aah, no that's actually a good idea. It wouldn't involve any guessing, so let's make a PR out of it.

Thanks!

Issue #1114, opened 2020-02-27
install onlyoffice "something got wrong during the installation...."


Laurent-grain:
## Steps to reproduce
1.A fresh install of my vm nextcloud (vm downloaded fromm t&m hanson)
2.I launch the script for installing onlyoffice
3 Everything seems to be OK because It say to me: .onlyoffice was succefully installed. I press Ok, and after a message say "something got wrong during the installation....."

## Expected behaviour
Tell us what should happen

## Actual behaviour
Tell us what happens instead

## Server configuration

**Nextcloud server version:** (see your admin page)
18.0.1
**Server OS** (Ubuntu server is default)
ubuntu 18.04
**How did you install the VM?** (Scripted install from master **OR** Released version)
script
## Network
**Do you use DHCP?**
No
**Is port 80 and/or 443 open?**
Yes
#### Log file (/var/log/nextcloud/nextcloud.log)
```
Insert your nextcloud.log file here
```
the log file is empty
#### Installation log (command output)
```
Insert the CLI output here
```

BoardWatcher:
I have the same issue as well.  Exactly the same conditions as described above.  This was my first attempt to install NextCloud.  Everything said installation was successful as above.  SSL configuration also complete.

enoch85:
@szaimen The PR you made is now presented. 

How can we solve this?

BoardWatcher:
Hi;

Yes;  I've determined that you cannot have both collabroa and OnlyOffice
installed at the same time. Rerun the apps.sh file and choose Office
again.  It will offer to remove collabora and install just office and
you'll be good to go.

Sent Digitally,
Charles Naslund, CDR USNR RET, ASAE, BSCS, MBA
Charles@Naslund.nz <charles@naslund.nz>

szaimen:
Hm... Probably this is the problem here (in ainstall_and_enable_app):
installcmd="$(occ_command_no_check app:install "$1")"
(Do you see the double quotes?)
But I don't get why the documentserver_communty app installs without problems and the onlyoffice app just not at all...

enoch85:
@Laurent-grain Could you please provide the full CLI output log so that we can debug this issue?

Thanks!

szaimen:
I guess there was no output. But the check that I have added was for testing if onlyoffice or documentserver_community is installed at the end.

[...]

PR #1115, opened 2020-02-28
fix install_and_enable_app

szaimen:
probably fix https://github.com/nextcloud/vm/issues/1114


enoch85:
Nope that's not it. Travis complains.

PR #1118, opened 2020-02-28
Update all files to use local lib if existant

szaimen:


szaimen:
I think this is ready. Hence marking as ready for review
but @enoch85 do you have an idea why travis reports those shellcheck issues although they should be disabled?

enoch85:
Big PR again... Will take some time to review.

Btw, `/var/scripts/main` - where does the "main" come from? We don't have that.

szaimen:
Currently we don't have it but we will create it:
https://github.com/nextcloud/vm/pull/1120/files#diff-90b7ea30bc531dd051547e9ec21aeb32R19-R21

You have to see all PRs that are mentioned in the Issue alltogether. I just split it up that you are able to review it.

szaimen:
I changed the local path to the lib so that everyone who has the lib.sh in "$SCRIPTS" will not automatically use the local lib file, but at the same time we are able to move on. While staying backwards compatible

szaimen:
actually this PR should be fairly easy to review since I am only adding the possiblity to use the local lib if existant to all scripts.

szaimen:
closing in favour of https://github.com/nextcloud/vm/pull/1124

PR #1120, opened 2020-02-29
create nextcloud_vm

szaimen:


szaimen:
@enoch85 I need help here: the idea is to have this one script, that just has to get downloaded once and works without needing other resources like lib.sh

for now two questions:
- Do we want to verify the integrity of the downloaded package?
- what is a good place where to download the .tar file to?

enoch85:
In such a big change I think it's better to change the lib location to your repository first, just to be able to test all the changes before we push to master.

If this is breaking it's catastrophic.

szaimen:
we don't change the lib location: please look at https://github.com/nextcloud/vm/pull/1120/files#diff-90b7ea30bc531dd051547e9ec21aeb32R19-R21

szaimen:
Actually this script will be a bit like cloning the repo to "$SCRIPTS"
but first we will download the release tar file and check the integrity

szaimen:
@enoch85 Please could you answer those two questions:

- Do we want to verify the state of the downloaded package with a checksum?
- what is a good place where to download the .tar file to?

szaimen:
closing in favour of https://github.com/nextcloud/vm/pull/1126

PR #1123, opened 2020-02-29
use the local scripts if existing

szaimen:


enoch85:
I still don't get why you added `/main`?

szaimen:
Because if we merge all PRs and check from now on if a local lib is existant in /var/scripts/ which is currently the case for everyone, it will also be used e.g. in the appscripts starting today although no one got asked if he wants to use the local lib. Also I don't find it important to have it in the /var/scripts directory.

enoch85:
Are you willing to visit all forums and search over the internet for all occurrences where the main scripts are linked? :)

szaimen:
> Are you willing to visit all forums and search over the internet for all occurrences where the main scripts are linked? :)

No, we don't move the main script as I already explained. It just gets move locally into the main folder.

szaimen:
Maybe this explains what we are doing:
https://github.com/nextcloud/vm/pull/1126/files#diff-90b7ea30bc531dd051547e9ec21aeb32R36-R56

enoch85:
> No, we don't move the main script inside the repository as I already explained. It just gets moved locally into the main folder.

Sounds reasonable :)

PR #1124, opened 2020-02-29
Update all files to use local lib if existant

szaimen:


szaimen:
@enoch85 any idea why travis reports these Shellcheck issues although they should be disabled?

enoch85:
[Here](https://github.com/nextcloud/vm/blob/master/nextcloud-startup-script.sh#L35-L62) it's only used one time.

szaimen:
Yes, I know. But it is a backwards compatibility decision...

enoch85:
Sorry, saw this now

PR #1126, opened 2020-02-29
create nextcloud_vm

szaimen:
for now two questions:

Do we want to verify the state of the downloaded package with a checksum?
what is a good place where to download the .tar file to?

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -794,6 +794,11 @@ fi
 # Force MOTD to show correct number of updates
 sudo /usr/lib/update-notifier/update-motd-updates-available --force
 
+if [ -f "$SCRIPTS"/nextcloud_install_production.sh ]
+then
+    rm "$SCRIPTS"/nextcloud_install_production.sh
+fi
+
 # Reboot
 print_text_in_color "$IGreen" "Installation done, system will now reboot..."
 reboot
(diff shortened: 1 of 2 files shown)


szaimen:
concerning this script, I mean we could include everything inside this into the install_production script. But i thought it would be good for making sure that this one file is as shown (you can compare it manually) so it should be as short as possible.
I mean otherwise the user could get the instructions if he wants to be absolutely sure to download the signed release file, extract the startup script and run the install_production script but I thought it would be good to have it in a separate file...

woa7:
> Do we want to verify the state of the downloaded package with a checksum?

IMHO: it must verify the downloadeds with at least checksum, 
or better verify it with gpg.

woa7:
> what is a good place where to download the .tar file to?

the normal place is /tmp,
best used with mktemp like
downloaddirVAR=$(mktemp -d -t NCVM.XXXXXX)
or
downloaddirVAR=$(mktemp -d)
(both work with the GNU coreutils and also with busybox only)

szaimen:
> IMHO: it must verify the downloadeds with at least checksum,
> or better verify it with gpg.

That was the plan 👍

szaimen:
> the normal place is /tmp,
> best used with mktemp like
> downloaddirVAR=$(mktemp -d -t NCVM.XXXXXX)
> or
> downloaddirVAR=$(mktemp -d)
> (both work with the GNU coreutils and also with busybox only)

Thanks! I will look into it later on!

szaimen:
superseded by https://github.com/nextcloud/vm/pull/1162

PR #1128, opened 2020-03-01
Fix infinite read when generating passwords; fix UTF-8 enabled systems

SirFerdek:
When locale is set to UTF-8, tr complains about invalid multibyte characters.
Fix this by setting locale for 'tr' command to LC_ALL=C

Fix infinite read from /dev/urandom when generating passwords.
The "fr | fold | head" chain never finishes execution on my setup.
Password is now generated by reading from /dev/urandom in 100-byte chunks.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -18,6 +18,18 @@ RORDIR=/opt/es/
 NC_APPS_PATH=$NCPATH/apps
 VMLOGS=/var/log/nextcloud
 
+# Helper function for generating random passwords
+gen_passwd() {
+    local length=$1
+    local charset="$2"
+    local password=""
+    while [ ${#password} -lt "$length" ]
+    do
+        password=$(echo "$password""$(head -c 100 /dev/urandom | LC_ALL=C tr -dc "$charset")" | fold -w "$length" | head -n 1)
+    done
+    echo "$password"
+}
+
 # Ubuntu OS
 DISTRO=$(lsb_release -sd | cut -d ' ' -f 2)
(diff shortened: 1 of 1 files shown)


enoch85:
Thanks for this PR!

Do you mind putting `$SHUF` inside the function as there's the only place it's used?

Thanks!

enoch85:
Sorry, scratch that. I saw now that there are different encoding in the passwords.

I will test this and merge if working. 

Thanks again!

lib.sh
```diff
@@ -1159,6 +1147,18 @@ do
...
+# Helper function for generating random passwords
+gen_passwd() {
+    local length=$1
+    local charset="$2"
+    local password=""
+    while [ ${#password} -lt "$length" ]
+    do
+        password=$(echo "$password""$(head -c 100 /dev/urandom | LC_ALL=C tr -dc "$charset")" | fold -w "$length" | head -n 1)
+    done
+    print_text_in_color "$Cyan" "$password"
```
SirFerdek: When you `print_text_in_color`, doesn't it actually change the output of this function, which is used as input when evaluating global variables? So now the passwords instead of "abcdefg1234" will be "<cyan escape chars>abcdefg1234<reset color escape chars>"?

SirFerdek: In other words: the echo was there to "return" the generated password, I don't really want to print it on screen in this place ;)

SirFerdek: I tried to run it, following errors are printed right after beginning of the installation:
`/dev/fd/63: line 30: print_text_in_color: command not found`

enoch85: I reverted my changes as I realized it wouldn't work.

PR #1130, opened 2020-03-01
Update nextcloud_install_production.sh

razorworks:
changed some comments,
add check for nginx and php-7.2


enoch85:
Thanks! :rocket:

PR #1132, opened 2020-03-01
remove trusted domain when uninstalling collabora/onlyoffice

szaimen:


szaimen:
I've already tested the function. Hence marking as ready for review.

enoch85:
Why a while loop?

The value is static, so that should be enough, or no?

szaimen:
Yes but we have to find out the slot that the domain is on to delete that slot. Otherwise we could use sed for that...

szaimen:
Ah no i am wrong. Can't we just delete the third one?
https://github.com/nextcloud/vm/blob/30aa9458589afb9953f4a0181f96bdd001cbd58b/apps/collabora.sh#L278

enoch85:
No, your first approach was better.

I still think there is an even better approach. Like, isn't it possible to just do `occ_command_no_check config:system:delete trusted_domains $SUBDOMAIN`?

szaimen:
> No, your first approach was better.
> 
> I still think there is an even better approach. Like, isn't it possible to just do `occ_command_no_check config:system:delete trusted_domains $SUBDOMAIN`?

no, that is unfortunately not possible. if you have only that domain it will work. but not if you have more domains inside trusted domains; it says that it has worked but it doesn't. So I think may approach is the way to go.

enoch85:
Can you please add back your code to the onlyoffice script, and also add this:

```
            # Disable Onlyoffice App if activated
            if is_app_installed onlyoffice
            then
                occ_command app:remove onlyoffice
            fi
```

... under both reinstall and uninstall? That way the scripts are equal.

Thanks!

enoch85:
The loop works, just tested. So I'm good with this when the requested stuff are added to `onlyoffice.sh`

enoch85:
I'm good with this :+1:

PR #1134, opened 2020-03-02
update automatic_updates...

szaimen:
to disable notifications for new app or software updates

### DIFF
--- a/static/automatic_updates.sh
+++ b/static/automatic_updates.sh
@@ -32,6 +32,7 @@ In the next step you will be able to choose to proceed or exit."
 
 if [[ "yes" == $(ask_yes_or_no "Do you want to enable automatic updates?") ]]
 then
+    occ_command config:app:set updatenotification notify_groups --value="[]"
     touch $VMLOGS/update.log
     crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/update.log"; } | crontab -u root -
     if [[ "yes" == $(ask_yes_or_no "Do you want to reboot your server after every update? *recommended*") ]]


szaimen:
I've already tested the command but not together with occ_command...
(I've only tried it with occ_command_no_check but it should work nonetheless)

szaimen:
What this does is it disables the notification about every single app update for admins and also disables notifications about Nextcloud-Updates so that only our notifiactions come through ;)
Also these are not needed because we run the update script every week and update NC and all apps

static/automatic_updates.sh
```diff
@@ -32,6 +32,7 @@ In the next step you will be able to choose to proceed or exit."
 
 if [[ "yes" == $(ask_yes_or_no "Do you want to enable automatic updates?") ]]
 then
+    occ_command config:app:set updatenotification notify_groups --value="[]"
     touch $VMLOGS/update.log
```
enoch85: Please run the permissions script here to make sure the update.log permissions are correct.

szaimen: do you mean this one?
https://github.com/nextcloud/vm/blob/master/static/setup_secure_permissions_nextcloud.sh
I don't get completely why but okay...

enoch85: Nvm, I was thinking wrong. :)

enoch85:
What if you would like to disable automatic updates? Then you might want to enable the regular notifications again.

szaimen:
you can easily do that by just selecting a group here:
image

szaimen:
We are only deleting the default admin group here.

enoch85:
The `occ command` works, but I don't know the consequences except what you're saying.

Do you think this is ready for merge?

enoch85:
OK, just tested, and it works.

PR #1141, opened 2020-03-04
update updatenotification to be able to run the cronjob every day

szaimen:


szaimen:
I've already tested my changes to the updatenotification.sh. Hence marking as read for review.

static/updatenotification.sh
```diff
@@ -19,18 +19,32 @@ debug_mode
...
+        sed -i "s/^REPORTEDNCVERSION.*/REPORTEDNCVERSION=$NCVERSION/" /$SCRIPTS/updatenotification.sh
     else
         notify_admin_gui \
-       "Update availabile!" \
+        "Update available!" \
         "Nextcloud $NCVERSION is available. Please run 'sudo bash /var/scripts/update.sh' from your CLI to update your server."
+        sed -i "s/^REPORTEDNCVERSION.*/REPORTEDNCVERSION=$NCVERSION/" /$SCRIPTS/updatenotification.sh
     fi
-else
-    print_text_in_color "$IGreen" "You already run the latest version! ($NCVERSION)"
```
enoch85: Why did you remove this?

enoch85: Aah, saw now that it's moved.

szaimen:
@enoch85 this PR should be ready to go...

nextcloud_install_production.sh
```diff
@@ -395,7 +395,7 @@ occ_command config:system:set upgrade.disable-web --value="true"
 print_text_in_color "$ICyan" "Configuring update notifications specific for this server..."
 download_static_script updatenotification
 check_command chmod +x "$SCRIPTS"/updatenotification.sh
-crontab -u root -l | { cat; echo "59 $AUT_UPDATES_TIME * * 6 $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
+crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * * $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
```
enoch85: Please change pack to 59, we set 59 for a reason, don't remember exactly now, but 0 doesn't work.

szaimen: Are you sure because in the current state, it will get executed at 18:59, no?

enoch85: Yes, I'm sure.

szaimen: this is done 👍

nextcloud_install_production.sh
```diff
@@ -395,7 +395,7 @@ occ_command config:system:set upgrade.disable-web --value="true"
 print_text_in_color "$ICyan" "Configuring update notifications specific for this server..."
 download_static_script updatenotification
 check_command chmod +x "$SCRIPTS"/updatenotification.sh
-crontab -u root -l | { cat; echo "59 $AUT_UPDATES_TIME * * 6 $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
+crontab -u root -l | { cat; echo "59 $AUT_UPDATES_TIME * * * $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
```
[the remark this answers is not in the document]
szaimen: Okay, that makes sense 👍
But what is a good time when we run it every day?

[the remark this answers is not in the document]
enoch85: I actually don't think we should send notifications more often than once per week.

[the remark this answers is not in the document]
szaimen: Yes, we are not sending a notification every day. I've managed to do the following with this PR:
If the notification about a new software update was sent once, no notification will be sent until the next update is available. But we check for it every day. And are able to send a notification to the user as soon as the update becomes available.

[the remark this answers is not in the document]
szaimen: @enoch85 You can test if it works by running the script and add manually below REPORTEDNCVERSION:
`NCVERSION=18.0.2`
when you 
run the script now, it will report, that 18.0.2 is available. But if you run it a second time it will send no notification because the varaiable REPORTEDNCVERSION will be set to 18.0.2
If you then set NCVERSION e.g. to 18.0.3 it will report for the user once that 18.0.3 is available but only this one time because it will set REPORTEDNCVERSION to 18.0.3... and so on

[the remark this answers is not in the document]
szaimen: I've already tested it by myself and it works!

szaimen:
@enoch85 any news on this PR?
It isn't related to the other PR's so it should be relatively easy to review.

szaimen:
> we always use `|` in sed commands. FYI. :)

I didn't know that this is possible. But if it works 👍

PR #1143, opened 2020-03-04
definitely fix the onlyoffice install error

### LINKED ISSUE #1114: install onlyoffice "something got wrong during the installation...."

szaimen:
fix #1114

### DIFF
--- a/apps/onlyoffice.sh
+++ b/apps/onlyoffice.sh
@@ -41,8 +41,8 @@ fi
 # Check if OnlyOffice is installed using the old method
 if does_this_docker_exist 'onlyoffice/documentserver'
 then
-    # Greater than 18.0.0 is 18.0.1 which is required
-    if version_gt "$CURRENTVERSION" "18.0.0"
+    # Greater than 18.0.1 is 18.0.2 which is required
+    if version_gt "$CURRENTVERSION" "18.0.1"
     then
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 Please merge this only if 18.0.2 is released. Then for most people the installations is still possible. Only for a few people wo have a connection that is slower than 2mb/s it will not work.
I'm sorry for overseeing this!

enoch85:
We have `git_apply_patch` ;)

Just tested and it doesn't work because the it include tests, but this works:

`git apply --exclude=tests/lib/InstallerTest.php /tmp/19439.patch`

szaimen:
ah, didn't know, that we have this :D

enoch85:
Let's merge this now, 18.0.2 is released in two days, and this will urge people to upgrade once out.

PR #1146, opened 2020-03-08
Add port check in cronjob

enoch85:
Question is, should we add this and potentially send out notifications every day, or should we just remind the user with a `msg_box`that port 80 needs to be open?

### DIFF
--- a/lets-encrypt/test-new-config.sh
+++ b/lets-encrypt/test-new-config.sh
@@ -56,6 +56,11 @@ cat << CRONTAB > "$SCRIPTS/letsencryptrenew.sh"
 #!/bin/sh
 if ! certbot renew --quiet --no-self-upgrade > /var/log/letsencrypt/renew.log 2>&1 ; then
         echo "Let's Encrypt FAILED!"--$DATE >> /var/log/letsencrypt/cronjob.log
+        # Check if port 80 is open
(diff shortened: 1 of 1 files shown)


szaimen:
I am voting for both, so that if you run it manually, it will remind you, and if it isn't open, you get informed by the nexcloud notification.

enoch85:
> you get informed by the nexcloud notification.

Problem is that the check will fail to see if the port is open, and should those users get spammed every day with a notification, even if the port actually is open?

szaimen:
You are right. So then maybe just inform the user if the cert renewal wasn't successful over notify_admin_gui and say that it could be caused by port 80 not beeing open?
But it seems like xou already planned to do it this way...
Of course we could store the info that the notification was already sent in the scriot itself, a bit like I am planning to do it https://github.com/nextcloud/vm/pull/1141 ...

szaimen:
But actually I would always report over notifY_admin_gui if a cert renewal wasn't successfull. And just say in there that it could be caused by port 80 not being open...

szaimen:
I mean what would you say how many notifications are okay? Every day for 7 days? Or just every second day? Or what would be the best in your opinion?

enoch85:
No, let's close this. I don't want to spam users with notifications. Will make a new PR with a msg_box instead.

PR #1147, opened 2020-03-10
add a reminder about port 80

enoch85:
cc @szaimen Does that make things clear enough?


szaimen:
> Does that make things clear enough?

I would say yes 👍

Issue #1129, opened 2020-03-01
FullTextSearch stopped working and won't reinstall successfully


BetaLeaf:
## Steps to reproduce
1. Attempt to reinstall FullTextSearch.sh ("sudo bash /var/scripts/apps/fulltextsearch/sh")

## Expected behaviour
It should work

## Actual behaviour
Doesn't work correctly

## Server configuration

**Nextcloud server version:** (see your admin page)
18.0.1

**Server OS** (Ubuntu server is default)
Ubuntu 18.04.3

**How did you install the VM?** (Scripted install from master **OR** Released version)
Scripted Install from Master

## Network
**Do you use DHCP?**
Yes, but ip assigned on router

**Is port 80 and/or 443 open?**
80 is open but not being used by Nextcloud (cameyo server on port 80)
443 is open and is the intended port to use

#### Installation log (command output)
```
ncadmin@cloud:~$ sudo bash /var/scripts/apps/fulltextsearch.sh
[sudo] password for ncadmin:
Installing Elastic Search & Full Text Search on Nextcloud...
Removing old version of Full Text Search and resetting the app...

In RunningService.php line 86:

  Index is already running


fulltextsearch:reset [--output [OUTPUT]] [--] [<provider>]

fulltextsearch disabled
fulltextsearch_elasticsearch disabled
files_fulltextsearch disabled
vm.max_map_count=262144
Max map count already set, skipping...
Press any key to continue. Press CTRL+C to abort
ffcad70a2e1a
Deleted Containers:
ffcad70a2e1affd84678c83af8301027d9851f883cfa80eb2c52a5ea1870d549

Total reclaimed space: 464.9kB
Deleted Images:
untagged: ark74/nc_fts:latest
untagged: ark74/nc_fts@sha256:ad184799b34904d44f9f1f26fc03f20c74c543de2181cd02a45af31db7e33377
deleted: sha256:4e6f05f7a75617aae94b26332559842db61c4b43184584f8da4ad2e4fb705580
deleted: sha256:f4506f3b910ac7bf88a51222db67de3804c0dcfe538d60866f9cb14c2de49dc2
deleted: sha256:099ff5b4bc19313d1ece09920d705d367e9a25bce1aa5fa1f916a79798f1a9ad
deleted: sha256:b928a21b4b8a5000e94ee877b74b8f9424e7f6c1a7227c2c7f64c7bd83417681
[...]

PR #1154, opened 2020-03-22
Update fulltextsearch to truncate the database table

szaimen:
Fix https://github.com/nextcloud/vm/issues/1129?


szaimen:
@enoch85 I haven't tested this but it should work.

Issue #1157, opened 2020-03-25
Scripted install, warning on 2GB machine


lukas-buergi:
## Steps to reproduce
1. Run nextcloud_install_production.sh on a machine with "2GB of RAM"

## Expected behaviour
It should not complain about having less than 2GB of RAM if the amount of RAM is close to 2GB

## Actual behaviour
It complains about not having 2GB of RAM on the "2GB" machine, because a few megabytes are missing (i.e. 19xx MB of RAM, probably because some small amount needs to be permanently mapped to some device so not all of the 2GB show up to the OS/userspace).

## Server configuration

**Nextcloud server version:** current

**Server OS** debian stable

## Suggested solution
Check for something like 100MB less than the round GB number, i.e. 1900MB.

enoch85:
Hi!

Thanks for your report.

RAM is read in 1024 bits, meaning that 2 GB is really 2048MB. So, If you set 2000MB RAM, then it will be 1.9GB RAM which is too small, yes. 

Not a bug.

enoch85:
Sorry, it's hardcoded. 

Here's the function of you're interested in improving it: https://github.com/nextcloud/vm/blob/master/lib.sh#L620-L639

Please send a PR if you do.

lukas-buergi:
@enoch85 Sure, here you go: https://github.com/nextcloud/vm/pull/1158

enoch85:
As it says in the `msg_box` when the check fails - you can disable it, on your own risk.

PR #1158, opened 2020-03-25
Accept 90% of demanded RAM as sufficient

lukas-buergi:
Usually not all RAM is available/listed in /proc/meminfo so a 2GB/8GB/whatever device usually has up to 10% less. The proposed change enables all devices labelled/marketed as a certain amount of GB to actually pass the test for this amount of GB. Otherwise none of my two tested devices pass the test for the amount of RAM that they are supposed to  have.

https://github.com/nextcloud/vm/issues/1157

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -622,10 +622,12 @@ fi
 # Example: ram_check 2 Nextcloud
 ram_check() {
 mem_available="$(awk '/MemTotal/{print $2}' /proc/meminfo)"
-if [ "${mem_available}" -lt "$((${1}*1002400))" ]
+mem_available_gb="$(echo "scale=2; $mem_available/(1024*1024)" | bc)"
+mem_required="$((${1}*(924*1024)))" # 100MiB/GiB margin less is also ok
+if [ "${mem_available}" -lt "${mem_required}" ]
 then
     print_text_in_color "$IRed" "Error: ${1} GB RAM required to install ${2}!" >&2
-    print_text_in_color "$IRed" "Current RAM is: ("$((mem_available/1002400))" GB)" >&2
+    print_text_in_color "$IRed" "Current RAM is: ("$mem_available_gb" GB)" >&2
     sleep 3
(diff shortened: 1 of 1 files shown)


enoch85:
Thank you very much for your PR, but it should be 100% or nothing at all. 

90% is not acceptable, not 99% either.

lukas-buergi:
> Thank you very much for your PR, but it should be 100% or nothing at all.
> 
> 90% is not acceptable, not 99% either.

Sure, then you'll just have to live with the fact that the average x GB machine fails the x GB check.

enoch85:
There's nothing wrong in the PR except that you check for 90%. If you change it to 100% I will test it and merge it if it's all good.

Thanks!

enoch85:
OK, I'm more fine with this.

What do you say?

enoch85:
Maybe you would want to have a look at this as well? https://github.com/nextcloud/vm/blob/master/lib.sh#L323-L327

lukas-buergi:
Well now the check is for 1000/1024=98% of demanded RAM in your most recent commit.

For 2/6/8GiB that amounts to 2048000/6144000/8192000 KiB. The systems I have available to check have 1951836/5583228/7994388 respectively so all fail the check for the marketed capacity (which is not a problem with my systems).

enoch85:
When I implemented this check (some years ago I belive?) I checked against VM RAM memory, specifically in VMware. All RAM sizes I tried was the exact sum of 1024*1024, which in my book is correct.

In other words, sests made on Ubuntu 16.04 / 18.04 with VMware ESXi 6.5 works with 100%, but for some reason seems to fail on your system. So, which system are you running?

lukas-buergi:
I'm talking about physical systems. If you set up a virtual machine it makes sense you get the amount of actually available RAM you ask for (even if that means the virtual machine actually occupies more physical memory than that). In a physical machine you usually don't get the full amount.

enoch85:
OK then, I rest my case. Let's merge this.

PR #1161, opened 2020-03-27
merge master

enoch85:


enoch85:
@szaimen Could you please create a PR where you change all the curl commands to testing instead, so that we can start working on your PRs separately?

I also want to start testing Ubuntu 20.04 soon so better get going with all the changes. :)

Thanks!

szaimen:
> @szaimen Could you please create a PR where you change all the curl commands to testing instead, so that we can start working on your PRs separately?
> 
> I also want to start testing Ubuntu 20.04 soon so better get going with all the changes. :)
> 
> Thanks!

I could do that but you are only able to test all changes that I have made to the scripts if this script is used first:
https://github.com/nextcloud/vm/pull/1126 and if you afterwards choose the "Security" option during the startup script.

PR #1162, opened 2020-03-28
merge nextcloud_vm into install_production

szaimen:
supersedes #1126

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -16,8 +16,7 @@ then
     fi
 
     # Test internet connection
-    ping -c1 -W1 github.com &>/dev/null
-    if [[ $? != 0 ]]
+    if ! ping -c1 -W1 github.com &>/dev/null
     then
         echo "Couldn't reach github.com. Exiting..."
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 This is ready for the testing branch. Although the ToDos are still missing, for testing all changes this should be already work.

enoch85:
@szaimen Thank you for this!

I don't know if you can push to the testing branch? If not, just let me know and I'll give you permissions. When that's done, just push as much as you want to the testing branch. :)

szaimen:
Yes, I can push to the testing branch, thanks! 
But I thought it would be good if you had a look at my changes... 
And I thought it is best if I push as least commits as possible to the branch which is best done with PR's, no?

enoch85:
> And I thought it is best if I push as least commits as possible to the branch which is best done with PR's, no?

Correct. 

This will get mixed up anyway as we will have conflicts from now on with master. 

So I'm thinking that if/when this is done, we push it to a new branch based of master, to be able to commit clean, but at the same time test the changes in testing.

PR #1172, opened 2020-04-15
Can elasticsearch container be bound only to localhost?

hatclub:
Is there any reason this container should expose ports to external network by default? 

I have seen that it's got the readonlyrest authenticating layer on it so I know it's not just "wide open", but it seems unnecessary from a security standpoint for it to be listening on 0.0.0.0 by default in this context?

### DIFF
--- a/apps/fulltextsearch.sh
+++ b/apps/fulltextsearch.sh
@@ -145,8 +145,8 @@ docker run -d --restart always \
 --name $fts_es_name \
 --ulimit memlock=-1:-1 \
 --ulimit nofile=65536:65536 \
--p 9200:9200 \
--p 9300:9300 \
+-p 127.0.0.1:9200:9200 \
+-p 127.0.0.1:9300:9300 \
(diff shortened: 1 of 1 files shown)


enoch85:
I'll need to consult @Ark74 on this.

Ark74:
Hi!
No not really, at some point it is useful for some minor admin task on elasticsearch I guess you could always do those running within the VM.
As you mention it, there is an authorization mechanism so, it's not open as such.

A PR on this regard is welcome.

Ark74:
@enoch85 if you don't have any issue.
@hatclub can fix his commit and we can merge it.

hatclub:
> @hatclub can fix his commit and we can merge it.

Sorry, what do you need from me here?

enoch85:
Fixed :)

So you'd say it's OK to merge @Ark74?

enoch85:
Just saw your approval... (I'm also blind)

PR #1173, opened 2020-04-15
Update fail2ban.sh

hatclub:
minor grammatical and typo fixes, messages adjusted for clarity (average user probably does not care or would not be surprised about settings being erased after asking for uninstall and they can't do anything about it by the time the message is displayed if they wanted to keep the config)

Moved erasure of locally generated files before purge of package to remove warning about not-empty directory (and allow the directories to be deleted)

Add check/deletion in case $VMLOGS exists as a file instead of a directory (as was the case on my system) otherwise attempt to mkdir won't work and this while loop goes around forever (What problem does the loop solve? should there be a max number of attempts if retries are regularly required?)

Edit generated jail.local - 
bantime increased to 14 days
ssh:
debian/ubuntu package jail.conf includes a jail called sshd already, just enable that and set maxretry, this also means we can avoid the crude sed of all "true" to "false" in debian-default as we won't get two jails even if we keep our enabled=true (not strictly necessary but maybe desired just in case upstream default changes) because all instances are called sshd 
nextcloud:
logpath was pointed at directory, not the log file earlier configured by the script

### DIFF
--- a/apps/fail2ban.sh
+++ b/apps/fail2ban.sh
@@ -31,20 +31,20 @@ then
 
     case "$choice" in
         "Uninstall Fail2Ban")
-            print_text_in_color "$ICyan" "Uninstalling Fail2Ban and resetting all settings..."
+            print_text_in_color "$ICyan" "Uninstalling Fail2Ban..."
             fail2ban-client unban --all
-            check_command apt purge fail2ban -y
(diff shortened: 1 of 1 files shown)


enoch85:
Thanks for this!

enoch85:
Looks good to me.

Issue #1155, opened 2020-03-22
Error on first setup


SomebodyOdd:
## Steps to reproduce
1. Download 18.0.2 version VM (I did 40 gb one, but I don't think that matters). **https://cloud.hanssonit.se/s/zjsqkrSpzqJGE9N**
2. Go through first setup phase. When asked for a keyboard layout, select Generic 105 (intl) layout, Russian. (as switch i picked Alt+Shift and default to all next steps about keyboard, but this might not matter).
<automatic reboot to set keyboard here, as per script, I guess>
3. When asked for a time zone, select Europe -> Moscow.
4. When asked for "find better mirror" say Yes.

## Expected behaviour
Script should go along.

## Actual behaviour
Got error instead (see screenshot below).

## Server configuration

**Nextcloud server version:** 
18.0.2 (based on a downloaded image)

**Server OS** (Ubuntu server is default)
Ubuntu server?
**How did you install the VM?** (Scripted install from master **OR** Released version)
Imported it into VMWare and followed scripts, that launched with first login
## Network
**Do you use DHCP?**
Yes, default NAT configuration of VMWare, which uses internal DHCP

**Is port 80 and/or 443 open?**
---
#### Installation log (command output)
Sorry, unable to copy from a vm console

#### Screenshoots

![изображение](https://user-images.githubusercontent.com/19632145/77259260-5c21d400-6c91-11ea-9437-b324b7376985.png)

enoch85:
Thanks for posting your issue!

`apt-select` can't handle two inputs, and in some rare cases it happens. Just skip selecting a new mirror for now, and I'll think of a fix in future versions.

SomebodyOdd:
Yeah, got to the end without any troubles after skipping this part. But I won't close the issue in case you need a reminder to fix this at some point! =)

enoch85:
A reminder is always good! 

Thanks for confirming!

Amalinda:
I just faced the same issue. Thanks for fixing this.

Amalinda:
I faced the same issue. However, the script should not be run twice. So, the simple fix is to not change the mirror.

enoch85:
@Amalinda Exactly what I would propose! I'm happy that users understand how it works. :)

enoch85:
The current fix is to remove the script all together from the next release. 

That's the "easy way out", but if someone has a fix, I'm happy to merge it!

enoch85:
[...]

PR #1176, opened 2020-04-15
20.04

enoch85:
TODO:

- [x]  Deprecate Ubuntu 16.04 from all scripts
- [x] Fix Nextcloud update script
- [ ]  Change base VMs to US keyboard
- [x]  Update documentation (activate-ssl.sh, Swedish keyboard)
- [x]  Don't allow PHP lower than 7.4 (and 7.2 for a while)
- [x]  Deprecate MySQL/MariaDB
- [x] Change to master branch
- [ ] ...


enoch85:
@szaimen New PR, this should be stable!

nextcloud-startup-script.sh
```diff
@@ -360,34 +337,6 @@ sed -i "s|;date.timezone.*|date.timezone = $(cat /etc/timezone)|g" "$PHP_INI"
...
-        apt-select
-    check_command apt-select -m up-to-date -t 5 -c -C "$(localectl status | grep "Layout" | awk '{print $3}')"
-    sudo cp /etc/apt/sources.list /etc/apt/sources.list.backup && \
-    if [ -f sources.list ]
-    then
-        sudo mv sources.list /etc/apt/
-    fi
-fi
-clear
-
```
szaimen, line 390: why did you remove that?

static/locales.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
+
+# shellcheck disable=2034,2059
+true
+# shellcheck source=lib.sh
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/20.04_testing/lib.sh)
+
```
szaimen, line 9: don't you need a check_root here?

old/locate_mirror.sh
```diff
@@ -0,0 +1,37 @@
+
+#!/bin/bash
+
+# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
+
+# shellcheck disable=2034,2059
+true
+# shellcheck source=lib.sh
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
```
szaimen, line 10: here also root_check?

README.md
```diff
@@ -32,7 +32,7 @@ Server installation. Simplified. :cloud:
 No problem, brave explorer! We made it simple. 
 
 In some cases we do pre-releases of the VM as well. Those can be found in the [TESTING](https://cloud.hanssonit.se/s/zjsqkrSpzqJGE9N?path=%2FTESTING) folder on the download server. If you want to try the latest version yourself, just follow the steps below:
-1. Download the latest [nextcloud_update.sh](https://raw.githubusercontent.com/nextcloud/vm/master/nextcloud_update.sh) to your server.
+1. Download the latest [nextcloud_update.sh](https://raw.githubusercontent.com/nextcloud/vm/20.04_testing/nextcloud_update.sh) to your server.
```
szaimen, line 35: does the this new update script really work on 18.04 (it probably won't work on 18.04, would it?)

szaimen:
if updating 20.04 in the current form, update.sh is downloading the master nextcloud_update.sh is this meant to be this way?

szaimen:
Of course if you want to merge into master I would restore the default lib path...

static/locales.sh
enoch85: True! Nice catch!

nextcloud-startup-script.sh
enoch85: `python-pip` doesn't exist in 20.04 yet, so I moved it to a script instead so that we easily can add it back later.

The script itself isn't perfect either: https://github.com/nextcloud/vm/issues/1155

If you have time, maybe you could have a look?

enoch85:
When I tested this everything was fine. I will change the `lib.sh` before pushing to master. :)

Thanks for testing!

enoch85:
Made the `KEYBOARD_LAYOUT` a lot faster (like 5 seconds), and used it in the scripts as it now loads the lib on every run anyway.

I feel satisfied!

nextcloud-startup-script.sh
enoch85: This is now fixed.

static/locate_mirror.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [[ "no" == $(ask_yes_or_no "Do you want to try to find a better mirror?") ]]
+then
+    print_text_in_color "$ICyan" "Keeping $REPO as mirror..."
+    sleep 1
+else
+    if [[ "$KEYBOARD_LAYOUT" =~ ,|/|_ ]]
+    then
+        msg_box "Your keymap contains more than one language, or a special character. ($KEYBOARD_LAYOUT)\nThis script can only handle one keymap at the time.\nWe will now exit."
+        exit 1
```
szaimen, line 28: Maybe a small heads up: would that cancel the whole startup script if it returns a 1?
I think the startup script should continue with the default mirror in that case, instead.

enoch85: The script will exit, and don't do any changes to the mirror, keeping the one that were there from the beginning.

szaimen:
BTW: I've tried the new version and the startup seems to need a lot of time: it hangs around 90s at A start job is running for /sys/devices/virtual/misc/vmbus!hv_kvp...
Also the default settings of 4GB RAM are a bit excessive?

enoch85:
Note to self: chage notify_admin_user to `|` instead of `/` in the `sed` command.

enoch85:
> start job is running for /sys/devices/virtual/misc/vmbus!hv_kvp...

That's an Ubuntu bug

> Also the default settings of 4GB RAM is a bit excessive?

This is just a testing VM, the new one will get 2 GB.

> And I don't know what has changed but I am no longer able to use the server-name of my smb-server

`smbclient` is no longer installed from the Ubuntu package repo. It's PECL and uses the latest version. Maybe the new version got more strict?

szaimen:
> > start job is running for /sys/devices/virtual/misc/vmbus!hv_kvp...
> 
> That's an Ubuntu bug
> 
> > Also the default settings of 4GB RAM is a bit excessive?
> 
> This is just a testing VM, the new one will get 2 GB.

Okay 👍

> > And I don't know what has changed but I am no longer able to use the server-name of my smb-server
> 
> `smbclient` is no longer installed from the Ubuntu package repo. It's PECL and uses the latest version. Maybe the new version got more strict?

The smb-script is using the package cifs-utils from the ubuntu repo. It works on 20.04 but is not able to connect to the smb-server via its server name. Only via ipaddress. I will have to investigate a bit more.

szaimen:
> The smb-script is using the package cifs-utils from the ubuntu repo. It works on 20.04 but is not able to connect to the smb-server via its server name. Only via ipaddress. I will have to investigate a bit more.

I've tested it a bit further and apparently the dns-resolving from another smb-host is working just fine. I will have to test this when I switch to 20.04 on my productive machine. It is possible that this is a test-machine related bug and that it will not occur on the productive-machine. Nonetheless using the ipaddress is also more stable especially if the smb-server is running all the time and therefore not changing the ipaddress.

enoch85:
OK, I feel confident with this now.

Please make a final check @szaimen in case I missed anything.

Thanks!

PR #1179, opened 2020-04-17
update notify_admin_gui to get rid of jq

szaimen:
fix https://github.com/nextcloud/vm/commit/faabf4a95b69b912572a19a0d07267f71bcf73c0#r38570608


szaimen:
@enoch85 I've already tested this and it works.

enoch85:
>  tested this and it works.

OK, let's try this.

PR #1181, opened 2020-04-18
Add option to disable user flows

szaimen:


szaimen:
@enoch85 Please read why that makes sense imo here: https://github.com/nextcloud/server/pull/19930#issue-387784337

enoch85:
Looks good

PR #1193, opened 2020-04-25
update notify_admin_gui

szaimen:


szaimen:
It works really reliably in my testing, hence marking as ready for review.

PR #1196, opened 2020-04-25
fix a bug with nextcloud_configuration

szaimen:
found this while testing the new release.


enoch85:
Oh, nice catch! :D

Thanks!

PR #1197, opened 2020-04-25
fix version_gt

szaimen:
Sry you were too fast with merging.
version_gt isn't working correctly without that.


szaimen:
BTW: the ssh fingerprint seems to have changed after the last reboot of the startup-script. 
Is it meant to be this way?

enoch85:
> BTW: the ssh fingerprint seems to have changed after the last reboot of the startup-script.
Is it meant to be this way?

Yes for security, otherwise all thousands of users would have the same SSH-key fingerprint. :)

And thanks for the fix!

enoch85:
All done?

szaimen:
> Yes for security, otherwise all thousands of users would have the same SSH-key fingerprint. :)

ah okay, makes sense 👍

PR #1201, opened 2020-04-27
update startup script and change netplan to true.

numb95:
update the startup script and change netplan `DHCP` values to true. ubuntu 20.04 (which is nextcloud vm based on) wouldn't accept `yes`. it should be t`rue`

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -76,8 +76,8 @@ network:
   renderer: networkd
   ethernets:
     $IFACE:
-      dhcp4: yes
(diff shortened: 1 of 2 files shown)


enoch85:
Thanks for your PR!

Please clone this repo and use `grep -r dhcpv4` to search for more occurances like this and then add them to this PR.

🚀

numb95:
@enoch85 Hi again. I edit another file and change the `no` to `false` to make it work in netplan.

enoch85:
Thank you! :+1:

PR #1205, opened 2020-04-29
update ssh instructions for windows

szaimen:


enoch85:
Good point and I like it! Problem is, what if you're not running Windows 1809 or later?

szaimen:
> Good point and I like it! Problem is, what if you're not running Windows 1809 or later?

Then you can use putty. Thats the reason why I haven't deleted those putty referring lines ;-)

enoch85:
I'm sorry to say that the text doesn't fit a console window.

Please narrow it down. Run the script inside both a Putty window, and a VMware / Hyper-V console window to see if t fits.

szaimen:
HM, that is not a problem on my system:
image
image

But I trust you if you say so.

What about this:

> To be 100% sure that all the keystrokes work correctly (like @),     
> please use an SSH terminal like Putty. If you are running Windows 
> 10 1809 or later, you can simply use SSH  from the command prompt.                                             
> Connect like this: ssh ncadmin@local.IP.of.this.server

enoch85:
Hmm, ok, let's merge it then. It didn't fit my Linux CLI prompt which I count as standard. ;)

PR #1210, opened 2020-05-01
possible fix for missing second disk import (zfs)

enoch85:
zpool.cache have a tendency to fail. This PR aims to make it more robust so that users don't manually have to import the zpool with `zpool import` just to get up and running in case the pool failed to import during boot.

New script https://github.com/nextcloud/vm/blob/master/static/change-to-zfs-mount-generator.sh

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -142,9 +142,6 @@ unset NCDB
 DEBUG=0
 debug_mode
 
-# Nextcloud 18 is required.
-lowest_compatible_nc 18
-
 # Check that this run on the PostgreSQL VM
 if ! is_this_installed postgresql-common
 then
@@ -172,6 +169,9 @@ fi
 
 # Set locales
 run_static_script locales
+
+# Nextcloud 18 is required
+lowest_compatible_nc 18
 
 # Is this run as a pure root user?
 if is_root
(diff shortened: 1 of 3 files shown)


enoch85:
@zalonic What's your view on this? (Since you helped out with the format-sdb script)

enoch85:
This is confirmed to have fixed it for at least two people with a broken system.

I will merge this to be released in Nextcloud 18.0.5/19.0.0 (whichever comes first).

It would still be nice with some input though @zalonic @penzoiders

penzoiders:
Unfortunately this fix is needed on ubuntu, on other systems (.rpm based) I never had such an issue. Your workaround looks fine. Although I can't test it in production since I don't use ZFS on my VMs (It's already underneath the storage layer, don't have much sense to have 2 slices of zfs for me).

enoch85:
Yeah, it's an Ubuntu bug. :(

Thanks for your input! :rocket:

zalonic:
Afternoon, sorry for not coming back to you sooner. I've checked the workaround and it seems okay so we've swapped out our Development Team NAS in production with the updated ZFS one referenced here. We are replicating the data back to the old stable one just in case as a backup and let you know on Friday if any issue crop up with the new configuration, as far as I can see so far, can't see any issues popping up.

enoch85:
Thanks @zalonic! Really appreciated!

zalonic:
Hi @enoch85, day 4 of running the updated NcVM, happy to say everything is operating as expected.

enoch85:
Thanks for the feedback! :D

PR #1211, opened 2020-05-01
Put docker-onlyoffice to old-directory

szaimen:
I think it makes sense to store the script here as long as the documentservery_community isn't really reliable, wich it still isn't in my testing.


szaimen:
@enoch85 I hope you are okay with that?

PR #1212, opened 2020-05-01
make onlyoffice work again

szaimen:
this seems to be needed to make it work again.

### DIFF
--- a/old/onlyoffice.sh
+++ b/old/onlyoffice.sh
@@ -5,9 +5,13 @@
 # shellcheck disable=2034,2059
 true
 # shellcheck source=lib.sh
-NC_UPDATE=1 && OO_INSTALL=1 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 this is the last one. have already successfull tested it. So no more PRs are needed to make it work again.

old/onlyoffice.sh
```diff
@@ -170,20 +174,20 @@ fi
...
-    if [ ! -f "$DHPARAMS" ]
+    if [ ! -f "$DHPARAMS_SUB" ]
     then
-        openssl dhparam -dsaparam -out "$DHPARAMS" 4096
+        openssl dhparam -dsaparam -out "$DHPARAMS_SUB" 4096
     fi
     printf "%b" "${IGreen}Certs are generated!\n${Color_Off}"
     a2ensite "$SUBDOMAIN.conf"
-    restart_webserver
+    check_command systemctl reload apache2
```
enoch85, line 186: Please revert this change, we use a function to only change on one place instead of several in case we need to change method.

szaimen: Is reverted 👍

Issue #1182, opened 2020-04-18
vague/unclear prompt "if you want to bypass this check..."


rjhornsby:
## Steps to reproduce
1. run nextcloud_install_production.sh
2. wait for the prompt in question that begins with "If you want to bypass this check..."

## Expected behaviour

After `nextcloud_install_production.sh` runs for a bit, it pauses to display an ncurses style prompt (see screenshot) that begins with "If you want to bypass this check..." and ends with "YOU HAVE BEEN WARNED!"

While there's the somewhat oblique reference about 3/4 into the lead sentence to something called "ram_check X" it's not clear what _this_ is referring to in the beginning "...bypass _this_ check..."

It's unclear what this prompt is trying to communicate. Certainly it's suggesting there are consequences for bypassing the check - but otherwise the dialog beginning with "If you want to bypass this check..." is confusing and vague.

## Actual behaviour

The only choice to the prompt is `<OK>`, so after puzzling over the dialog for a couple of minutes, hit `<OK>` - then and only then does it become clear what's going on because the script exits with an error:

```
Error: 2 GB RAM required to install Nextcloud!
Current RAM is: (.90 GB)
```

The prompt only makes sense with the added context of this error message. The prompt was raised because of an error. Unfortunately, nothing in the prompt itself (prior to seeing the error message, after accepting the `<OK>`) indicates a problem.

Maybe a prompt that includes something to the effect of "there's an error (not enough RAM), script will now exit. If you wish bypass the memory requirement check [do this], but note the consequences" would help.

Alternatively, if there was no prompt at all and just a quick "what to comment out" shell message along with the error that would be enough. The pause/prompt isn't really necessary - the script is about to exit anyway. IMHO.

## Server configuration

**Nextcloud server version:** 18.0.3

**Server OS** Raspbian Buster

#### Screenshots

attached below (sorry don't have an imgur account)

![Screen Shot 2020-04-18 at 14 40 02](https://user-images.githubusercontent.com/5077610/79670032-947bdb80-8185-11ea-8234-3212dadf3456.png)

enoch85:
Hi, 

The script is intended for Ubuntu Server, I can see that you run this on Buster. 

Please try Ubuntu 18.04 and report back if you got the same issue.

Thanks!

enoch85:
@rjhornsby Hey, feel free to make a PR with the changes you think is better.

The check is warning about that you can bypass the RAM and CPU check, but warns if you do.

[...]

PR #1219, opened 2020-05-04
Clarify low memory prompt #1182

rjhornsby:
Clarify whiptail prompt to include a specific error message re insufficient memory

Fix https://github.com/nextcloud/vm/issues/1182

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -631,7 +631,7 @@ fi
 ram_check() {
 install_if_not bc
 mem_available="$(awk '/MemTotal/{print $2}' /proc/meminfo)"
-mem_available_gb="$(printf '%0.2f\n' $(echo "scale=3; $mem_available/(1024*1024)" | bc))"
+mem_available_gb="$(printf '%0.2f\n' "$(echo "scale=3; $mem_available/(1024*1024)" | bc)")"
 mem_required="$((${1}*(924*1024)))" # 100MiB/GiB margin and allow 90% to be able to run on physical machines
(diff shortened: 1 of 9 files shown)


lib.sh
```diff
@@ -631,18 +631,20 @@ fi
 ram_check() {
 install_if_not bc
 mem_available="$(awk '/MemTotal/{print $2}' /proc/meminfo)"
-mem_available_gb="$(echo "scale=2; $mem_available/(1024*1024)" | bc)"
+mem_available_gb="$(printf '%0.2f\n' $(echo "scale=3; $mem_available/(1024*1024)" | bc))"
```
enoch85, line 634: Why this change if I may ask?

rjhornsby: It adds a leading zero, so instead of 

> .94GB installed, 2GB required

you get

> 0.94GB installed, 2GB required

which more closely aligns the formatting between the two numbers. When looking the rendered whiptail dialog, it was a bit more readable with the leading zero when quickly mentally comparing `0.94` and `2`

Screen Shot 2020-05-04 at 15 31 29

enoch85: OK, tested now and it looks good as far as I've tested.

enoch85:
Please fix travis with something like this: 

`mem_available_gb="$(printf '%0.2f\n' "$(echo "scale=3; $mem_available/(1024*1024)" | bc)")"`

Not tested with travis, but gets the correct output.

enoch85:
Thank you very much for your contribution, and welcome as a contributor to the Nextcloud VM! :)

Issue #1221, opened 2020-05-07
Upgrade Netplan YAML


enoch85:
# This is the network config written by 'subiquity'
network:
  ethernets:
    ens160:
      addresses:
      - 192.168.100.149/24
      gateway4: 192.168.100.254
      nameservers:
        addresses:
        - 192.168.100.20
        - 192.168.100.21
        search:
        - localdomain.local
  version: 2

PR #1224, opened 2020-05-07
make static ip great again + smaller fixes

enoch85:

Fix https://github.com/nextcloud/vm/issues/1221


enoch85:
@chwba Your fix is in here. FYI.

Issue #1223, opened 2020-05-07
Scripts that should go in configuration.sh


enoch85:
- Static IP
- TLS

Something you could work on @szaimen, if you have time?

szaimen:
@enoch85 are you sure with that?

I thought that configuration.sh was only for nextcloud configuration and this seems to be more a server configuration...

Also the nextcloud configuration isn't necessary to run during the install script, but static ip and tls is...

enoch85:
Hmm, forgot about that. Just think we should add another @szaimen fix (apps.sh, configuration.sh) for those scripts... I get the quesion several times per week, and it would be nice to be able to say:

"Run `sudo bash /var/scripts/extraconf.sh` to...[fix your issue]" .

Copy paste on the configuration and app scripts maybe?

szaimen:
Okay, have an idea how this could be done. Should it be called extraconf.sh?
Will come up with a POC

PR #1225, opened 2020-05-07
create server_configuraton

szaimen:
fix https://github.com/nextcloud/vm/issues/1223

### DIFF
--- a/static/server_configuration.sh
+++ b/static/server_configuration.sh
@@ -38,7 +38,7 @@ case "$choice" in
         clear
         run_static_script automatic_updates
     ;;&
-        *"Activate TLS"*)
+    *"Activate TLS"*)
         clear
         msg_box "The following script will install a trusted
 TLS certificate through Let's Encrypt.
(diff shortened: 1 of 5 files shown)


szaimen:
@enoch85 this is how I would do it.
Do you have already some thoughts on that?
It should already work but didn't test.

static/server_configuration.sh
```diff
@@ -0,0 +1,63 @@
...
+    *"Static IP"*)
+        clear
+        run_static_script static_ip
+        rm -f "$SCRIPTS"/lib.sh
+    ;;&
+    *"Automatic updates"*)
+        clear
+        run_static_script automatic_updates
+    ;;&
+        *"Activate TLS"*)
```
enoch85, line 41: this needs better indentation

szaimen: You are right. Is fixed.

enoch85:
Looks good to me, but let's wait with merging until it's tested.

szaimen:
BTW: do we need this here?
https://github.com/nextcloud/vm/blob/e5f07cc895ec557220d3977cc066bdb76502e541/static/static_ip.sh#L12-L32

enoch85:
> BTW: do we need this here?

Yes, since it should run also without internet.

szaimen:
> > BTW: do we need this here?
> 
> Yes, since it should run also without internet.

👍 
Would you add it?

enoch85:
Tested and works as intended.

PR #1228, opened 2020-05-08
Main menu

szaimen:
I think it is time to introduce a main menu since we have that many different menus already.
POC

### DIFF
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -17,7 +17,7 @@ debug_mode
 root_check
 
 # Install Apps
-choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar\nYou can view this menu later bei running 'sudo bash $SCRIPTS/apps.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Fail2ban" "(Extra Bruteforce protection)" OFF \
 "Adminer" "(PostgreSQL GUI)" OFF \
 "Netdata" "(Real-time server monitoring)" OFF \
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -17,7 +17,7 @@ debug_mode
 root_check
 
 # Install Apps
-choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar\nYou can view this menu later by running 'sudo bash $SCRIPTS/apps.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar\nYou can view this menu later by running 'sudo bash $SCRIPTS/menu.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Fail2ban" "(Extra Bruteforce protection)" OFF \
 "Adminer" "(PostgreSQL GUI)" OFF \
 "Netdata" "(Real-time server monitoring)" OFF \
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -17,7 +17,7 @@ debug_mode
 root_check
 
 # Install Apps
-choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar\nYou can view this menu later bei running 'sudo bash $SCRIPTS/apps.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar\nYou can view this menu later by running 'sudo bash $SCRIPTS/apps.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Fail2ban" "(Extra Bruteforce protection)" OFF \
 "Adminer" "(PostgreSQL GUI)" OFF \
 "Netdata" "(Real-time server monitoring)" OFF \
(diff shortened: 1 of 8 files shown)


static/change-root-profile.sh
```diff
@@ -44,9 +44,10 @@ ROOT-PROFILE
 # Add Aliases
 {
 echo "alias nextcloud_occ='sudo -u www-data php $NCPATH/occ'"
-echo "alias run_update_nextcloud='bash $SCRIPTS/update.sh'"
+echo "alias update_nextcloud='bash $SCRIPTS/update.sh'"
```
szaimen, line 47: I know we had that discussion already but I think that makes sense (is also only applicable to new users and hence doesn't destroy anything)

szaimen:
Maybe we should also mention the new script in this section:
https://github.com/nextcloud/vm/blob/a71ef7206002664203bdce534af245df9d3ff44b/nextcloud-startup-script.sh#L535

static/change-root-profile.sh
enoch85: Do you change the FAQ and DOCSas well? :)

static/nextcloud_configuration.sh
```diff
@@ -18,7 +18,7 @@ debug_mode
 root_check
 
 # Configure Nextcloud
-choice=$(whiptail --title "Nextcloud Configuration" --checklist "Which settings do you want to configure?\nSelect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+choice=$(whiptail --title "Nextcloud Configuration" --checklist "Which settings do you want to configure?\nSelect by pressing the spacebar\nYou can view this menu later bei running 'sudo bash $SCRIPTS/configuration.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
```
enoch85, line 21: typo: bei

static/server_configuration.sh
```diff
@@ -32,7 +32,7 @@ debug_mode
 root_check
 
 # Server configurations
-choice=$(whiptail --title "Server configurations" --checklist "Choose what you want to configure\nSelect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+choice=$(whiptail --title "Server configurations" --checklist "Choose what you want to configure\nSelect by pressing the spacebar\nYou can view this menu later bei running 'sudo bash $SCRIPTS/server_configuration.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
```
enoch85, line 35: typo: bei

apps/additional_apps.sh
```diff
@@ -17,7 +17,7 @@ debug_mode
 root_check
 
 # Install Apps
-choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+choice=$(whiptail --title "Which apps do you want to install?" --checklist "Automatically configure and install selected apps\nSelect by pressing the spacebar\nYou can view this menu later bei running 'sudo bash $SCRIPTS/apps.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
```
enoch85, line 20: same typo

enoch85:
Why not.. :)

Make it solid then I'll review. 

Thanks!

static/change-root-profile.sh
szaimen: Yes, would do that. Could you send me the links to the related sites? 

Am not really experienced with the docs.

szaimen: BTW: was was the reason to not host the docu on github?
would probably make the collaboration on docs a bit easier....

static/nextcloud_configuration.sh
szaimen: fixed

szaimen: fixed

static/server_configuration.sh
szaimen: fixed

apps/additional_apps.sh
szaimen: fixed

szaimen:
> Why not.. :)
> 
> Make it solid then I'll review.
> 
> Thanks!

i would say it is fairly robust now

static/change-root-profile.sh
szaimen: I've reverted the change because it doesn't have anything to do with the PR.
have opened #1233 for that.

szaimen:
I think this is ready for review now.

enoch85:
When I have time, sorry.

enoch85:
The changes I made will redirect users to use **one** single endpoint for configuring everything in Nextcloud. It's way easier then to remember, "hmm, what did this script do?"

It's also cleaner since the `$SCRIPTS` folder won't be flooded with a lot of different scripts.

Not tested, but it looks good.

enoch85:
OK, let's merge this. Won't affect the VM until next release anyway.

szaimen:
> The changes I made will redirect users to use **one** single endpoint for configuring everything in Nextcloud. It's way easier then to remember, "hmm, what did this script do?"
> 
> It's also cleaner since the `$SCRIPTS` folder won't be flooded with a lot of different scripts.
> 
> Also, easier to maintain as if we add another script, just put it in the main menu script.
> 
> Not tested, but it looks good.

Okay, makes sense, we should probably disable the downloading of them during the startup script?
(at least apps.sh and configuration.sh)
Having update.sh and server_configuration.sh on the server still makes sense imo.
Also for update.sh I think the code for executing update.sh still makes sense (since it could be possible, that you have written something to the file to be able to apply a rule after an update...)
```    "Update Nextcloud")
        if [ -f $SCRIPTS/update.sh ]
        if network_ok
        then
            bash $SCRIPTS/update.sh
        else
            if network_ok
            then
                run_static_script update
            fi
            run_static_script update
        fi
```

and we should revert those changes:
https://github.com/nextcloud/vm/pull/1228/files#diff-11ba175bd07d41f8a89a19da883fff86R20

enoch85:
**Docs updated** :D

> Okay, makes sense, we should probably disable the downloading of them during the startup script?

No, since they are used during the first setup. After the setup is done, you would easily get them with the menu script.

> Also for update.sh I think the code for executing update.sh still makes sense

Sorry, yes it does, since we have automatic updates. Will add your failproof method.

szaimen:
> **Docs updated** :D

Very Nice!

You seem to have thought already about everything. So nothing left to do here. 👍

Please delete the cleanup_after_main_menu branch.

enoch85:
> You seem to have thought already about everything.

Heehe, I try to imagine all possible scenarios, but it's always good with some help :)

PR #1230, opened 2020-05-08
Allow beta or RC versions

szaimen:
POC
how update.sh could be changed to accept new inputs to automatically set all variables for beta and release versions.

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -289,17 +289,23 @@ then
     rm -f /tmp/minor.version
 elif [ -f /tmp/prerelease.version ]
 then
-    if grep -q beta /tmp/prerelease.version
+    msg_box "WARNING, you are about to update to a Beta version of Nextcloud, there's no turning back, because you cannot downgrade. Please only continue if you have made a backup."
+    if [[ "no" == $(ask_yes_or_no "Do you really want to do this?") ]]
     then
-        NCREPO="https://download.nextcloud.com/server/prereleases"
-        NCVERSION=$(curl -s -m 900 $NCREPO/ | sed --silent 's/.*href="nextcloud-\([^"]\+\).zip.asc".*/\1/p' | sort --version-sort | tail -1)
-        STABLEVERSION="nextcloud-$NCVERSION"
-        rm /tmp/prerelease.version
+        sleep 1
     else
-        NCREPO="https://download.nextcloud.com/server/prereleases"
-        NCVERSION=$(cat /tmp/prerelease.version)
-        STABLEVERSION="nextcloud-$NCVERSION"
-        rm /tmp/prerelease.version
+        if grep -q beta /tmp/prerelease.version
+        then
+            NCREPO="https://download.nextcloud.com/server/prereleases"
+            NCVERSION=$(curl -s -m 900 $NCREPO/ | sed --silent 's/.*href="nextcloud-\([^"]\+\).zip.asc".*/\1/p' | sort --version-sort | tail -1)
+            STABLEVERSION="nextcloud-$NCVERSION"
+            rm /tmp/prerelease.version
+        else
+            NCREPO="https://download.nextcloud.com/server/prereleases"
+            NCVERSION=$(cat /tmp/prerelease.version)
+            STABLEVERSION="nextcloud-$NCVERSION"
+            rm /tmp/prerelease.version
+        fi
     fi
 fi
 
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -547,7 +547,7 @@ then
     mkdir -p "$VMLOGS"
 fi
 
-CURRENTVERSION_after=$(occ_command_no_check status | grep "versionstring" | awk '{print $3}')
+CURRENTVERSION_after=$(occ_command status | grep "versionstring" | awk '{print $3}')
 if [[ "$NCVERSION" == "$CURRENTVERSION_after" ]] || [ -n "$PRERELEASE_VERSION" ]
 then
 msg_box "Latest version is: $NCVERSION. Current version is: $CURRENTVERSION_after.
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -287,6 +287,20 @@ then
     export NCVERSION
     export STABLEVERSION="nextcloud-$NCVERSION"
     rm -f /tmp/minor.version
+elif [ -f /tmp/prerelease.version ]
+then
+    if grep -q beta /tmp/prerelease.version
+    then
+        NCREPO="https://download.nextcloud.com/server/prereleases"
+        NCVERSION=$(curl -s -m 900 $NCREPO/ | sed --silent 's/.*href="nextcloud-\([^"]\+\).zip.asc".*/\1/p' | sort --version-sort | tail -1)
+        STABLEVERSION="nextcloud-$NCVERSION"
+        rm /tmp/prerelease.version
+    else
+        NCREPO="https://download.nextcloud.com/server/prereleases"
+        NCVERSION=$(cat /tmp/prerelease.version)
+        STABLEVERSION="nextcloud-$NCVERSION"
+        rm /tmp/prerelease.version
+    fi
 fi
 
 # Major versions unsupported
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -290,10 +290,10 @@ then
 elif [ -f /tmp/prerelease.version ]
 then
     PRERELEASE_VERSION=yes
-    rm -f /tmp/prerelease.version
     msg_box "WARNING! You are about to update to a Beta/RC version of Nextcloud.\nThere's no turning back, because it's not possible to downgrade.\n\nPlease only continue if you have made a backup, or took a snapshot."
     if [[ "no" == $(ask_yes_or_no "Do you really want to do this?") ]]
     then
+        rm -f /tmp/prerelease.version
         unset PRERELEASE_VERSION
     else
         if grep -q beta /tmp/prerelease.version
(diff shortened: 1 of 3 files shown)


enoch85:
Aah, I see what you're doing here. Nice!

Makes it easier to test beta for sure. Should add a BIG WARNING though, do it on your own risk, since there's no turning back (you can't downgrade)

szaimen:
> Aah, I see what you're doing here. Nice!
> 
> Makes it easier to test beta for sure. Should add a BIG WARNING though, do it on your own risk, since there's no turning back (you can't downgrade)

Where should this security advice be added?
already in the update.sh script?

szaimen:
and probably this section should get reworked:
https://github.com/nextcloud/vm/blob/master/README.md#i-want-to-test-a-release-candidate-rc-or-beta
Would you do that? is too much documentation for me 😅

static/update.sh
```diff
@@ -21,6 +21,12 @@ mkdir -p "$SCRIPTS"
 if [ "${1}" = "minor" ]
 then
     echo "$((NCMAJOR-1))" > /tmp/minor.version
+elif [ "${1}" = "beta" ]
+then
+    echo "beta" > /tmp/prerelease.version
+elif [ "${1}" != "" ]
```
enoch85, line 27: Add a value like "latestbeta" or something, looking for nothing is dangerous here.

enoch85:
> already in the update.sh script?

Maybe add something like:

```
elif [ -f /tmp/prerelease.version ]
then
    msg_box "WARNING, you are about to update to a beta version of Nextcloud, there's no turning back"
  ask_yes_no "no" "do you reallty want this?"
   then
       sleep 1
    else
    if grep -q beta /tmp/prerelease.version
    .....



   fi
```

Sloppy written, but you get the point.

> is too much documentation for me sweat_smile
 
This made me laugh! Which documentation have you even written? xD

szaimen:
> This made me laugh! Which documentation have you even written? xD

I've tried, if you remember ;)
(the doc you couldn't open first...)

static/update.sh
szaimen: the thing was, that I wanted to enable the people to enter a version like 19.0.0RC1 here...
I think all other cases are catched by the latest beta version...

szaimen:
> > already in the update.sh script?
> 
> Maybe add something like:
> 
> ```
> elif [ -f /tmp/prerelease.version ]
> then
>     msg_box "WARNING, you are about to update to a beta version of Nextcloud, there's no turning back"
>   ask_yes_no "no" "do you reallty want this?"
>    then
>        sleep 1
>     else
>     if grep -q beta /tmp/prerelease.version
>     .....
> 
> 
> 
>    fi
> ```
> 
> Sloppy written, but you get the point.

okay, have tried to come up with a better formulation.

static/update.sh
szaimen: I am not looking for nothing but for everything then nothing...

szaimen: or do you have a better idea how to allow something like 16.0.0RC1?

enoch85: `grep` for RC at least :)

szaimen: is fixed 👍

szaimen:
I would say it is ready for review now.

static/update.sh
```diff
@@ -24,7 +24,7 @@ then
 elif [ "${1}" = "beta" ]
 then
     echo "beta" > /tmp/prerelease.version
-elif [ "${1}" != "" ]
+elif [[ "${1}" == *"RC"* ]]
```
enoch85, line 27: better, thanks! :)

enoch85:
When updating to latest beta:

```
Checking for new Nextcloud version...
/dev/fd/63: line 949: printf: 0beta7: invalid number
```

szaimen:
> When updating to latest beta:
> 
> ```
> Checking for new Nextcloud version...
> /dev/fd/63: line 949: printf: 0beta7: invalid number
> ```

What have you entered? 
beta or e.g. 19.0.0RC1?
I just copied the lines over from the readme file...

enoch85:
> So if it doesn't work, they were probably false before...

Maybe, but now when we have the chance to fix it, maybe we should. It works, but it's not "looking good". The problem is that `print` is looking for numbers, so "beta" or RC" doesn't work.

Also, when running the update it says it failed because it of the above reason.

szaimen:
> Maybe, but now when we have the chance to fix it, maybe we should. It works, but it's not "looking good". The problem is that `print` is looking for numbers, so "beta" or RC" doesn't work.
> 
> Also, when running the update it says it failed because it of the above reason.

So we should fix the print function?

enoch85:
It would be easier to check for `/tmp/prerelease.version` and skip the version checks if it's there.

szaimen:
> It would be easier to check for `/tmp/prerelease.version` and skip the version checks if it's there.

could you point me to the code that has to be addressed?

enoch85:
https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L313

https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L525

szaimen:
https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L525
this line doesn't has to be addressed if I understand it correctly. because it should already work?

or what is the output of 
https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L525
when upgrading to a rc or beta release?

enoch85:
Exactly, remove those checks if running beta or RC.

szaimen:
@enoch85 I think it should work now

nextcloud_update.sh
```diff
@@ -547,7 +547,7 @@ then
     mkdir -p "$VMLOGS"
 fi
 
-CURRENTVERSION_after=$(occ_command_no_check status | grep "versionstring" | awk '{print $3}')
+CURRENTVERSION_after=$(occ_command status | grep "versionstring" | awk '{print $3}')
```
enoch85, line 550: Why this change?

The update script is run in the startup-script (last step) and needs to work at all times. This wouldn't allow to pass if there was an error.

szaimen: oh, I thought it was this way before. Can change it back.

BTW: wouldn't it make sense to use occ_command here: I imagine that this only fails, if occ isn't responding or existing. That would cause it to fail which it should, shouldn't it?

szaimen: btw: it was this way before...

szaimen: Please look here:
https://github.com/nextcloud/vm/blob/78a458cec7da2cf81d1b0e27290b5217a898142d/nextcloud_update.sh#L524

enoch85: OK, let's keep it as is then.

enoch85:
One small issue, when upgrading to a pre-release or beta-release, then when running the update script again, it won't recognize the new versoin since that's using another download repo.

On 19.0.0RC1:

```
Checking for new Nextcloud version...
You already run the latest version! (18.0.4)
```

szaimen:
> One small issue, when upgrading to a pre-release or beta-release, then when running the update script again, it won't recognize the new versoin since that's using another download repo.
> 
> On 19.0.0RC1:
> 
> ```
> Checking for new Nextcloud version...
> You already run the latest version! (18.0.4)
> ```
> 
> This would potentially mean:
> 1 Someone have auto updates turned on
> 2. Wants to try the latest beta
> 3. Upgrades
> 4. Don't think more about it
> 5. A new stable version is out
> 6. Nextcloud crashes during the next auto update because it will try to upgrade to the latest stable which is less than then latest RC

of course that shouldn't happen. I will try to find out whether we could add a check.

nextcloud_update.sh
```diff
@@ -308,15 +331,18 @@ https://shop.hanssonit.se/product/upgrade-between-major-owncloud-nextcloud-versi
...
-else
-    print_text_in_color "$IGreen" "You already run the latest version! ($NCVERSION)"
-    exit 0
+    print_text_in_color "$ICyan" "Checking for new Nextcloud version..."
+    if version_gt "$NCVERSION" "$CURRENTVERSION"
+    then
+        print_text_in_color "$ICyan" "Latest release is: $NCVERSION. Current version is: $CURRENTVERSION."
+        print_text_in_color "$IGreen" "New version available, upgrade continues!"
+    else
+        print_text_in_color "$IGreen" "You already run the latest version! ($CURRENTVERSION)"
```
szaimen, line 343: @enoch85 I changed this report to report the latest installed version.

nextcloud_update.sh
```diff
@@ -308,15 +331,18 @@ https://shop.hanssonit.se/product/upgrade-between-major-owncloud-nextcloud-versi
...
+# Check if new version is larger than current version installed. Skip version check if you want to upgrade to a prerelease.
+if [ -z "$PRERELEASE_VERSION" ]
 then
-    print_text_in_color "$ICyan" "Latest release is: $NCVERSION. Current version is: $CURRENTVERSION."
-    print_text_in_color "$IGreen" "New version available, upgrade continues!"
-else
-    print_text_in_color "$IGreen" "You already run the latest version! ($NCVERSION)"
-    exit 0
+    print_text_in_color "$ICyan" "Checking for new Nextcloud version..."
+    if version_gt "$NCVERSION" "$CURRENTVERSION"
```
szaimen, line 338: I don't think that it would update you to a version below your actual one because of this check.
So as I understand it, it would only update you automatically if a release is added to the stable directory.

szaimen: So eg if only 18.0.4 is found on the stable directory, ncversion would always be less then 19.0.0RC1, aka currentversion and hence not upgrading.
it would only upgrade automatically if e.g. 19.0.1 is released to the stable branch.

enoch85: Problem here is that `19.0.0RC1` isn't a pure number as it contains `RC1` so I don't know what will happen to be honest.

Sure, 19 is higher than 18, but if the check fails, maybe it skips it completely anyway?

nextcloud_update.sh
enoch85: Nice catch!

nextcloud_update.sh
enoch85: the `version` function is made to work with numbers, nothing else. Hence I'm worried it might fail.

szaimen: actually I've tried that for you and although the test instance is now on 19.0.0RC1, currentversion:
`CURRENTVERSION=$(sudo -u www-data php $NCPATH/occ status | grep "versionstring" | awk '{print $3}')`
returns only 19.0.0 on my test instance, which should not make any problem with version_gt.

enoch85: OK then, let's merge this.

PR #1231, opened 2020-05-09
add change_mount option

szaimen:


szaimen:
Have tested the changes to the script already, hance marking as ready for review.

PR #1237, opened 2020-05-10
improve smbmount another time

szaimen:


apps/smbmount.sh
```diff
@@ -123,23 +131,35 @@ do
...
+            if [ -f $SMB_CREDENTIALS/SMB$count ]
+            then
+                check_command rm $SMB_CREDENTIALS/SMB$count
+            fi
             break
         else
-            # Install and enable files_external
-            if ! is_app_enabled files_external
+            # Check if Nextcloud is existing
+            if [ -f $NCPATH/occ ]
```
enoch85, line 141: Maybe add this to all the functions tha use 'occ' instead?

enoch85:
Nice work!

apps/smbmount.sh
szaimen: What do you mean by that?

szaimen:
> Nice work!

Thanks!
It is already tested. I am only waiting for travis if there are some suggestions...

szaimen:
travis is fixed

enoch85:
Is this done?

apps/smbmount.sh
enoch85: add:

```
if [ -f $NCPATH/occ ]
then
    some_occ_command
fi
```

...in each function that use `occ`

szaimen: So shall I add this to the occ_command or occ_command_no_check functions?
Is that your suggestion?

concerning the script: the only places where I use them is already covered by this one check.

szaimen:
> Is this done?

yes, we could only debate, whether we can remove or change those lines now:
https://github.com/nextcloud/vm/blob/bc5a76c0ce3d1b4584e305f5bdc0a9f747291aa7/apps/smbmount.sh#L41-L49

enoch85:
>  remove or change those lines now

Since you now seem to use `.smbcredentials` those lines can be removed. 

`.smbcredentials` are the file that should be secured as there's no important stuff in `fstab`.

szaimen:
> > remove or change those lines now
> 
> Since you now seem to use `.smbcredentials` those lines can be removed.
> 
> `.smbcredentials` are the file that should be secured as there's no important stuff in `fstab`.

okay, then I remove those.
and yes, the files in .smbcredentials are secured.

szaimen:
this is done 👍

apps/smbmount.sh
szaimen: @enoch85 I need an answer here before merging..

PR #1246, opened 2020-05-17
introduce run_script and download_script

szaimen:
Since it is possible that we introduce new subfolders in the future, I think it makes sense to unify the algorithm of run_static_script and download_static_script etc. into one function each. 
We can use that to add more functions like this very easily in the future without having to copy that code every time.

It will be also really useful to fix # 11

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -848,19 +848,22 @@ download_script() {
 # Use it for functions like run_static_script
 run_script() {
     rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
-    if curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS"
+    if download_script "${1}" "${2}"
     then
-        bash "${SCRIPTS}/${2}.sh"
-        rm -f "${SCRIPTS}/${2}.sh"
-    elif curl_to_dir "${!1}" "${2}.php" "$SCRIPTS"
-    then
-        php "${SCRIPTS}/${2}.php"
-        rm -f "${SCRIPTS}/${2}.php"
-    elif curl_to_dir "${!1}" "${2}.py" "$SCRIPTS"
-    then
-        install_if_not python3
-        python3 "${SCRIPTS}/${2}.py"
-        rm -f "${SCRIPTS}/${2}.py"
+        if [ -f "${SCRIPTS}/${2}".sh ]
+        then
+            bash "${SCRIPTS}/${2}.sh"
+            rm -f "${SCRIPTS}/${2}.sh"
+        elif [ -f "${SCRIPTS}/${2}".php ]
+        then
+            php "${SCRIPTS}/${2}.php"
+            rm -f "${SCRIPTS}/${2}.php"
+        elif [ -f "${SCRIPTS}/${2}".py ]
+        then
+            install_if_not python3
+            python3 "${SCRIPTS}/${2}.py"
+            rm -f "${SCRIPTS}/${2}.py"
+        fi
     else
         print_text_in_color "$IRed" "Downloading ${2} failed"
         print_text_in_color "$ICyan" "Script failed to download. Please run: 'sudo curl -sLO ${!1}/${2}.sh|php|py' again."
--- a/lib.sh
+++ b/lib.sh
@@ -871,31 +871,31 @@ run_script() {
 # Initial download of script in ../static
 # call like: download_static_script name_of_script
 download_static_script() {
-download_script STATIC ${1}
+download_script STATIC "${1}"
 }
 
 # Initial download of script in ../lets-encrypt
 # call like: download_le_script name_of_script
 download_le_script() {
-download_script LETS_ENC ${1}
+download_script LETS_ENC "${1}"
 }
 
 # Run any script in ../master
 # call like: run_main_script name_of_script
 run_main_script() {
-run_script GITHUB_REPO ${1}
+run_script GITHUB_REPO "${1}"
 }
 
 # Run any script in ../static
 # call like: run_static_script name_of_script
 run_static_script() {
-run_script STATIC ${1}
+run_script STATIC "${1}"
 }
 
 # Run any script in ../apps
 # call like: run_app_script collabora|nextant|passman|spreedme|contacts|calendar|webmin|previewgenerator
 run_app_script() {
-run_script APP ${1}
+run_script APP "${1}"
 }
 
 version(){
--- a/lib.sh
+++ b/lib.sh
@@ -829,105 +829,73 @@ rm -r "$GPGDIR"
 rm -f releases
 }
 
-# Initial download of script in ../static
-# call like: download_static_script name_of_script
-download_static_script() {
-    # Get ${1} script
-    rm -f "${SCRIPTS}/${1}.sh" "${SCRIPTS}/${1}.php" "${SCRIPTS}/${1}.py"
-    if ! { curl_to_dir "${STATIC}" "${1}.sh" "$SCRIPTS" || curl_to_dir "${STATIC}" "${1}.php" "$SCRIPTS" || curl_to_dir "${STATIC}" "${1}.py" "$SCRIPTS"; }
+# call like: download_script folder_variable name_of_script
+# e.g. download_script APP additional_apps
+# Use it for functions like download_static_script
+download_script() {
+    rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
+    if ! { curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${!1}" "${2}.php" "$SCRIPTS" || curl_to_dir "${!1}" "${2}.py" "$SCRIPTS"; }
     then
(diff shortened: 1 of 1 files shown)


szaimen:
This is ready for review; was tested by me in an earlier state.

enoch85:
Even if I can read code, it would be nice with a description. Especially since this seems like a quite big change. :)

szaimen:
> Even if I can read code, it would be nice with a description. Especially since this seems like a quite big change. :)

I added a description. Hope that explains the motivation behind this PR.

lib.sh
```diff
@@ -829,105 +829,73 @@ rm -r "$GPGDIR"
...
+# call like: run_script folder_variable name_of_script
+# e.g. run_script APP additional_apps
+# Use it for functions like run_static_script
+run_script() {
+    rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
+    if curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS"
     then
-        print_text_in_color "$IRed" "{$1} failed to download. Please run: 'sudo curl -sLO ${STATIC}/${1}.sh|.php|.py' again."
-        print_text_in_color "$ICyan" "If you get this error when running the nextcloud-startup-script then just re-run it with:"
-        print_text_in_color "$ICyan" "'sudo bash $SCRIPTS/nextcloud-startup-script.sh' and all the scripts will be downloaded again"
```
enoch85, line 855: Please don't remove this text.

lib.sh
```diff
@@ -829,105 +829,73 @@ rm -r "$GPGDIR"
...
-download_le_script() {
-    # Get ${1} script
-    rm -f "${SCRIPTS}/${1}.sh" "${SCRIPTS}/${1}.php" "${SCRIPTS}/${1}.py"
-    if ! { curl_to_dir "${LETS_ENC}" "${1}.sh" "$SCRIPTS" || curl_to_dir "${LETS_ENC}" "${1}.php" "$SCRIPTS" || curl_to_dir "${LETS_ENC}" "${1}.py" "$SCRIPTS"; }
+# call like: run_script folder_variable name_of_script
+# e.g. run_script APP additional_apps
+# Use it for functions like run_static_script
+run_script() {
+    rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
+    if curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS"
```
enoch85, line 851: Why do you use `!` here: `curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS"`

lib.sh
enoch85: Sorry, remove it. It makes sense to not have it there.

lib.sh
enoch85: I think I get it, you need to do that because we can't use `$1` here due to the varaibles needed like APP, STATIC, LETS_ENC and so on.

enoch85:
The cleaner solution would be to actually replace all the old function calls. More work, but much cleaner since there wouldn't be any mixed code. 

Yeah, backwards compatible ofc. But I think we should do this in like 3 versions or so and only use this new function.

enoch85:
> Yeah, backwards compatible ofc. But I think we should do this in like 3 versions or so and only use this new function.

So for backwards compatibility, keep the old functions in the lib.sh file for now, but change all the current functions to this new one. Please also add the new catalogue vars to this PR.

Thanks!

lib.sh
szaimen: yes, that's right, couldn't explain it better 😅

szaimen: a different way would be printing the variabes out directly inside the run_static_script functions but I like it more this way

lib.sh
szaimen: Yeah, github mixed the added and remove lines up so that it is very confusing now.

szaimen:
> > Yeah, backwards compatible ofc. But I think we should do this in like 3 versions or so and only use this new function.
> 
> So for backwards compatibility, keep the old functions in the lib.sh file for now, but change all the current functions to this new one. Please also add the new catalogue vars to this PR.
> 
> Thanks!

actually I like it this way more like I implemented it.

Because this way we can later on simply add the following code to fix #11


```
run_script() {
    rm -f "${SCRIPTS}/${3}.sh" "${SCRIPTS}/${3}.php" "${SCRIPTS}/${3}.py"
    if [ -f $SCRIPTS/${2}/${3}.sh ]
    then
        cp $SCRIPTS/${2}/${3}.sh $SCRIPTS
        bash "${SCRIPTS}/${3}.sh"
        rm -f "${SCRIPTS}/${3}.sh" 
    elif [ -f $SCRIPTS/${2}/${3}.php ]
    then
        cp $SCRIPTS/${2}/${3}.php $SCRIPTS
        php "${SCRIPTS}/${3}.php"
        rm -f "${SCRIPTS}/${3}.php"
    elif [ -f $SCRIPTS/${2}/${3}.py ]
    then
        install_if_not python3
        cp $SCRIPTS/${2}/${3}.py $SCRIPTS
        python3 "${SCRIPTS}/${3}.py"
        rm -f "${SCRIPTS}/${3}.py" 
    elif curl_to_dir "${!1}" "${3}.sh" "$SCRIPTS"
    then
        bash "${SCRIPTS}/${3}.sh"
        rm -f "${SCRIPTS}/${3}.sh"
    elif curl_to_dir "${!1}" "${3}.php" "$SCRIPTS"
    then
        php "${SCRIPTS}/${3}.php"
        rm -f "${SCRIPTS}/${3}.php"
    elif curl_to_dir "${!1}" "${3}.py" "$SCRIPTS"
    then
        install_if_not python3
        python3 "${SCRIPTS}/${3}.py"
        rm -f "${SCRIPTS}/${3}.py"
    else
        print_text_in_color "$IRed" "Downloading ${3} failed"
        print_text_in_color "$ICyan" "Script failed to download. Please run: 'sudo curl -sLO ${!1}/${3}.sh|php|py' again."
        exit 1
    fi
}

run_main_script() {
run_script GITHUB_REPO main "${1}"
}

download_script() {
    rm -f "${SCRIPTS}/${3}.sh" "${SCRIPTS}/${3}.php" "${SCRIPTS}/${3}.py"
    if [ -f $SCRIPTS/${2}/${3}.sh ]
    then
        cp $SCRIPTS/${2}/${3}.sh $SCRIPTS
    elif [ -f $SCRIPTS/${2}/${3}.php ]
    then
        cp $SCRIPTS/${2}/${3}.php $SCRIPTS
    elif [ -f $SCRIPTS/${2}/${3}.py ]
    then
        cp $SCRIPTS/${2}/${3}.py $SCRIPTS
    elif ! { curl_to_dir "${!1}" "${3}.sh" "$SCRIPTS" || curl_to_dir "${!1}" "${3}.php" "$SCRIPTS" || curl_to_dir "${!1}" "${3}.py" "$SCRIPTS"; }
    then
        print_text_in_color "$IRed" "{$3} failed to download. Please run: 'sudo curl -sLO ${!1}/${3}.sh|.php|.py' again."
        exit 1
    fi
}

download_static_script() {
download_script STATIC static "${1}"
}
```

This and other changes won't be possible if we implemented it like you suggest to only use the new function.
Also it is better to read imo.

szaimen:
I've actually thought the same you suggest to enable only the new function but I think it is better readable like this and doesn't overcomplicate things if we have this helper function...

szaimen:
Imagin this: 
if we would implement everywhere 
`run_script STATIC some_script`
in order to fix #11 the new way, we would have to change it again which would break backwards compatibility again and would resolve in having to change it everywhere again instead of just one time in the library which is really the best place to change it imo.

So for backwardscompatibility and ease of use and programming expense, I vote for the current solution.

lib.sh
enoch85: OK, great. But from where did you get that idea? I would like the background of it. Is it consider stable to do like that, or could it be mixed up with some other stuff?

Actually, I would prefer using the real variable. Then  we could check with `if [ -n $APP ]; then...` to make it more failproof.

szaimen: it is called "indirect expansion"
here you can find an explanation:
http://ahmed.amayem.com/bash-indirect-expansion-exploration/#ipt_kb_toc_1527_3

szaimen: the other solution which is possible is this:
```
download_script() {
    if ! { curl_to_dir "${1}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${1}" "${2}.php" "$SCRIPTS" || curl_to_dir "${1}" "${2}.py" "$SCRIPTS"; }
    then
        print_text_in_color "$IRed" "{$2} failed to download. Please run: 'sudo curl -sLO ${1}/${2}.sh|.php|.py' again."
        exit 1
    fi
}

download_static_script() {
download_script "$STATIC" "${1}"
}
```
vs
```
download_script() {
    if ! { curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${!1}" "${2}.php" "$SCRIPTS" || curl_to_dir "${!1}" "${2}.py" "$SCRIPTS"; }
    then
        exit 1
    fi
}
download_static_script() {
download_script STATIC "${1}"
}
```
Which one do you like more?

szaimen: this is another explanation:
https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

enoch85:
I'm fine with this for now. Let's make improvements later.

PR #1247, opened 2020-05-17
update lib.sh to change variables to functions

szaimen:


szaimen:
I've already tested that the concept works, which is the case. Hence marking as ready for review.

enoch85:
The reason why we use the current way is that we do not want to run the variables every time the `lib.sh` is fetched. 

I haven't tested this, so I can't say, but are you 100% sure that the variables aren't set even if it's in a function?

lib.sh
```diff
@@ -969,11 +995,9 @@ any_key() {
 }
 
 lowest_compatible_nc() {
-if [ -z "$NC_UPDATE" ]
+if [ -z "$CURRENTVERSION" ]
```
enoch85, line 998: Why this change?

szaimen: because I want to get rid of the NC_UPDATE variable sometime in the future.
Currentversion is only set, if the new functions was called.

szaimen:
> The reason why we use the current way is that we do not want to run the variables every time the `lib.sh` is fetched.
> 
> I haven't tested this, so I can't say, but are you 100% sure that the variables aren't set even if it's in a function?

yes, that is the case. I have tested it by running bash -x on a script (and see which variables get called during the pulling of the new library)

lib.sh
szaimen: Btw: now that i've changed all variables, NC_UPDATE isn't anymore called regularly

enoch85:
I tested this yesterday, just to see if it works to use functions instead, and it does.

But, some variables are nested into different scripts, and I'm very careful here, since this is the way we done it for the past years. Double, tripple, and quadrouple check all the variables that are changed here. `grep -r $VARIABLE` for all the variables run when this is set.

I know that `download_verify_nextcloud_stable` use `$NCVERSION`, which is used in `nc_update`, but is that really set? I'm also not 100% sure about checking `$CURRENTVERSION` 

And third, now we no longer have a way to check if the needed functions are actually loaded, which ww check in --------------------------------^.

Thanks!

static/main_menu.sh
```diff
@@ -27,19 +27,19 @@ case "$choice" in
...
         then
-            run_static_script nextcloud_configuration
+            run_script STATIC configuration
         fi
     ;;
     "Server Configuration")
         if network_ok
         then
-            run_static_script server_configuration
+            run_script STATIC server_configuration
```
szaimen, line 42: Can we please revert all changes in this commit?
As I said it will complicate things to fix #11 by a lot!

szaimen:
> But, some variables are nested into different scripts, and I'm very careful here, since this is the way we done it for the past years. Double, tripple, and quadrouple check all the variables that are changed here. `grep -r $VARIABLE` for all the variables run when this is set.

I don't get what you mean. Have you seen, that we have actually a backup for the legacy way of calling them to remain backwards compatibility and in the case that we have forgotten one?
looke e.g. here: https://github.com/nextcloud/vm/pull/1247/files#diff-68f7bc5e839823a7568557cf9e87f4e2R41

> I know that `download_verify_nextcloud_stable` use `$NCVERSION`, which is used in `nc_update`, but is that really set? I'm also not 100% sure about checking `$CURRENTVERSION`

you can just test it. In my testing it works without any flaw.

> And third, now we no longer have a way to check if the needed functions are actually loaded, which ww check in --------------------------------^.

What are you talking about? you can just test, if the variables you want to retrieve are actually not zero. E.g. [ -n "$CURRENTVERSION" ] or e.g. [ -n "$NCVERSION" ]

static/main_menu.sh
szaimen: I've reverted that.

szaimen:
@enoch85 I would like to get this PR in before #1256. Do you think this is doable?

enoch85:
> where have we anywhere tested if functions get loaded?

One example, I'm not sure if there's more: https://github.com/nextcloud/vm/blob/master/lib.sh#L790

Just because what you tested works, it doesn't mean that everything works. Just sayin. Happened to me several times. :)

enoch85:
> I don't get what you mean. Have you seen, that we have actually a backup for the legacy way of calling them to remain backwards compatibility and in the case that we have forgotten one?
looke e.g. here: https://github.com/nextcloud/vm/pull/1247/files#diff-68f7bc5e839823a7568557cf9e87f4e2R41

Ok, nice one. Question is - how long should we keep it? I don't want to clutter the scripts with different approaches. We use one standard, that's it.

szaimen:
> > I don't get what you mean. Have you seen, that we have actually a backup for the legacy way of calling them to remain backwards compatibility and in the case that we have forgotten one?
> > looke e.g. here: https://github.com/nextcloud/vm/pull/1247/files#diff-68f7bc5e839823a7568557cf9e87f4e2R41
> 
> Ok, nice one. Question is - how long should we keep it? I don't want to clutter the scripts with different approaches. We use one standard, that's it.

The thing is, that we 

> > where have we anywhere tested if functions get loaded?
> 
> One example, I'm not sure if there's more: https://github.com/nextcloud/vm/blob/master/lib.sh#L790
> 
> Just because what you tested works, it doesn't mean that everything works. Just sayin. Happened to me several times. :)

Okay, but it seems like this is network related, because you just check once in the library and not looping over it. because ncversion will only be set if the website can get reached.

Don't want to say that you don't have a point here but if there was no check before, I won't see why we need to add additional checks, because the functions will work as reliable as the current NC_UPDATE=1 way of getting the needed variables.

szaimen:
> > I don't get what you mean. Have you seen, that we have actually a backup for the legacy way of calling them to remain backwards compatibility and in the case that we have forgotten one?
> > looke e.g. here: https://github.com/nextcloud/vm/pull/1247/files#diff-68f7bc5e839823a7568557cf9e87f4e2R41
> 
> Ok, nice one. Question is - how long should we keep it? I don't want to clutter the scripts with different approaches. We use one standard, that's it.

What about until the next Ubuntu 22.04 release? I think that is always a good date to get rid of legacy components - also it is nearly 2 years ahead. But that is what you have to decide because many things in the old VMs won't work without that backup.

We could add a comment that it should get removed for Ubuntu 22.04?

enoch85:
> What about until the next Ubuntu 22.04 release?

In general I think that we should keep some backwards compatibility for at least one OS before. So in 20.04 we ditched 16.04. In 22.04 we should ditch 18.04. It's nothing I have written in stone, but I think that's a good approach. It also depends on how much Ubuntu changes ofc. 16.04 to 18.04 was a huge change. 18.04 to 22.04 - not so huge. 

But :point_up: is regarding OS versions, I think this is more a matter of keeping backwards compatibility for a shorter period of time, just to make sure that old VMs (like max 6 months) should run without an issue. 

I mean the most critical thing here is those who haven't yet deployed their VM. We can expect that they did so within 6 months, right? The update script always uses the latest `lib.sh` anyway, so that won't be an issue, and people that want to use the latest script for apps and such, it's also fixed with `menu.sh` already as that will download the latest script in question, already containing the latest `lib.sh`

What do you think?

enoch85:
> Okay, but it seems like this is network related, because you just check once in the library and not looping over it. because ncversion will only be set if the website can get reached.

Don't want to say that you don't have a point here but if there was no check before, I won't see why we need to add additional checks, because the functions will work as reliable as the current NC_UPDATE=1 way of getting the needed variables.

The whole reason for that change (not using NC_UPDATE for the whole script, and put it in the function instead) is that it takes time to load the variables and it hammers Nextclouds servers more than we need to. IIRC, that's the main reason for it.

But, maybe it's as easy as just check for $CURRENTVERSION as you did, and if if doesn't exist, just run `nc_update`?

What's making me most uncertain here is that you say that you've tested everything, but then you missed this... In my head "what else did he miss"? But at the same time, that's why we have these discussions. :+1:  Even if I'm thorough as ***, I still miss things as well, we are all humans (and all that) :)

szaimen:
> > Okay, but it seems like this is network related, because you just check once in the library and not looping over it. because ncversion will only be set if the website can get reached.
> 
> Don't want to say that you don't have a point here but if there was no check before, I won't see why we need to add additional checks, because the functions will work as reliable as the current NC_UPDATE=1 way of getting the needed variables.
> 
> The whole reason for that change (not using NC_UPDATE for the whole script, and put it in the function instead) is that it takes time to load the variables and it hammers Nextclouds servers more than we need to. IIRC, that's the main reason for it.

yes and also we have less variables in every script and it is cleaner to be able to run the functions every time that we need them. Also They wouldn't get called during the script start which would speed up the first start...

> What's making me most uncertain here is that you say that you've tested everything, but then you missed this... In my head "what else did he miss"? But at the same time, that's why we have these discussions. 👍 Even if I'm thorough as ***, I still miss things as well, we are all humans (and all that) :)

Actually, I didn't missed it because else I had changed it (if it would have been some of the changed variables). I only haven't remembered now.
But yes, I am a human, too and I can miss things. 
That is one of the reason why I have brought in the legacy components like https://github.com/nextcloud/vm/pull/1247/files#diff-68f7bc5e839823a7568557cf9e87f4e2R41

> But, maybe it's as easy as just check for $CURRENTVERSION as you did, and if if doesn't exist, just run `nc_update`?

Yes, that is the whole idea 👍

szaimen:
> > What about until the next Ubuntu 22.04 release?
> 
> In general I think that we should keep some backwards compatibility for at least one OS before. So in 20.04 we ditched 16.04. In 22.04 we should ditch 18.04. It's nothing I have written in stone, but I think that's a good approach. It also depends on how much Ubuntu changes ofc. 16.04 to 18.04 was a huge change. 18.04 to 22.04 - not so huge.
> 
> But ☝️ is regarding OS versions, I think this is more a matter of keeping backwards compatibility for a shorter period of time, just to make sure that old VMs (like max 6 months) should run without an issue.
> 
> I mean the most critical thing here is those who haven't yet deployed their VM. We can expect that they did so within 6 months, right? The update script always uses the latest `lib.sh` anyway, so that won't be an issue, and people that want to use the latest script for apps and such, it's also fixed with `menu.sh` already as that will download the latest script in question, already containing the latest `lib.sh`
> 
> What do you think?

I mean you earn money by selling them so it is definitely your choice.
But we need to make sure that components that are needed after the inital startup still work  which actually also fits to nextcloud_configuration, additional_apps etc.
And as a customer, I would definitely expect to be able to use it again for a new or additional instance longer then 6 months - but as I said that is your choice.

BTW: this will also be solved, if we finally fix #11 because you then will have the choice to use all scripts offline without downloading them during the first startup, which will make sure that you can use the bought VM to spin up a new VM endless. (f you choose the right option during the startup script of course ;)

szaimen:
What about merging this? Or is there still something to discuss?

enoch85:
> Yes, that is the whole idea +1

Great, then please look through the code once more and see if you can find more occurrences.

> What about merging this? 

No, there are still bugs here, which I mentioned but that isn't fixed.

enoch85:
+ I don't want a failsafe here. Either it should all work, or we should find were it doesn't and fix t.

enoch85:
Here's an example bug:

```
download_verify_nextcloud_stable() {
# Check the current Nextcloud version
while [ -z "$NCVERSION" ]
do
    print_text_in_color "$ICyan" "Fetching the latest Nextcloud version..."
    nc_update
    print_text_in_color "$IGreen" "$NCVERSION"
done
```

Gives:

```
+ '[' -z '' ']'
+ print_text_in_color '\e[0;96m' 'Fetching the latest Nextcloud version...'
+ printf '%b%s%b\n' '\e[0;96m' 'Fetching the latest Nextcloud version...' '\e[0m'
Fetching the latest Nextcloud version...
+ nc_update
++ sudo -u www-data php /var/www/nextcloud/occ status
++ grep versionstring
++ awk '{print $3}'
sudo: php: command not found <<<<<<<<<<<<<<--------------------------------------------
+ CURRENTVERSION=
++ curl -s -m 900 https://download.nextcloud.com/server/releases/
++ sed --silent 's/.*href="nextcloud-\([^"]\+\).zip.asc".*/\1/p'
++ sort --version-sort
++ tail -1
+ NCVERSION=18.0.4
+ STABLEVERSION=nextcloud-18.0.4
+ NCMAJOR=18
+ NCBAD=16
+ print_text_in_color '\e[0;92m' 18.0.4
+ printf '%b%s%b\n' '\e[0;92m' 18.0.4 '\e[0m'
18.0.4
+ '[' -z 18.0.4 ']'
```

The current way does only fetch `NCVERSION`, not everything...

szaimen:
> > Yes, that is the whole idea +1
> 
> Great, then please look through the code once more and see if you can find more occurrences.

I've changed download_verify_nextcloud_stable now and I've looked looked for all "[ -z" and "[ -n" and "=1" occurences and there is no old colde that still needs to be fixed, imo (despite of the script in the "old" directory); also the code in download_verify_nextcloud_stable wasn't badly needed to get fixed: it would have worked this way.

PR #1251, opened 2020-05-17
Update format-chosen.sh

omnidelic:
Add Intel NUC to SYSVENDOR test to correctly identify 2nd disk during install.


omnidelic:
Not sure if SYSNAME should be assigned to "machines" instead?

Haxy89:
I think static/format-sdb.sh should be updated too

omnidelic:
You're right, I'll add that to the PR :)

enoch85:
OK, code wise it looks good. Im not sure that all "Intel Client Systems" are Intel NUCs, but let's merge and see if we get any bug report about it in the future. Probably not. :)

Issue #1245, opened 2020-05-13
Add more folders - organize


enoch85:
Maybe we should create more specific folders like:

- Configurations scripts
- Really static scripts (never changes)
- Network related
- Etc, etc...

The add more functions like:
- `run_configuration_script`
- `run_network_script`

cc @szaimen

szaimen:
hm yes, the only problem is, that everything relies on the current sort-order. So if somebody is trying to start over with an older VM, it could happen, that it will not work, if we move some scrips to another folder.

e.g. the nextcloud_configuration script has to be in the static folder because otherwise it will not work for anyone that uses an older VM

What could be a solution to solve that problem?

I could only imagine to put a run_configuration_script into the old nextcloud_configuration which would then download the new script but it would not be really neat because we couldn't delete the nextcloud_configuration script inside the static folder...

BTW: what about new functions like:
```
# call like: run_script folder_variable name_of_script
# e.g. run_script APP additional_apps
run_script() {
    rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
    if curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS"
    then
        bash "${SCRIPTS}/${2}.sh"
        rm -f "${SCRIPTS}/${2}.sh"
    elif curl_to_dir "${!1}" "${2}.php" "$SCRIPTS"
    then
        php "${SCRIPTS}/${2}.php"
        rm -f "${SCRIPTS}/${2}.php"
    elif curl_to_dir "${!1}" "${2}.py" "$SCRIPTS"
    then
        install_if_not python3
        python3 "${SCRIPTS}/${2}.py"
        rm -f "${SCRIPTS}/${2}.py"
    else
        print_text_in_color "$IRed" "Downloading ${2} failed"
        print_text_in_color "$ICyan" "Script failed to download. Please run: 'sudo curl -sLO ${STATIC}/${1}.sh|php|py' again."
        exit 1
    fi
}

# call like: download_script folder_variable name_of_script
# e.g. download_script APP additional_apps
download_script() {
    rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
[...]

Issue #1116, opened 2020-02-28
What is needed to fix #11


szaimen:
So here a short list what is probably needed to fix https://github.com/nextcloud/vm/issues/11 I will edit this list, if more comes to my mind

- [x]  1. edit download_static_script(), download_le_script(), run_main_script(), run_static_script(), run_app_script() in order to use the local scripts if existing at the correct place #1123
- [x] 2. edit all scripts in order to use the local lib.sh if existing at the correct place #1124
- [x] 3. provide an option to choose during the startup script if you want to use the local files or online files (Security vs bugfixes) and remove all local files (or at least most of them) when choosing security #1127
- [ ] 4. create a script that downloads the release file with the code from the github repository, compares it again the gpg key and extracts it to the right directory #1162
There are still ToDos missing. Advices: https://github.com/nextcloud/vm/pull/1126#issuecomment-593889476 downwards.
- [x] 5. edit all other scripts to use local files or inform the user if not possible; grep for curl: (nextcloud_install_prouduction, bitwarden, adminer, netdata, webmin, install_docker) #1125
- [ ] 6. sign and release the code

szaimen:
@enoch85 please check out this list to get an idea what all PRs doing together
all are made so that they work without each other. But only of all are merged they will fix #11.

szaimen:
I hope that all PRs have a fitting size so that they can get reviewed.

szaimen:
Btw: until #1120 isn't merged AND used to make the VMs, everything will work as it currently is. Also if all other PRs are merged. 

So I would suggest merging #1120 at the last one. Then when using it, the true magic will happen ;)

szaimen:
@enoch85 did you have time to look into this or the PRs a bit? :)

enoch85:
Sorry, no time yet. :/

enoch85:
OK, so I've merged all your PRs into the testing branch. 

Running the installer now. Don't have so much time to test, but at least I want to see how it looks.

enoch85:
OK, so first test done. No errors, but the scripts doesn't work as intended. 

Please test for yourself and improve. 

One example is [here](https://github.com/nextcloud/vm/blob/testing/nextcloud-startup-script.sh#L273-L315), where you first ask if you want to run from Github or locally, then download scripts from Github either way.

Also the branch inside lib.sh wasn't set so files never downloaded as they should.

szaimen:
Actually I am sure that already everything works.

[...]

PR #1256, opened 2020-05-22
update the way scripts are run and downloaded

enoch85:
@szaimen The setup script is only used in the released versions, so we might as well change the function.

Also run the right scripts (no need to download the apps, configuration scripts again which apps.sh and configuration does).

This is just a start, let's change all of them in the same PR, just don't have time now.

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -192,7 +192,7 @@ Please create a user with sudo permissions and the run this command:
 sudo -u [user-with-sudo-permissions] sudo bash /var/scripts/nextcloud-startup-script.sh
 
 We will do this for you when you hit OK."
-       download_static_script adduser
+       download_script STATIC adduser
        bash $SCRIPTS/adduser.sh "$SCRIPTS/nextcloud-startup-script.sh"
        rm $SCRIPTS/adduser.sh
        else
@@ -236,26 +236,26 @@ echo
 print_text_in_color "$ICyan" "Getting scripts from GitHub to be able to run the first setup..."
 # Scripts in static (.sh, .php, .py)
 download_le_script activate-tls
-download_static_script temporary-fix
-download_static_script update
-download_static_script trusted
-download_static_script test_connection
-download_static_script setup_secure_permissions_nextcloud
-download_static_script change_db_pass
-download_static_script nextcloud
-download_static_script update-config
-download_static_script menu
-download_static_script server_configuration
-download_static_script nextcloud_configuration
+download_script STATIC temporary-fix
+download_script STATIC update
+download_script STATIC trusted
+download_script STATIC test_connection
+download_script STATIC setup_secure_permissions_nextcloud
+download_script STATIC change_db_pass
+download_script STATIC nextcloud
+download_script STATIC update-config
+download_script STATIC menu
+download_script STATIC server_configuration
+download_script STATIC nextcloud_configuration
 download_app_script additional_apps
 
 if home_sme_server
 then
-    download_static_script nhss_index
+    download_script STATIC nhss_index
     mv $SCRIPTS/nhss_index.php $HTML/index.php && rm -f $HTML/html/index.html
     chmod 750 $HTML/index.php && chown www-data:www-data $HTML/index.php
 else
-    download_static_script index
+    download_script STATIC index
     mv $SCRIPTS/index.php $HTML/index.php && rm -f $HTML/html/index.html
     chmod 750 $HTML/index.php && chown www-data:www-data $HTML/index.php
 fi
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -236,7 +236,6 @@ echo
 print_text_in_color "$ICyan" "Getting scripts from GitHub to be able to run the first setup..."
 # Scripts in static (.sh, .php, .py)
 download_le_script activate-tls
-download_static_script update-config
 download_static_script temporary-fix
 download_static_script update
 download_static_script trusted
@@ -244,6 +243,7 @@ download_static_script test_connection
 download_static_script setup_secure_permissions_nextcloud
 download_static_script change_db_pass
 download_static_script nextcloud
+download_static_script update-config
 download_static_script menu
 download_static_script server_configuration
 download_static_script nextcloud_configuration
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -235,7 +235,7 @@ touch "$SCRIPTS/you-can-not-run-the-startup-script-several-times"
 echo
 print_text_in_color "$ICyan" "Getting scripts from GitHub to be able to run the first setup..."
 # Scripts in static (.sh, .php, .py)
-download_le_script activate-tls
+download_script LETS_ENC activate-tls
 download_script STATIC temporary-fix
 download_script STATIC update
 download_script STATIC trusted
@@ -247,7 +247,7 @@ download_script STATIC update-config
 download_script STATIC menu
 download_script STATIC server_configuration
 download_script STATIC nextcloud_configuration
-download_app_script additional_apps
+download_script APP additional_apps
 
 if home_sme_server
 then
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -515,7 +515,7 @@ apt autoclean
 
 # Set trusted domain in config.php
 bash $SCRIPTS/trusted.sh
-rm  -f $SCRIPTS/update.sh
+rm  -f $SCRIPTS/trusted.sh
 
 # Success!
 msg_box "Congratulations! You have successfully installed Nextcloud!
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -235,19 +235,19 @@ touch "$SCRIPTS/you-can-not-run-the-startup-script-several-times"
 echo
 print_text_in_color "$ICyan" "Getting scripts from GitHub to be able to run the first setup..."
 # Scripts in static (.sh, .php, .py)
-download_script LETS_ENC activate-tls
-download_script STATIC update-config
-download_script STATIC temporary-fix
-download_script STATIC update
-download_script STATIC trusted
-download_script STATIC test_connection
-download_script STATIC setup_secure_permissions_nextcloud
-download_script STATIC change_db_pass
-download_script STATIC nextcloud
-download_script STATIC menu
-download_script STATIC server_configuration
-download_script STATIC nextcloud_configuration
-download_script APP additional_apps
+download_le_script activate-tls
+download_static_script update-config
+download_static_script temporary-fix
+download_static_script update
+download_static_script trusted
+download_static_script test_connection
+download_static_script setup_secure_permissions_nextcloud
+download_static_script change_db_pass
+download_static_script nextcloud
+download_static_script menu
+download_static_script server_configuration
+download_static_script nextcloud_configuration
+download_app_script additional_apps
 
 if home_sme_server
 then
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -171,7 +171,7 @@ then
 fi
 
 # Set locales
-run_static_script locales
+run_script STATIC locales
 
 # Nextcloud 18 is required
 lowest_compatible_nc 18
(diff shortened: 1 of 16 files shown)


enoch85:
@szaimen Feel free to push to this branch.

nextcloud-startup-script.sh
```diff
@@ -235,18 +235,20 @@ touch "$SCRIPTS/you-can-not-run-the-startup-script-several-times"
...
+download_script STATIC temporary-fix
+download_script STATIC update
+download_script STATIC trusted
+download_script STATIC test_connection
+download_script STATIC setup_secure_permissions_nextcloud
+download_script STATIC change_db_pass
+download_script STATIC nextcloud
+download_script STATIC menu
+download_script STATIC server_configuration
+download_script STATIC nextcloud_configuration
```
szaimen, line 249: I am against this change because we will break backwards compatibility if we change the order of the variables like I have planned to do.

Why do you want this so madly?

I would only do this after #11 is fixed because it will probably change to be able to fix it...

As additionally the moving of files to another folder will complicate things in the future if we do it this way...

enoch85: This follows the rest of the structure and idea with the startup script - all scripts should be made available to run directly with bash. Your "get latest script" script actually removes the file then re-downloads it again. 

The idea here is to not having to download things **again** if something should happen to internet while while the first setup is run.

Are you following on what I mean?

Another option would be to do it your way and just trust that once out connectivity check is done (first thing that happens when the startup script is run) execute and download whatever as many times as we want.

szaimen: No, that wasn't what I meant.
I would just revert changing to the new function "download_script" because it has the potential to break backwards compatibility in the future.

szaimen: Actually doing what you have proposed makes sense but not with the new function.
I would do it this way:
```
download_le_script activate-tls
download_static_script update-config
download_static_script temporary-fix
download_static_script update
download_static_script trusted
download_static_script test_connection
download_static_script setup_secure_permissions_nextcloud
download_static_script change_db_pass
download_static_script nextcloud
download_static_script menu
download_static_script server_configuration
download_static_script nextcloud_configuration
```

szaimen: I've done that now.

enoch85:
> backwards compatibility

There's no need for that here. The lib file is imported on each run, and you haven't even started that PR yet (afaik) + the startup script is only delivered with the released VMs - so what will actually break here, please explain.

szaimen:
> > backwards compatibility
> 
> There's no need for that here. The lib file is imported on each run, and you haven't even started that PR yet (afaik) + the startup script is only delivered with the released VMs - so what will actually break here, please explain.

I will try to do that:
1. it will complicate things by a lot if we want to fix #11 becausee I have to edit every script again if I change an algorithm like I am planning to do instead of just edit it in the library
this would also make this change probably not possiblebecause I want to add an additional foldername which would sit at the wrong position
please have a look at: https://github.com/nextcloud/vm/issues/1116#issuecomment-629559468
2. Doing magic to move some script in the future to other folders will not be that easy possible, will make it much more difficult https://github.com/nextcloud/vm/issues/1245#issuecomment-631060610
3. Many more reasons, like much more places to edit and the potential to make a mistake, 
4. the old variant is better to read, and much more I currently not remember.

szaimen:
> you haven't even started that PR yet (afaik)

That is not really true, I have already written down some ideas: 
https://github.com/nextcloud/vm/issues/1116#issuecomment-629559468
And that was acutally the next thing I will do after #1247 is merged
Because everything is prepared then to implement this feature very clean.

szaimen:
@enoch85 if you really want to change everything to the new function, then please after #11 is fixed.
Maybe also #1245 should get fixed before we move everything to the new function.

That would be the next thing I would work on after #11 is fixed

enoch85:
>  I have to edit every script again

Yes, maybe so. But I prefer to keep it clean. If I would think the same as you when starting this, we should have code from 2015 in this repo (we have, but you get the point) - keep it clean please. :)

Maybe I'm tired now (I am!) but why wouldn't your fix in #11 work with this change?

2: I'm sorry, but I don't follow you here. Read the comment, but why look for those scripts at all? It can be downloaded with the function anyway, right?

> Many more reasons, like much more places to edit and the potential to make a mistake,

That's why it's important to test all changes, like I talked about earlier. When we did the refactor of this repo and moved all `$vars` and `functions` to `lib.sh` we spent like 3 weeks, night and day. It worked, but I tested every line of the code.... I don't know how you do your testing, but just running a script with exit code 0 doesn't mean that it works for script / fcuntion number 10 or 511. :D

>  if you really want to change everything to the new function, then please after #11 is fixed.

I think it's better to do the other way around. First fix this, then fix #11 

What's happening now is:
1. You make a change to a function which is great!
2. You add backwards compatibility and still uses the old way of doing it just to plan for a legacy fix (yes, legacy since you plan to use old functions in it)

Wouldn't it be better to:
1. Fix all the scripts to use the new way of doing things
2. Use backwards compatibility just to be sure old VMs will work for a while (I'm talking about released VMs, not already in use). The `lib.sh` gets updated anyway, so everyone will get the new function in all versions of the VM, except maybe for the startup script if we change to much. Hence the need for backwards compatibility for a few versions which we then remove.
3. When everything is in place and we have clean code that we decided to use for a while, when fix #11 to be future proof, without the need for legacy stuff.

Do you agree?

szaimen:
Maybe I have to say it differently:
if we do this change, we have many more variables in all scripts, which you obviously tried to solve many years ago by calling e.g. NC_UPDATE=1
If we make this change, we have much more variables in all scripts, that we have to maintain instead of just having them in one place in the library which is the way to go, imo.

So the idea that I am trying to follow is: have as many as possible variables in the library and get them directly from the library by calling functions.

As an example: `run_script Static whatever`
would add a variable "STATIC" to each script which I try to avoid because if we change the name of the variable, or make some other change later on, we have to change it in many places and not in one place in the library.

That is also what I try to solve here: #1256
Having less variables in the scripts and calling functions to get the needed variables.
Because functions are great in reusing code - as you told me a few months ago ;-)

BTW: what I menat with incompatibility is:
if we have in the startup script e.g.
`download_script STATIC server_configuration`
Then changing it to `download_script STATIC static server_configuration` in the library like I plan to do to fix #11 would break the compatibility because the script provides the old function in the 2nd place but it has to be at the 3rd place now. Which would complicate things by a lot.
Also does that new one look good to you if it is in every script? Probably not and that is why the "legacy" functions are better. 

Does that make sense to you?

enoch85:
>  we have many more variables in all scripts, 

How come, am I missing something here? We actually have less variables since we remove legacy stuff. We don't need to have legacy in all scripts, especially for this, it's "just a function" which gets updated on every lib.sh run anyway.

Where / what would we need to add if we do like I intend?

> which you obviously tried to solve many years ago by calling e.g. NC_UPDATE=1

We call for example NC_UPDATE to not run in in every scrip because those variables that are only fetched when set causes bugs if they are run when they are run in all scripts, or they cause the curl command to take much longer time.

> So the idea that I am trying to follow is: have as many as possible variables in the library and get them directly from the library by calling functions or directly if they aren't expensive.

Agreed.

> As an example: run_script Static whatever
would add a variable "STATIC" to each script which I try to avoid because if we change the name of the variable, or make some other change later on, we have to change it in many places and not in one place in the library. 

Hehe, some variables don't change, hence the importance of naming them in the first place. Also, it takes 1 second for `sed` to change them in all scripts if we decide that later on. STATIC has always been static...

The new approach is cleaner. Maybe we could improve it even more? I was thinking of a way to check all the names of available folders in this repo, and run which works. It would take longer time though, but be MUCH more generic. Then we could add any folder name we want.

A regular `if [ -f ]` would work here.

> And have you really looked at this code:

Yes, I looked at it. I understand what you're trying to do but I think the old way works fine (copy the script to the new folder and use it in both places) until we find another solution which doesn't involve using old functions nested into each other (yes, I know `configure_php_fpm`, but that was my mistake, let's not do that again.

enoch85:
Not tested and just brainstorming, but what about something like this?

```
download_script() {
print_error() {
    exit 1
}
    if ! { curl_to_dir "${GITHUB_REPO}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.sh" "$SCRIPTS"; }
    then
        print_error
    elif ! { curl_to_dir "${GITHUB_REPO}" "${2}.py" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.py" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.py" "$SCRIPTS"; }
    then
        print_error
    elif ! { curl_to_dir "${GITHUB_REPO}" "${2}.php" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.php" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.php" "$SCRIPTS"; }
    then
        print_error
    fi
}
```

szaimen:
> Not tested and just brainstorming, but what about something like this?
> 
> ```
> download_script() {
> print_error() {
>     print_text_in_color "$IRed" "{$2} failed to download. Please run: 'sudo curl -sLO ${!1}/${2}.sh|.php|.py' again."
>     print_text_in_color "$ICyan" "If you get this error when running the nextcloud-startup-script then just re-run it with:"
>     print_text_in_color "$ICyan" "'sudo bash $SCRIPTS/nextcloud-startup-script.sh' and all the scripts will be downloaded again"
>     exit 1
> }
>     rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
>     if ! { curl_to_dir "${GITHUB_REPO}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.sh" "$SCRIPTS"; }
>     then
>         print_error
>     elif ! { curl_to_dir "${GITHUB_REPO}" "${2}.py" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.py" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.py" "$SCRIPTS"; }
>     then
>         print_error
>     elif ! { curl_to_dir "${GITHUB_REPO}" "${2}.php" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.php" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.php" "$SCRIPTS"; }
>     then
>         print_error
>     fi
> }
> ```

This is actually much more nested then any other solution before.

> Not tested and just brainstorming, but what about something like this?
> 
> ```
> download_script() {
> print_error() {
>     print_text_in_color "$IRed" "{$2} failed to download. Please run: 'sudo curl -sLO ${!1}/${2}.sh|.php|.py' again."
>     print_text_in_color "$ICyan" "If you get this error when running the nextcloud-startup-script then just re-run it with:"
>     print_text_in_color "$ICyan" "'sudo bash $SCRIPTS/nextcloud-startup-script.sh' and all the scripts will be downloaded again"
>     exit 1
> }
>     rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
>     if ! { curl_to_dir "${GITHUB_REPO}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.sh" "$SCRIPTS"; }
>     then
>         print_error
>     elif ! { curl_to_dir "${GITHUB_REPO}" "${2}.py" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.py" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.py" "$SCRIPTS"; }
>     then
>         print_error
>     elif ! { curl_to_dir "${GITHUB_REPO}" "${2}.php" "$SCRIPTS" || curl_to_dir "${STATIC}" "${2}.php" "$SCRIPTS" || curl_to_dir "${APP}" "${2}.php" "$SCRIPTS"; }
>     then
>         print_error
>     fi
> }
> ```

I am against it because it would need much longer to run any script and you won't be able to name two scripts in the whole repo the same which you would need to control. (the same name is here e.g. apps.sh or apps.py or apps.php)
Also it is actually much more nested: one function literally in another?

szaimen:
So then, let's do this and use run_script, and download_scipt everywhere.
I finally wan't to continue with #11!

szaimen:
I've chaned everything to the new function.
Can we merge this now?

enoch85:
Great, let's tune it later. 

We can start with this at least.

PR #1263, opened 2020-05-28
run scripts locally

### LINKED ISSUE #11: Check downloaded files with GPG before using them.

szaimen:
fix #11
@enoch85 it should be already working.
It is best to review the changes per commit.

### DIFF
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -34,42 +34,42 @@ case "$choice" in
     *"Fail2ban"*)
         clear
         print_text_in_color "$ICyan" "Downloading Fail2ban.sh..."
-        run_script APP fail2ban apps
+        run_script APP fail2ban
     ;;&
     *"Adminer"*)
         clear
         print_text_in_color "$ICyan" "Downloading Adminer.sh..."
-        run_script APP adminer apps
+        run_script APP adminer
     ;;&
     *"Netdata"*)
         clear
         print_text_in_color "$ICyan" "Downloading Netdata.sh..."
-        run_script APP netdata apps
+        run_script APP netdata
     ;;&
     *"OnlyOffice"*)
         clear
         print_text_in_color "$ICyan" "Downloading OnlyOffice.sh..."
-        run_script APP onlyoffice apps
+        run_script APP onlyoffice
     ;;&
     *"Collabora"*)
         clear
         print_text_in_color "$ICyan" "Downloading Collabora.sh..."
-        run_script APP collabora apps
+        run_script APP collabora
     ;;&
     *"Bitwarden"*)
         clear
         print_text_in_color "$ICyan" "Downloading Bitwarden.sh..."
-        run_script APP tmbitwarden apps
+        run_script APP tmbitwarden
     ;;&
     *"FullTextSearch"*)
         clear
         print_text_in_color "$ICyan" "Downloading FullTextSearch.sh..."
-        run_script APP fulltextsearch apps
+        run_script APP fulltextsearch
     ;;&
     *"PreviewGenerator"*)
         clear
         print_text_in_color "$ICyan" "Downloading PreviewGenerator.sh..."
-        run_script APP previewgenerator apps
+        run_script APP previewgenerator
     ;;&
     *"LDAP"*)
         clear
@@ -84,12 +84,12 @@ case "$choice" in
     *"Talk"*)
         clear
         print_text_in_color "$ICyan" "Downloading Talk.sh..."
-        run_script APP talk apps
+        run_script APP talk
     ;;&
     *"SMB-mount"*)
         clear
         print_text_in_color "$ICyan" "Downloading SMB-mount.sh..."
-        run_script APP smbmount apps
+        run_script APP smbmount
     ;;&
     *)
     ;;
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -5,7 +5,7 @@
 # shellcheck disable=2034,2059
 true
 # shellcheck source=lib.sh
-. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+source /var/scripts/main/lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
 
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -5,7 +5,7 @@
 # shellcheck disable=2034,2059
 true
 # shellcheck source=lib.sh
-source /var/scripts/main/lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -5,7 +5,7 @@
 # shellcheck disable=2034,2059
 true
 # shellcheck source=lib.sh
-. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+[ -f /var/scripts/main/lib.sh ] && source /var/scripts/main/lib.sh || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
--- a/apps/additional_apps.sh
+++ b/apps/additional_apps.sh
@@ -34,42 +34,42 @@ case "$choice" in
     *"Fail2ban"*)
         clear
         print_text_in_color "$ICyan" "Downloading Fail2ban.sh..."
-        run_script APP fail2ban
+        run_script APP fail2ban apps
     ;;&
     *"Adminer"*)
         clear
         print_text_in_color "$ICyan" "Downloading Adminer.sh..."
-        run_script APP adminer
+        run_script APP adminer apps
     ;;&
     *"Netdata"*)
         clear
         print_text_in_color "$ICyan" "Downloading Netdata.sh..."
-        run_script APP netdata
+        run_script APP netdata apps
     ;;&
     *"OnlyOffice"*)
         clear
         print_text_in_color "$ICyan" "Downloading OnlyOffice.sh..."
-        run_script APP onlyoffice
+        run_script APP onlyoffice apps
     ;;&
     *"Collabora"*)
         clear
         print_text_in_color "$ICyan" "Downloading Collabora.sh..."
-        run_script APP collabora
+        run_script APP collabora apps
     ;;&
     *"Bitwarden"*)
         clear
         print_text_in_color "$ICyan" "Downloading Bitwarden.sh..."
-        run_script APP tmbitwarden
+        run_script APP tmbitwarden apps
     ;;&
     *"FullTextSearch"*)
         clear
         print_text_in_color "$ICyan" "Downloading FullTextSearch.sh..."
-        run_script APP fulltextsearch
+        run_script APP fulltextsearch apps
     ;;&
     *"PreviewGenerator"*)
         clear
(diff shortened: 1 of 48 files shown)


nextcloud_install_production.sh
```diff
@@ -2,6 +2,71 @@
...
+
+    # Set ownership and permission
+    chown -R root:root "$SCRIPTS"
+    chmod -R +x "$SCRIPTS"
+
+    # Run the nextcloud_install_production script in a new process
+    exec "$SCRIPTS"/nextcloud_install_production.sh
+fi
+
+# -----------------------------------------------------------------------------------------------------------------------------------
```
szaimen, line 68: I know in this part are still some Todos left but they will only work after a new release. For now it gets the scripts directly from the branch. Later on they should be downloaded from the archive: see the "#Download the release package" section. The order is the following: remove the hashtatgs in that section, make a new release and it will work.

The big advantage is that the initial installation will work no matter if the network works or not since everything is already downloaded during the startup script.

We can add a checksum and possibly sign the release package later on, it is planned but not necessary for this feature to work.

szaimen:
To test all the changes:
0. make sure that it is a new ubuntu 20.04 installation and the /var/scripts directory doesn't exist.
1. simply run `curl -sLO https://raw.githubusercontent.com/nextcloud/vm/run_locally/nextcloud_install_production.sh && sud bash nextcloud_install_production.sh`
2. choose the "security" option during the startup script (you can choose between security or bugfixes; when choosing bugfixes, all local scripts get removed and you get the newest scripts from master)

vagrant/install.sh
```diff
@@ -1,13 +1,7 @@
...
-# shellcheck disable=2034,2059
-true
-# shellcheck source=lib.sh
-source /var/scripts/main/lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
+apt install curl -y
 
-check_command git clone https://github.com/nextcloud/vm.git
-
-cd vm || exit && print_text_in_color "$IRed" "Could not cd into the 'vm' folder."
+curl -sLO https://raw.githubusercontent.com/nextcloud/vm/run_locally/nextcloud_install_production.sh
```
szaimen, line 5: - [ ] needs to be changed back to the master branch after merge to master

nextcloud_install_production.sh
```diff
@@ -2,6 +2,71 @@
 
 # T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
 
+# Variables:
+# Todo: make DOWNLOAD_DIR readable again; is disabled for now to fix travis
+SCRIPTS=/var/scripts
+#DOWNLOAD_DIR=/tmp/NC_VM
+
+# Download and validate Nextcloud VM
+if ! [ -d $SCRIPTS ]
```
szaimen, line 11: @enoch85 is the /var/scripts directory always non-existant on a fresh ubuntu 20.04 installation?

apps/webmin.sh
```diff
@@ -42,6 +42,16 @@ else
     print_text_in_color "$ICyan" "Installing Webmin..."
 fi
 
+# Check if local script is available
+if [ -f "$SCRIPTS"/apps/webmin.sh ]
+then
```
enoch85, line 47: Since this is happening frequently, maybe it would be better to make it a function instead?

enoch85:
IMHO I think this is a too big PR to merge at once. 

Also, you added a lot of new logic, but didn't explain what it does, or why you added it.

Will this PR only allow local scripts from now on, or will it first check for local scripts, and if it doesn't exist then download it?

I think it would be better if you added this to another testing branch so that we can test it properly before merging to master.

szaimen:
> IMHO I think this is a too big PR to merge at once.

I am completely on your side with that.

> Also, you added a lot of new logic, but didn't explain what it does, or why you added it.

Which logic do you want to have explained? Please point to that code and I will try to do it.

> Will this PR only allow local scripts from now on, or will it first check for local scripts, and if it doesn't exist then download it?

It will "first check for local scripts, and if it doesn't exist then download it"

> I think it would be better if you added this to another testing branch so that we can test it properly before merging to master.

That is why I created the branch "run_locally" and added instructions: https://github.com/nextcloud/vm/pull/1263#issuecomment-635357376
We can take this as a test branch (is only not called testing) ;-)

apps/webmin.sh
szaimen: This happens (only) around 5 times, are you sure it makes sense?

szaimen:
BTW: I changed some logic concerning the checking of local scripts which is much more streamlined now; I think it should be possible to check the whole PR now since it is quite good readable now, imo.
So no need to check individual commits anymore

apps/webmin.sh
szaimen: if you say yes, then I add it.

szaimen:
it works like this: 
1. download all scripts during the beginning of the install production script by downloading the complete branch/archive to $SCRIPTS
2. use local lib if existing at the right place, else get the newest lib from github
3. download/run script by looking for a script first in subfolders of $SCRIPTS and otherwise if nothing was found get the newest one from github
4. let me choose between security or bugfixes, if security gets chosen, all local scripts will get used and nothing gets downloaded from the repository; if bugfixes gets chosen, all local scripts get removed and the scripts get downloaded from the repository.

szaimen:
@enoch85 the PR is rebased and should work again.

apps/additional_apps.sh
```diff
@@ -2,10 +2,10 @@
 
 # T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
 
-# shellcheck disable=2034,2059
+# shellcheck disable=2015,2034,2059
 true
 # shellcheck source=lib.sh
-. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+[ -f /var/scripts/main/lib.sh ] && source /var/scripts/main/lib.sh || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
```
szaimen, line 8: @enoch85, if you don't like that I had to disable SC2015 in order to fix travis, instead of writing it like this: 
```
[ -f /var/scripts/main/lib.sh ] && source /var/scripts/main/lib.sh || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
```
We could also write it like this:
```
[ -f /var/scripts/main/lib.sh ] && source /var/scripts/main/lib.sh
[ ! -f /var/scripts/main/lib.sh ] . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
``
Which solution do you like more? (We could remoce the disabling of SC2015 afterwards again)

Comment (enoch85) [reply apps/additional_apps.sh:8]: What about

```
if [ -f /var/scripts/main/lib.sh ]
then
    source /var/scripts/main/lib.sh
else
    source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
fi
```

Never tested with `source` for curl, but `.` means the same...

**Comment by enoch85** [comment]
> @enoch85 the PR is rebased and should work again.

I still think it's too big. Just look what happened to your other PR. It got reverted because all scripts weren't tested.

#11 is about using GPG keys, and to do that we need to change the whole way we do releases, not just run local lib files. I've been thinking about this, and I'm tending to make this a separate repo for the security concerned ones. I've not deiced yet. 

Since @ypid created the issue and I'm sure he has the knowledge to actually sort this matter, why include him in this? 

I'm not so keen to merge a PR which changes the way it's been working for years with stable results, happy users, and almost no bugs.

File: apps/additional_apps.sh (hunk 5 above)
Comment (szaimen) [reply apps/additional_apps.sh:8]: hm, I think it is a bit too long to be placed on the beginning of every file; I prefer a one-liner.
What about:
```
if [ -f /var/scripts/main/lib.sh ]; then source /var/scripts/main/lib.sh; else source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh); fi
```

btw: `source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)` seems to work!

Comment (szaimen) [reply apps/additional_apps.sh:8]: Please make the decision, then I change it accordingly.

**Comment by szaimen** [comment]
> I still think it's too big. Just look what happened to your other PR. It got reverted because all scripts weren't tested. Partly my fault.

That is the reason we can test it in the run_locally branch - no need to create an additional testing branch. The PR got reverted because there seems to be some edge cases where variables are nested into each other (my guess); in most situation it has worked (have tested that this was the case: the concept has worked)

> #11 is about using GPG keys, and to do that we need to change the whole way we do releases, not just run local lib files. I've been thinking about this, and I'm tending to make this a separate repo for the security concerned ones. I've not deiced yet.

I think it is too much effort to maintain a second repo only to for all those that want this feature.
I am sure that the way I planned it will work successfully. The way how we do releases can get changed later on; as I have commented: it is not necessary to be implemented in the first place.
We can also change that (important) part later; btw: here is a drone configuration how to do releases: https://github.com/lampepfl/dotty/commit/d04a8c9ddad2e9af638844b2312ce5b18d905bf4 maybe we could do it a bit like that?

> Since @ypid created the issue and I'm sure he has the knowledge to actually sort this matter, why include him in this?

That is okay for me, haven't thought about this - would be quite good to get his opinion on the current state.

> I'm not so keen to merge a PR which changes the way it's been working for years with stable results, happy users, and almost no bugs.

It shouldn't get changed until it is well tested - no rush to merge that PR;
But it would be nice, if you would reply to my comments and look at the code - I think then you will understand it much better.

**Comment by szaimen** [comment]
> I'm not so keen to merge a PR which changes the way it's been working for years with stable results, happy users, and almost no bugs.

BTW: from a users point of view nothing gets changed despite they have on whiptail more during the startup script where you also can choose to remove all pre-downloaded local scripts. If you choose to do so or remove them by hand later on, nothing will be different in comparison to the current form of running the VM with the advantage, that you will always be able to setup the VM based on the image that you have purchased/downloaded once (if you choose not to remove the pre-downloaded scripts) - that is the biggest advantage aside from having no longer the direct dependency on the repo (with the risk of security flaws, etc).

**Comment by ypid** [comment]
>> Since @ypid created the issue and I'm sure he has the knowledge to actually sort this matter, why include him in this?

> That is okay for me, haven't thought about this - would be quite good to get his opinion on the current state.

Hey guys, I am happy to help. I just don’t have time to read over the whole discussion and review this PR. Note that I am not using this project I just happened to get involved somehow. So what I can offer is review a proposal/specification on how to do secure script distribution. Something summarized. I guess I have mentioned it already. Just use the feature that git provides together with OpenPGP so there should not be much programming needed nor wanted (in this area).

**Comment by enoch85** [comment]
> there should not be much programming needed nor wanted (in this area).

@ypid Thanks for confirming.

File: nextcloud_install_production.sh [hunk 6]
```diff
@@ -2,6 +2,74 @@
...
+    # Download the release package (.tar.gz) file
+    # Todo: Remove the hashtags after it is released and change the download path if necessary
+    #rm -rf $DOWNLOAD_DIR
+    #mkdir $DOWNLOAD_DIR
+    #cd $DOWNLOAD_DIR
+    #curl -sLO https://github.com$(curl -s -m 900 https://github.com/nextcloud/vm/tags/ | grep -om1 "/nextcloud/vm/archive/[0-9.]*.tar.gz")
+
+    # Todo: Verify the state of the downloaded package with a checksum
+
+    # Todo: Verify the release package and integrity with the gpg key?
```
ypid, line 39: I guess what #11 talks about would then be handled here. So currently, there is not much for me to review.

szaimen:
> > there should not be much programming needed nor wanted (in this area).

I guess if that is the case, is this PR no longer needed?

@ypid Please confirm!

szaimen:
@ypid here is a summary of what the PR does: https://github.com/nextcloud/vm/pull/1263#issuecomment-636410001

enoch85:
In my opinion, this PR is not needed.

Running scripts locally is done in some cases since we need the variables without an internet connection. Other than that, it's not needed in my opinion.

ypid:
> @ypid here is a summary of what the PR does: #1263 (comment)

> if security gets chosen, all local scripts will get used and nothing gets downloaded from the repository; if bugfixes gets chosen, all local scripts get removed and the scripts get downloaded from the repository.

Not too sure if that addresses #11 ;-)

szaimen:
So I'll close it then

PR #1266, opened 2020-05-31
Revert "update lib.sh to change variables to functions"

enoch85:
Reverts nextcloud/vm#1247

This breaks a lot of scripts since the variables aren't imported until **after** the curl command is run, and everything is nested into each other. Collabora is one example.

### DIFF
--- a/apps/collabora.sh
+++ b/apps/collabora.sh
@@ -71,13 +71,13 @@ then
                     count=$((count+1))
                 fi
             done
-
+            
             msg_box "Collabora was successfully uninstalled."
             exit
         ;;
         "Reinstall Collabora")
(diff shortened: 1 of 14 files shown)


enoch85:
FYI @szaimen I knew there was a reason for doing it this way, now I remember. :)

szaimen:
@enoch85 I still think that it was prettier with functions but if you don't want to do it this way, I'm fine with that.

BTW: there is a way to fix collabora.
We should also think about trying to change them again to functions after collabora is migrated to the new way of installing collabora, don't you think so?

enoch85:
What needs to be done is test *all* the scripts before merging, with and without internet.

When that's done, and it still works - I'll consider to merge a PR that converts everything to functions instead.

PR #1270, opened 2020-06-01
remove local lib.sh

enoch85:
it's only needed during setup anyway


enoch85:
cc @szaimen It casues issues with menu.sh due to the new functions we added.

Issue #1278, opened 2020-06-04
Inconsistencies in update notification


k-matti:
Running updatenotification.sh gets me info about new version of Nextcloud 
`Checking for new Nextcloud version...
The notification regarding the new Nextcloud update has been already reported! (19.0.0)`

But in Administration -> Overview I can see:

![image](https://user-images.githubusercontent.com/66017397/83741383-6cbed500-a658-11ea-837b-8bfea60b37a7.png)

How can I update to 18.0.5?
Should I use update.sh script? But It wants to update to version 19
Why I didn't get notification about 18.0.5?

szaimen:
You can update to 18.0.5 by running 
`sudo bash /var/scripts/update.sh minor`

szaimen:
@enoch85 I mean we could also check if a new minor update is available but that would complicate things by a lot!

k-matti:
> @enoch85 I mean we could also check if a new minor update is available but that would complicate things by a lot!

Script to check if update exists should works that way.

In current implementation I need to check administration page every time because I don't get any update info about minor chages. I don't want to update to 19.0.0 for now

szaimen:
@enoch85 what do you think?

enoch85:
@k-matti The VM updates faster than the GUI. This is because Nextcloud doesn't push new releases all at once.

The default behaviour is to go to the latest major relase, but as @szaimen said - we made it possible to update to minor versions as well.

The automatic updater in the VM does minor updates, so that would be an option for you maybe @k-matti? Then you don't have to *check administration page every time because I don't get any update info about minor chages* since that would happen automatically.

I don't think we should over complicate things here.

k-matti:
„ The automatic updater in the VM does minor updates, so that would be an option for you maybe @k-matti?”

update.sh script is not a problem here. I know I can run „update.sh minor” problem is with updatenotification script. I have version 18.0.4 and want to be inform that there is version 18.0.5 available not only 19.0.0. Now update notification is misleading for the admins. The script informs that there is new version 19 but nothing about 18.0.5. What If admin dont want to update to 19 imidiately but want to update to minor version. Not update info then

k-matti:
I was thinking about this option in nc settings panel:
![image](https://user-images.githubusercontent.com/66017397/83849794-3cd30880-a710-11ea-802a-30779dcc7f10.png)
Shouldn't it trigger update to 18.0.5 notification the same as updatenotification.sh script?

PR #1281, opened 2020-06-09
improve updatenotification.sh

szaimen:
fix https://github.com/nextcloud/vm/issues/1278

### DIFF
--- a/static/updatenotification.sh
+++ b/static/updatenotification.sh
@@ -19,32 +19,58 @@ debug_mode
 # Check if root
 root_check
 
-REPORTEDNCVERSION=""
+NCMIN=$(curl -s -m 900 $NCREPO/ | sed --silent 's/.*href="nextcloud-\([^"]\+\).zip.asc".*/\1/p' | sort --version-sort | grep "${CURRENTVERSION%%.*}" | tail -1)
+REPORTEDMAJ=""
+REPORTEDMIN=""
 
-if [ "$CURRENTVERSION" == "$NCVERSION" ]
+if [ "$CURRENTVERSION" == "$NCVERSION" ] && [ "$CURRENTVERSION" == "$NCMIN" ]
 then
     print_text_in_color "$IGreen" "You already run the latest version! ($NCVERSION)"
     exit
 fi
 
-if [ "$REPORTEDNCVERSION" == "$NCVERSION" ]
+if [ "$REPORTEDMAJ" == "$NCVERSION" ] && [ "$REPORTEDMIN" == "$NCMIN" ]
 then
-    print_text_in_color "$ICyan" "The notification regarding the new Nextcloud update has been already reported! ($NCVERSION)"
+    print_text_in_color "$ICyan" "The notification regarding the new Nextcloud update has been already reported!"
     exit
 fi
 
-if version_gt "$NCVERSION" "$CURRENTVERSION"
+if [ "$NCVERSION" == "$NCMIN" ] && version_gt "$NCMIN" "$REPORTEDMIN"
 then
-    if crontab -l -u root | grep $SCRIPTS/update.sh
+    sed -i "s|^REPORTEDMAJ.*|REPORTEDMAJ=$NCVERSION|" $SCRIPTS/updatenotification.sh
+    sed -i "s|^REPORTEDMIN.*|REPORTEDMIN=$NCMIN|" $SCRIPTS/updatenotification.sh
+    if crontab -l -u root | grep -q $SCRIPTS/update.sh
     then
         notify_admin_gui \
         "New Nextcloud version!" \
-        "Nextcloud $NCVERSION just became available. Since you are running Automatic Updates at $AUT_UPDATES_TIME:00, you don't need to bother about updating the server manually, as that's already taken care of."
-        sed -i "s|^REPORTEDNCVERSION.*|REPORTEDNCVERSION=$NCVERSION|" $SCRIPTS/updatenotification.sh
+        "Nextcloud $NCVERSION just became available. Since you are running Automatic Updates on Saturdays at $AUT_UPDATES_TIME:00, you don't need to bother about updating the server manually, as that's already taken care of."
     else
         notify_admin_gui \
         "Update available!" \
         "Nextcloud $NCVERSION is available. Please run 'sudo bash /var/scripts/update.sh' from your CLI to update your server."
-        sed -i "s|^REPORTEDNCVERSION.*|REPORTEDNCVERSION=$NCVERSION|" $SCRIPTS/updatenotification.sh
     fi
+    exit
+fi
+
+if version_gt "$NCMIN" "$REPORTEDMIN"
+then
+    sed -i "s|^REPORTEDMIN.*|REPORTEDMIN=$NCMIN|" $SCRIPTS/updatenotification.sh
+    if crontab -l -u root | grep -q $SCRIPTS/update.sh
+    then
+        notify_admin_gui \
+        "New Nextcloud version!" \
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 This time, I've tested all possible cases. 
The only thing that could improve is the notification about new updates: I think it would make sense to write into the title if it is a major or minor update that became available. What do you think?

enoch85:
@k-matti Please review this PR.

Thanks!

szaimen:
@k-matti if you don't understand something, please don't hesitate to ask!

static/updatenotification.sh
```diff
@@ -19,32 +19,58 @@ debug_mode
...
+    print_text_in_color "$ICyan" "The notification regarding the new Nextcloud update has been already reported!"
     exit
 fi
 
-if version_gt "$NCVERSION" "$CURRENTVERSION"
+if [ "$NCVERSION" == "$NCMIN" ] && version_gt "$NCMIN" "$REPORTEDMIN" && version_gt "$NCMIN" "$CURRENTVERSION"
 then
-    if crontab -l -u root | grep $SCRIPTS/update.sh
+    sed -i "s|^REPORTEDMAJ.*|REPORTEDMAJ=$NCVERSION|" $SCRIPTS/updatenotification.sh
+    sed -i "s|^REPORTEDMIN.*|REPORTEDMIN=$NCMIN|" $SCRIPTS/updatenotification.sh
```
k-matti, line 41: Why double notification here?

static/updatenotification.sh
```diff
@@ -19,32 +19,58 @@ debug_mode
...
-        "Update available!" \
-        "Nextcloud $NCVERSION is available. Please run 'sudo bash /var/scripts/update.sh' from your CLI to update your server."
-        sed -i "s|^REPORTEDNCVERSION.*|REPORTEDNCVERSION=$NCVERSION|" $SCRIPTS/updatenotification.sh
+        "New minor Nextcloud Update!" \
+        "Nextcloud $NCMIN just became available. Please run 'sudo bash /var/scripts/update.sh minor' from your CLI to update your server to Nextcloud $NCMIN."
     fi
+    exit
+fi
+
+if version_gt "$NCMIN" "$REPORTEDMIN" && version_gt "$NCMIN" "$CURRENTVERSION"
```
k-matti, line 55: Why check ncmin again here nad in if started in line 38?

szaimen:
> How updatenotification.sh will work with build in notification after this changes?
> image
> 
> Context:
> When I was on version 18.0.4 and version 18.0.5 came up I didn't get any notification from updatenotification.sh but it is clear why, on the other hand why I didn't get any notification from build in notification as on the image? 
> Is this build in notification service disabled when using updatenotification.sh on vm?
> 

Yes, these build in notifications for updates are disabled if you enabled automatic updates.

static/updatenotification.sh
```diff
@@ -19,32 +19,58 @@ debug_mode
 # Check if root
 root_check
 
-REPORTEDNCVERSION=""
+NCMIN=$(curl -s -m 900 $NCREPO/ | sed --silent 's/.*href="nextcloud-\([^"]\+\).zip.asc".*/\1/p' | sort --version-sort | grep "${CURRENTVERSION%%.*}" | tail -1)
```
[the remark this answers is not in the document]
szaimen: @enoch85 what do you think, does that make sense? ⬆️

k-matti:
> > How updatenotification.sh will work with build in notification after this changes?
> > image
> > Context:
> > When I was on version 18.0.4 and version 18.0.5 came up I didn't get any notification from updatenotification.sh but it is clear why, on the other hand why I didn't get any notification from build in notification as on the image?
> > Is this build in notification service disabled when using updatenotification.sh on vm?
> 
> Yes, these build in notifications for updates are disabled if you enabled automatic updates.

I have automatic updates disabled, only update notification enabled.
Can you show me the script and line number where this build in notification gets disabled during setting up vm?

static/updatenotification.sh
szaimen: This doesn't send a notification but substitutes both variables reportedmin and reportedmaj through new values to track which version got already reported and is defjnjtely needed here.

What sends the notification is notify_admin_gui

static/updatenotification.sh
szaimen: Because the check in line 48 covers the case that ncversion and ncmin are the same which is possible. And in this case we only send one notification. 

This check here covers ncmin to check if there is a new minor update available and sends a notification for that.

Later the check for ncversion covers the case that there is a new major update available and also sends a notification in that case. 

So the idea is that if ncmin and ncversion are the same, you get only one notification. If they are not the same (e.g. there is a new minor and major update available) you get two notifications.

But if only e.g. a major update is available you get only one notification and if only a minor update is available you get also only one notification.

szaimen:
Yes, it is here:
https://github.com/nextcloud/vm/blob/4a9579a580a72f04e9404a9924e701a9e4affbcd/static/automatic_updates.sh#L35
That changes the settings of you screenshot. So if you put the admin group or any other group into that input field, you should get nextcloud notifications about updates again.

k-matti:
> Yes, it is here:
> 
> https://github.com/nextcloud/vm/blob/4a9579a580a72f04e9404a9924e701a9e4affbcd/static/automatic_updates.sh#L35
> 
> That changes the settings of you screenshot. So if you put the admin group or any other group into that input field, you should get nextcloud notifications about updates again.

That is the problem, I had admin group there when update from 18.0.4 to 18.0.5 came up but no notification. Only notification i got was from updatenotification.sh when nc 19 came up

szaimen:
This is not a problem, it is mesnt to be this way since you would get two times notification about new updates if we wouldn't disable the build in notifications.

E.g. we send the notification about a new minor release, and the nextcloud system also sends one => you have two notifications.

This is the reason why I think this PR solves all cases.

enoch85:
Waiting for @k-matti before merging.

k-matti:
How to update changed files in my /var/scripts/ folder?

szaimen:
> How to update changed files in my /var/scripts/ folder

There is no official way afaik. You could just manually replace them by substituting them through the new scripts.

k-matti:
It is not good then...

enoch85:
@k-matti 

> It is not good then...

You're welcome to create a PR with a fix for it. :)

enoch85:
We currently have `menu.sh` which runs most of the configuration scripts in this repo. It could be extended of course.

k-matti:
Downloading the newest version of scripts should happen when update.sh is starting

enoch85:
@k-matti Exactly which scripts did you have in mind? Should we clone the whole repo on every update?

We don't want to put too much traffic on Github servers and since the VM is used by thousands of users it would be pretty heavy to update all scripts in every run - instead of just getting the script manually instead, if you want.

It's also hard to know which script we should download, all users maybe don't want all scripts? Me for instance like it as clean as possible.

If there are any major change that affects the usability that will go in update.sh for a temporary time, not otherwise.

k-matti:
From my experience as a developer I can imagine that every time when update.sh script is trigger (manually or by crone) it should download /scrips folder from master branch to /var/scripts (or somehow check if there are new versions of scripts, but first approach is easier)

This repo is still in change, what about people that downloaded it years ago. 
There were plenty bug fixes and improvements in /scripts. 
There is big improvement in updatenotification.sh from this PR, I know about it (and I know I need to manually download new version) but what about other users?
It change almost every week, on the other hand update.sh is trigger to run every saturdays by default but always in old version

It's just my thoughts, you can ignore it.

PR #1282, opened 2020-06-09
Create smartctl.sh

enoch85:
Is addon really a good folder name?

FYI @szaimen

### DIFF
--- a/addons/smartctl.sh
+++ b/addons/smartctl.sh
@@ -28,11 +28,12 @@ then
     notify_admin_gui "S.M.A.R.T results weekly scan (sda)" "$(smartctl --all /dev/sda)"
 else
     # get all disks into an array
-    disks="$(ls /dev/sd* | grep -v '[0-9]')"
+    disks="$(fdisk -l | grep Disk | grep /dev/sd | awk '{print$2}' | cut -d ":" -f1)"
     # loop over disks in array
-    for disk in $(echo ${disks[@]}); do
+    for disk in "$(printf ${disks[@]})"
+    do
         if [ -n "$disks" ]; then
-             notify_admin_gui "S.M.A.R.T results weekly scan ($disk)" "$(smartctl --all $disk)"
+             notify_admin_gui "S.M.A.R.T results weekly scan (\$disk)" "$(smartctl --all $disk)"
         fi
     done
 fi
--- a/addons/smartctl.sh
+++ b/addons/smartctl.sh
@@ -24,9 +24,16 @@ install_if_not smartmontools
 # Add a crontab to check the disk, and post the output with notify_admin_gui ever week (maybe with updatenotification?)
 if home_sme_server
 then
-    smartctl --all /dev/nvme0n1
-    smartctl --all /dev/sda
+    notify_admin_gui "S.M.A.R.T results weekly scan" "$(smartctl --all /dev/nvme0n1)"
+    smartctl --all "S.M.A.R.T results weekly scan" "$(/dev/sda)"
 elif [check which disks are used]
     smartctl --all /dev/sda
     smartctl --all /dev/sdb
 fi
+
+# Add crontab “At 06:12 on Monday.” 
+if ! crontab -l | xargs grep -l "$SCRIPTS/smartctl.sh"
+then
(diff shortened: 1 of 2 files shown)


szaimen:
> Is addon really a good folder name?

I guess so, don't have a better idea currently for a better folder name.

enoch85:
@szaimen So what would classify as "addons"? Isn't configuration better, or no?

szaimen:
Hm, but this script isn't configuring anything, is it?

I would say a folder named configuration would fit better for those menu scripts like main_menu.sh or server_configuration.sh and nextcloud_condiguration.sh.

enoch85:
I would like to add:

- Menus
- Configuration
- Network
- Addons

Anything more?

szaimen:
> I would like to add:
> 
> - Menus
> - Configuration
> - Network
> - Addons
> 
> Anything more?

No, I think those should fit all. At least in combination with stable, old and apps.

enoch85:
@szaimen Ready for review. Pretty basic stuff so no need for an explanation.

enoch85:
No need to review. We can improve it later. :+1:

static/server_configuration.sh
```diff
@@ -36,9 +36,14 @@ choice=$(whiptail --title "Server configurations" --checklist "Choose what you w
 "Activate TLS" "(Enable HTTPS with Let's Encrypt)" ON \
 "Security" "(Add extra security based on this http://goo.gl/gEJHi7)" OFF \
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
+"Disk Check" "(Check for S.M.A.R.T errors on your disks every week on Mondays)" OFF \
 "Automatic updates" "(Automatically update your server every week on Sundays)" OFF 3>&1 1>&2 2>&3)
 
 case "$choice" in
+    *"Disk Check"*)
+        clear
+        run_script STATIC smartctl
```
szaimen, line 45: This wont work because you need to add a variable to lib.sh for the addons folder.

Also wouldn't it be better to download it during the startup script and execute it if it exists? So that it is Independent from the internet connection again?

szaimen: Aso if all options are checked, why should it get executed at the beginning before the security and static ip option?
I would execute it after static ip and before automatic updates.

addons/smartctl.sh
```diff
@@ -0,0 +1,47 @@
...
+then
+    notify_admin_gui "S.M.A.R.T results weekly scan (nvme0n1)" "$(smartctl --all /dev/nvme0n1)"
+    notify_admin_gui "S.M.A.R.T results weekly scan (sda)" "$(smartctl --all /dev/sda)"
+else
+    # get all disks into an array
+    disks="$(fdisk -l | grep Disk | grep /dev/sd | awk '{print$2}' | cut -d ":" -f1)"
+    # loop over disks in array
+    for disk in $(printf "${disks[@]}")
+    do
+        if [ -n "$disks" ]
```
szaimen, line 35: Is this check really needed?

addons/smartctl.sh
```diff
@@ -0,0 +1,47 @@
...
+# Add a crontab to check the disk, and post the output with notify_admin_gui ever week (maybe with updatenotification?)
+if home_sme_server
+then
+    notify_admin_gui "S.M.A.R.T results weekly scan (nvme0n1)" "$(smartctl --all /dev/nvme0n1)"
+    notify_admin_gui "S.M.A.R.T results weekly scan (sda)" "$(smartctl --all /dev/sda)"
+else
+    # get all disks into an array
+    disks="$(fdisk -l | grep Disk | grep /dev/sd | awk '{print$2}' | cut -d ":" -f1)"
+    # loop over disks in array
+    for disk in $(printf "${disks[@]}")
```
szaimen, line 33: Maybe remove the for loop and do the following?

`fdisk -l | grep Disk | grep /dev/sd | awk '{print$2}' | cut -d ":" -f1 | while read -r disk`

addons/smartctl.sh
enoch85: Just in case...

addons/smartctl.sh
enoch85: So you mean that you would choose disk yourself?

PR #1286, opened 2020-06-11
one small improvement to updatenotification.sh

szaimen:
@enoch85 this change was already tested by me so ready to go.


enoch85:
OK, what does it do, and what's the difference?

szaimen:
It just sets reportedmaj and reportedmin at the beginning to the value of currentversion.

enoch85:
> It just sets reportedmaj and reportedmin at the beginning to the value of currentversion.

Haha yes, I can see that, but why? :D

szaimen:
It fixes a logic error which I have forgotten in the last update to this script.
But it is only a cosmetic fix and nothing to worry about.

Issue #1284, opened 2020-06-10
Create a new "official" branch


enoch85:
Had a talk with @karlitschek today and there will be some changes to how the VM are presented on the [downloads page](https://nextcloud.com/install/#instructions-server).

To avoid confusion for the end users (some think that the bought versions of Hansson IT gives the [enterprise support](https://nextcloud.com/enterprise/), that's not the case) there will be a new "Official VM" which is stipped from everything related to Hansson IT.

Basically there will be two options:
1. The main and official (stripped-down vanilla) Nextcloud VM (no Hansson IT or anything like that)
2. An option to download the full blown Hansson IT VM

### What I'm thinking:

**Make a VM that automatically deploys the latest version.**
Make it run the install script (modified and stripped down) with a default username and password. (ncadmin/nextcloud) which will install everything from the ground up. This would remove the need for making new releases and one single OVA could be used, and not changed.

**No need for a startup script**
The current VM comes with a lot of extras which are set in the startup script. It installs apps that the user wants, adds TLS, and make some modifications to make sure the user gets a full blown Nextcloud experience (OnlyOffice, Collabora, Talk and so on). The official VM will not - it should only be a plain Nextcloud installation. So in other words, the only thing needs changing is the default password. We could make that whole thing a function and use it in the installation script directly, as well in the startup script of the full blown VM to avoid extra maintenance work.

**Change permissions**
The current VM use a built in update script to make sure everything (the whole server) is updated with the latest dependencies, it also sets secure permissions. The offiicial VM will not contain a update script. Updates should be handled in the GUI. So in other words, set `www-data` permissions all over.

**Remove the whole /var/scripts folder**
Since the official Nextcloud VM will be a "vanilla" install, there's no need to offer any improvements or modifications at all - hence the whole scripts folder in the VM is not needed.

### What we need to do:
1. Create a new branch in this repo with a modified install script.
2. Create a new OVA (I will do that) which automatically runs the install script once booted.
3. Release the official OVA on Nextclouds server 
4. Add some documentation containing passwords and such.

### When the new OVA is ready
1. Make a PR to change the Nextcloud Downloads page. Make the official VM stand out, and add an option below (or on the side) to download the full blown VM made by Hansson IT (as discussed)

@szaimen You're very welcome to come aboard on this.
@karlitschek Please let me know if we are on the same page here.
@jospoortvliet FYI

Feel free to edit this post if needed.

child-duckling:
Want this!

[...]

PR #1302, opened 2020-06-29
[very much WIP] Official VM

enoch85:

This is the first PR to fix https://github.com/nextcloud/vm/issues/1284 which is a plain install on Ubuntu Server without *any* fixes at all, just a plain vanilla installation.

FYI @karlitschek 

### If you want to test:
1.Create a clean Ubuntu Server 20.04 with ncadmin / nextcloud as user and pass
2. Run `prep_vm.sh`
3. Reboot
4. The VM should automatically start installing Nextcloud when you login for the first time


enoch85:
Let's merge this to avoid further conflicts

PR #1306, opened 2020-07-01
Nextcloud docker (Collabora) DNS does not resolve (nextcloud#1213)

### LINKED ISSUE #1213: Nextcloud docker (Collabora) DNS does not resolve (with possible solution)
Collabora docker image incorrectly requires Internet gateway port forwarding and loopback connection from docker image to external Internet gateway IP address back to the Nextcloud server due to bad default DNS setup.

Can the VM be improved or at least FAQ documentation.

## Steps to reproduce
1. Install Nextcloud VM with Collabora docker image.
2. Set fixed IP address, gateway and DNS server in Ubuntu netplan config file. Note DNS server is on another server in the subnet.
3. Stop port forwarding on Internet gateway to Nextcloud server on 192.168.0.100:443.
4. Try to edit a document using Nextcloud Collabora

## Expected behaviour
It is possible to edit a document

## Actual behaviour
Collabora cannot open the document. nextcloud.domain.net inside Collabora docker image resolves to external IP address of Internet gateway. This is not good as it means that all Collabora document opens are going via the Internet gateway when they can be connected directly on the VM.

The problem is with systemd-resolve or docker settings. The VM /etc/hosts file does not appear to be used inside the Collabora docker image and it gets DNS settings from systemd-resolve which has per-interface DNS resolvers. Setting a fixed DNS server against an interface in Netplan only affects that interface and not the docker interface.

I fixed the problem by setting the systemd-resolve default DNS server. See proposed fix below. Maybe there is a better way of solving this?

Please fix or document in FAQ what needs to be done

## Server configuration

18.0.4

Default - Ubuntu 18.04.4

Release install

## Network
**Do you use DHCP?**
No

On Internet gateway, port forwarding done during installation only. Removed afterwards.

#### Netplan config file (/etc/netplan/01-netcfg.yaml)
```
network:
   version: 2
   renderer: networkd
   ethernets:
       enp0s3: #object name
         dhcp4: no # dhcp v4 disable
         dhcp6: no # dhcp v6 disable
         addresses: [192.168.0.100/24] # client IP address
         gateway4: 192.168.0.1 # gateway address
         nameservers:
           addresses: [192.168.0.50] #name servers
```

#### Log file (docker logs xxx)
```
wsd-00006-00036 2020-04-27 17:10:00.790250 [ docbroker_001 ] ERR  Cannot get file info from WOPI storage uri [https://nextcloud.domain.net/index.php/apps/richdocuments/wopi/files/xxx?access_token=xxx&access_token_ttl=0]. Error: Connection refused| wsd/Storage.cpp:564
```

#### SystemD resolver before fix (systemd-resolve --status)
```
Global
         DNS Servers: 9.9.9.9 ...
          DNSSEC NTA: 10.in-addr.arpa
                      16.172.in-addr.arpa
...
Link 3 (docker0)
      Current Scopes: none
       LLMNR setting: yes
MulticastDNS setting: no
      DNSSEC setting: no
    DNSSEC supported: no

Link 2 (enp0s3)
      Current Scopes: DNS
       LLMNR setting: yes
MulticastDNS setting: no
      DNSSEC setting: no
    DNSSEC supported: no
         DNS Servers: 192.168.0.50
```

#### Proposed fix (/etc/systemd/resolved.conf)
```
...
[Resolve]
DNS=192.168.0.50
...
```

aaaskew:
Included here are my proposed changes for the docker DNS not following the Internet interface DNS.

I beleive the code is what is required with a further change to support only one DNS server when setting a static IP address (so that it produces the same result as DHCP)

This is my first pull request ever so I am not sure if I have created forks, branches, etc without need. I would have liked to do a clone to code under Linux and pushed it back but I have committed this all via the GitHub web interface.

I would be happy to accept a better way of extracting multi-lined output from 'systemd-resolve --status' and putting it all on one line, space separated, but the regex is the way i achieved it.

fixes #1213


aaaskew:
Suggested code in pull request #1306

static/static_ip.sh
```diff
@@ -177,14 +178,14 @@ ENTERNS2
...
-# Check if DNS is set manaully and set variables accordingly
-if [ -n "$NSIP1" ]
-then
-    DNS1="$NSIP1"
-fi
-
-if [ -n "$NSIP2" ]
+# Use the updated DNS IP addresses
+DNSs="$NSIP1"
+DNS1="$NSIP1"
```
enoch85, line 183: Have a look at this: https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L203

The default behaviour is setting Quad9s DNS, and that's how the VM are released. In other words, NSIP1 is never empty.

Point 2: Why do you set the same DNS for both DNS1 and DNS2?

Point 3 Why don't you use DNS2 in the netplan config? 

Point 4: Maybe I'm tired but I'm not following here. Please explain.

static/static_ip.sh
```diff
@@ -219,12 +221,13 @@ network:
          addresses: [$ADDRESS/24] # client IP address
          gateway4: $GATEWAY # gateway address
          nameservers:
-           addresses: [$DNS1,$DNS2] #name servers
+           addresses: [$DNSs] #name servers
```
enoch85, line 224: Why this change?

static/static_ip.sh
```diff
@@ -177,14 +178,14 @@ ENTERNS2
...
 
-# Check if DNS is set manaully and set variables accordingly
-if [ -n "$NSIP1" ]
-then
-    DNS1="$NSIP1"
-fi
-
-if [ -n "$NSIP2" ]
+# Use the updated DNS IP addresses
+DNSs="$NSIP1"
```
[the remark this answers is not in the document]
enoch85: Hmm, you're never changing the fallback.

The script says that you should enter a second NS (server which is the FallBackserver), but you just add the server to the primary DNS.

Any reason for that?

static/static_ip.sh
enoch85: Point 3, Sorry I can see that you do. It's not used as Fallback in resolved though. Any reason for that?

aaaskew:
I appologise in advance for the long explaination but there are misconceptions as to what the FallbackDNS SystemD option does. I have included my explaination and the relevant man page section.

Changes to static_ip.sh
-----------------------

Please test by running the 'systemd-resolve --status' command as this will show the 'Global DNS Servers:' list and 'Link 2 (enp0s3) DNS Servers:' list that I want to make the same. If 

Link 2 (enp0s3)
DNS Servers: 192.168.2.1
             149.112.112.112

Then

Global
DNS Servers: 192.168.2.1
             149.112.112.112


The fact that you have one internal and one external is I believe because you kept the default DNS server for $DNS2.

I will update the text to make clean what should be set for the second DNS server or if no DNS server should be set.

Question:
---------

DNS1=
DNS2=

are defined in lib.sh and used in static_ip.sh. static_ip.sh is then used in nextcloud-startup-script.sh.

Do I need to set these variables or are they then ignored once the end of he static_ip.sh file is reached?

The reason I ask is I use the new variable '$DNSs' instead after then questions have been answered to fill in the NetPlan configuration with either 1 or 2 DNS servers.

My changes also assumed that if nothing was changed then $NSIP1 and $NSIP2 would take $DNS1 and $DNS2 default vaules so I could remove the old code that seemed to be redundant based on the code paths before. Tell me if I am wrong there.


SystemD DNS:
------------

- Per-link DNS servers: A list of multiple DNS servers, set by NetPlan in the Nextcloud VM
- SystemD resolved.conf Global DNS= servers: Currently hard coded to some Internet servers. Should be the same as the Per-link DNS server list so that links that do not have per-link DNS servers applied, like docker interfaces, will use this list.
- SystemD resolved.conf Global FallbackDNS= servers: Will NEVER be used if DNS= is set!
- SystemD ResolveD compiled-in list of DNS servers: Will NEVER be used unless DNS= and FallbackDNS= are not defined!
- The DNS servers used could be affected by Domains=~ lists but this is not used in the Nextcloud VM.

My understanding is some people did not like the fact that there is a hidden compiled-in list of DNS servers so they asked to have a FallbackDNS option that they could set to empty to stop it being used, or at least make a list visible. Others argued that 'it should just work' without configuration, so there is a compiled in list.

What this means is that the default Ubuntu settings will use per-link (DHCP or statically defined) DNS servers and if not, fall back to the SystemD ResolveD compiled-in list.

Note the manual mentions that one entry from the DNS= list will be used in parallel to a per-link DNS entry for redundancy. The reduncancy is not between DNS= and FallbackDNS=.

I believe the Nextcloud VM image should use this same resolved.conf file when installed. Then when you set the DNS server using DHCP or static provisioning, then the per-link DNS server list will be copied into the resolved.conf DNS= setting.

What we should not do: Set two DNS servers, one inside home/company network and one on the Internet. This would cause problems if the home/company DNS is a split-DNS that resolves the Nextcloud FQDNs to local IP addresses while the Internet DNS would resolve the Nextcloud FQDNs to external addresses. You already have tests that use forced external DNS lookups, which is good when we need to check external conectivity, but all other DNS lookups should produce local results to make sure we don't use hairpin/loopback NAT.

Ubuntu 20.04 default /etc/systemd/resolved.conf:
------------------------------------------------
...
[Resolve]
#DNS=
#FallbackDNS=
#Domains=
...

Extract from man resolved.conf:
-------------------------------
OPTIONS
       The following options are available in the "[Resolve]" section:

       DNS=
           A space-separated list of IPv4 and IPv6 addresses to use as system
           DNS servers. DNS requests are sent to one of the listed DNS servers
           in parallel to suitable per-link DNS servers acquired from systemd-
           networkd.service(8) or set at runtime by external applications. For
           compatibility reasons, if this setting is not specified, the DNS
           servers listed in /etc/resolv.conf are used instead, if that file
           exists and any servers are configured in it. This setting defaults
           to the empty list.

       FallbackDNS=
           A space-separated list of IPv4 and IPv6 addresses to use as the
           fallback DNS servers. Any per-link DNS servers obtained from
           systemd-networkd.service(8) take precedence over this setting, as
           do any servers set via DNS= above or /etc/resolv.conf. This setting
           is hence only used if no other DNS server information is known. If
           this option is not given, a compiled-in list of DNS servers is used
           instead.

static/static_ip.sh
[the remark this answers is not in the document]
aaaskew: See my big comment below but basically FallbackDNS should be commented out I believe.

static/static_ip.sh
```diff
@@ -160,6 +160,7 @@ ENTERNS1
 +-------------------------------------------------------+
 |    Please enter the local nameserver address you want |
 |    to set. Just hit enter to choose the current NS2.  |
+|    If you only have one nameserver then enter none.   |
```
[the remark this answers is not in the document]
aaaskew: I have updated this to make it a bit clearer now.

static/static_ip.sh
aaaskew: I have not changed the nextcloud_install_production.sh file but I believe it needs updating based on my big comments below.

aaaskew: For points 2, I hope I don't set the same DNS for DNS1 and DNS2. I put DNS1 and DNS2 into the DNSs variable and use that for the netpan file

static/static_ip.sh
aaaskew: Having the 1 DNSs variable allows me to set either [1.1.1.1,2.2.2.2] or [1.1.1.1] in the field with the comma or not based on if I only have 1 DNS server set or 2.

static/static_ip.sh
aaaskew: Updated code to pass through DNS1 and DNS2 to DNSs if no DNS servers set

static/static_ip.sh
```diff
@@ -159,11 +159,14 @@ ENTERNS1
         cat << ENTERNS2
 +-------------------------------------------------------+
 |    Please enter the local nameserver address you want |
-|    to set. Just hit enter to choose the current NS2.  |
-|    If you only have one nameserver then enter none.   |
+|    to set. The 3 options are:                         |
+|    - Hit enter to choose the current NS2.             |
+|    - Enter a new IP address for NS2.                  |
+|    - Enter the text 'none' if you only have one       |
+|      nameserver.                                      |
```
enoch85, line 166: Please make this 1 NS so that it fits in one line.

enoch85:
Thanks for a great explanation! 

I will test this when I have time. Probably tomorrow.

Just to explain; the reason for defaulting to external DNS is because this VM is downloaded by thousands of users, mostly (I guess) are not versed with DNS (or what it even does). Many people in the past failed to get internet connection because they didn't have DNS setup correctly in their environment - hence the VM will force it to make sure it always works.

Regarding the install script, please make the change you think is valid, and we can discuss that one too. I think that it should be an option though. Maybe to set option 3 (which is the local option) to use this new function?

Not tested, but will [this ever be set](https://github.com/nextcloud/vm/blob/8955e1786de83ced4e8b4b5b67e7759507a62ad6/static/static_ip.sh#L200) as it's not specified in the netplan config? Not tested but I think it looks strange.

enoch85:
Same thing [here](https://github.com/nextcloud/vm/blob/8955e1786de83ced4e8b4b5b67e7759507a62ad6/static/static_ip.sh#L190), I mean you only use `DNSs` in the netplan config...

enoch85:
Also, I do belive we should have an IPv6 DNS in there, maybe that's better suited for FallbackDNS?

enoch85:
Another bug, when already running the script and you run it again - current NS is wrong in the `msg_box` since it states `DNS1` and `DNS2` which defaults to Quad9.

aaaskew:
I have made changes to the nextcloud_install_production.sh file as you mention this as a point where DNS is set. As per my explanation above, I have removed the FallbackDNS= settings, as these are not used and put all of the options on the DNS= line. There are still ongoing arguments about how SystemD Resolver handles the list of DNS servers. At the moment it will pick the first in the list and keep using that until it fails, then it will make all requests to the second until that fails, etc. This means that the IPv6 resolvers will be used if the IPv4 resolvers fail.

As we are not using Domains to use different resolvers based on the domain queried, when picking 'Local' only the default route is set and not an external DNS server. These two DNS servers are not equivilant as the local DNS server might contain local domain information that the Internet resolver will not have. If the local resolver fails once, then all subsequent queries will produce invalid results for local domain queries.

I have added a little extra text as a home setup is likley to have the DNS server on the default gateway but companies will not have this. An option is missing to set the DHCP set DNS server list but this might be done in a future update.

The last bug you mentioned about re-running the script reverting back to the Quad9 DNS servers is still outstanding but I don't think I introduced that issue. I believe it was already the case. To fix that, the static_ip script, or the lib.sh script needs to set the default DNS1 and DNS2 values to what is currently in SystemD. I could do that but can we make sure everything else I have changed is ok first?

enoch85:
We should be careful with the installation script. I think that we should at least use a `while :` loop or similar which checks if DNS is working and if not, ask the user for a new input or fallback to default which is external.

> I believe it was already the case

Yup, that was the case before as well. But I think we should improve it now when we have the chance. The script was only ment to be run once, but hey, maybe you want to use it again - in that case the text would be confusing. How it should look like:

1. Show the default nameservers.
2. Show the current nameservers.

Also, please look into if all variables is really needed now when everything is in one line? The logic is still a bit confusingl I think.

Thanks for your efforts so far! This looks promising.

static/static_ip.sh
enoch85: :+1:

nextcloud_install_production.sh
```diff
@@ -198,24 +198,21 @@ esac
...
+        sed -i "s|#DNS=.*|DNS=9.9.9.9 149.112.112.112 2620:fe::fe 2620:fe::9|g" /etc/systemd/resolved.conf
     ;;
     "Cloudflare")
-        sed -i "s|#DNS=.*|DNS=1.1.1.1 2606:4700:4700::1111|g" /etc/systemd/resolved.conf
-        sed -i "s|#FallbackDNS=.*|FallbackDNS=1.0.0.1 2606:4700:4700::1001|g" /etc/systemd/resolved.conf
+        sed -i "s|#DNS=.*|DNS=1.1.1.1 1.0.0.1 2606:4700:4700::1111 2606:4700:4700::1001|g" /etc/systemd/resolved.conf
     ;;
     "Local")
         sed -i "s|#DNS=.*|DNS=$GATEWAY|g" /etc/systemd/resolved.conf
-        sed -i "s|#FallbackDNS=.*|FallbackDNS=149.112.112.112 2620:fe::9|g" /etc/systemd/resolved.conf
```
enoch85, line 218: Add a check here if the NS can resolve or ask for a new input. This should only happen when choosing "local".

static/static_ip.sh
```diff
@@ -177,15 +180,24 @@ ENTERNS2
...
 
 if [ -n "$NSIP2" ]
 then
-    DNS2="$NSIP2"
+    if [[ "none" == "$NSIP2" || "'none'" == "$NSIP2" ]]
+    then
+        DNS2=
+    else
+        DNSs="$NSIP1,$NSIP2"
+        DNS2="$NSIP2"
```
enoch85, line 199: Is DNS2 really needed? Where is it used?

enoch85:
Just FYI, the default DNS is in the lib file: https://github.com/nextcloud/vm/blob/master/lib.sh#L47-L48

enoch85:
Just tested, the second DNS is never set in Netplan.

Also, I'm not sure why this is needed?

```
DNSs="$DNS1"
if [ -n "$DNS2" ]
then
    DNSs="$DNS1,$DNS2"
fi
```

Since `DNSs` is set *after* those lines - why do you need to set it before? DNS1 and DNS2 are always Quad9 servers as they are hardcoded in lib.sh.

aaaskew:
Sorry, I was still in the process of changing files based on your comments so your comments an hour ago might not have seen all the code.

lib.sh
-----

- The default Quad9 DNS servers are no longer the defaults on a second pass. If you run the statis_ip.sh scripts again, it will pick values from the /etc/systemd/resolved.conf file global settings instead. This also means if you change the defaults in nextcloud_install_production.sh then they will be used instead of Quad9.
- The logic of the test_connection() function has been changed. I managed to return from the function with a non working DNS server when I think the intention of the function was to exit 1. The nested if statements had a logical hole. I have re-written them so it should no longer be possible to end in that way. However I wanted to use the function and not exit but return with error, so I added a parameter to allow me to do this. I could only see this function being used in two places and in both cases I changed it to return error rather than exit on error. Maybe the exit is no longer required?

static_ip.sh
-----

- As the lib.sh defaults could now contain 1 or 2 servers based on content of /etc/systemd/resolved.conf, if the static_ip.sh file is re-run then you need to set the initial state if DNSs to be based on 1 or 2 servers also. If you choose not to change the DNS servers then setting DNSs after entering new values will not be hit.
- `DISPLAY_DNS2` is needed as it is possible to not have a second DNS server. This sets the visual representation of this to 'none' in the prompts so people see that it is not set. A `DISPLAY_DNS1` is not needed as we should always have 1 DNS server.
- The static_ip.sh script now loops until a success or you specifically ask to quit on failure. I believe this is required as this script is called from nextcloud-startup-script.sh which then goes on to ask about setting up a Wi-Fi networking card.

nextcloud_install_production.sh
-----

- I have changed the logic to loop on setting of DNS settings. It should only allow you to continue if you have picked an Internet DNS server.
- I believe the previous logic would have exited if there was any failure of DNS. This is no longer the case but it could be put back so that it loops bad on Local DNS but quits on bad Internet DNS.


Could you say if you still get: 'Just tested, the second DNS is never set in Netplan.'? I am copying files to a base install of Nextcloud VM and testing and then copying the changes back to Github web interface so I hope I have everything.

aaaskew:
Found the 'else' causing problems for setting a second DNS

enoch85:
#### test_connection.sh

It's not supposed to exit, instead sleep for 300 seconds as it does right now.

#### static_ip.sh

Wifi is only asked for if it's the Home/SME server, not otherwise.

-------

I will test this and let you know.

lib.sh
```diff
@@ -263,6 +284,22 @@ You can always contact us for further support if you wish: https://shop.hanssoni
     fi
 }
 
+# A function to set the systemd-resolved default DNS servers based on the
+# current Internet faceing interface. This is needed for docker interfaces
```
enoch85, line 288: facing faceing -- small typo

lib.sh
```diff
@@ -385,6 +422,7 @@ calculate_max_children() {
 }
 
 test_connection() {
+local non_fatal="$1"
```
enoch85, line 425: exactly what does this do?

lib.sh
```diff
@@ -230,14 +251,14 @@ domain_check_200() {
     install_if_not dnsutils
 
     # Try to resolve the domain with nslookup using $DNS as resolver
-    if nslookup "${1}" $DNS1 >/dev/null 2>&1
+    if nslookup "${1}" 9.9.9.9 >/dev/null 2>&1
```
enoch85, line 254: Why not use the variable here?

nextcloud_install_production.sh
```diff
@@ -198,30 +198,44 @@ esac
...
+        "Cloudflare")
+            sed -i "s|^#\?DNS=.*$|DNS=1.1.1.1 1.0.0.1 2606:4700:4700::1111 2606:4700:4700::1001|g" /etc/systemd/resolved.conf
+        ;;
+        "Local")
+            sed -i "s|^#\?DNS=.*$|DNS=$GATEWAY|g" /etc/systemd/resolved.conf
+        ;;
+        *)
+        ;;
+    esac
+    if test_connection "continue" && network_ok
```
enoch85, line 222: why not use *only* `netowrk_ok` here? Then you don't need to do anything with `test_connection` right?

enoch85: Done

lib.sh
enoch85: not needed now when we don't use test_connection in the install script anymore (the thing you added)

nextcloud_install_production.sh
enoch85: done

enoch85:
OK, so it's getting pretty stable now I'd say. 

Can you please test all the scenarios you can think of and report back if it's working, or fix what's wrong? Also, please remove the changes in `test_connection()` since they're not needed anymore. And third, please answer my questions above.

It wouldn't hurt with some more # comments either. :)

Thanks!

enoch85:
Don't know about the failing travis yet... We can figure that out once everything else is ready to merge.

Tested now and all seems to work. Good job! :+1:

lib.sh
aaaskew: Fixed

lib.sh
aaaskew: I am not sure that test_connection is unused. In fact I did not add it as it was already there is both places. There is a test_connection.sh but also a test_connection() function in lib.sh. The fact that the test_connection change you made in nextcloud_install_production.sh to add "$@" as a parameter would seem to indicate that it is the internal lib.sh function used rather than the external script?

When I tested static_ip.sh just now, it definitely used the lib.sh function. for the static_ip.sh case, I need that test_connection function to return 1 or error and not exit 1 on error. The reason is that I need to loop if the test fails. If I don't loop, then the exit will skip out of the static_ip.sh back into nextcloud-startup-script.sh which is not what we want any longer as we want users to try again with working details (if they can).

So I think that lib.sh and static_ip.sh are correct now but not nextcloud_install_production.sh.

lib.sh
aaaskew: Added INTERNET_DNS variable at the top of lib.sh to distinguish it from DNS1 that might be an internal DNS server. In the places where DNS1 was used in lib.sh, an external DNS server is needed, so we have to use a Quad9 default as we can't say what the changed DNS1 might be.

lib.sh
```diff
@@ -44,6 +44,9 @@ WANIP4=$(curl -s -k -m 5 https://ipv4bot.whatismyipaddress.com)
 [ -n "$LOAD_IP6" ] && WANIP6=$(curl -s -k -m 5 https://ipv6bot.whatismyipaddress.com)
 INTERFACES="/etc/netplan/01-netcfg.yaml"
 GATEWAY=$(ip route | grep default | awk '{print $3}')
+# Internet DNS required when a check needs to be made to a server outside the home/SME
+INTERNET_DNS="9.9.9.9"
```
enoch85, line 48: I think `INTERNET_DNS=$DNS` in that case.

enoch85: And this feels very much double code... Why not use DNS only?

aaaskew:
I don't think that backing out all of my changes to the lib.sh test_connection() function is the right thing to do as it was buggy and caused the code that used it to proceed on error or if things where ok!

The code was:
if nolookup ok
  Online!
elif ! nslookup ok
  try again
  if nslookup ok
    Online
  fi
  (no else so if not ok skip to end!) <<< Bug
else
  if ! nslookup ok
    Error and quit the script with exit 1
fi
Online

All of that is re-testing nslookup multiple times with buggy logic.

It is now

if ! nslookup ok
  try again
  if ! nslookup ok
    Error
  fi
fi
Online.

I think we need to keep this new logic, which does not have the hole. The error case is now a return 1 rather than an exit 1 as I need that for static_ip.sh. But this means that the nextcloud_install_production.sh code needs to be changed.

enoch85:
The most important place the `test_connection` code is used is here: https://github.com/nextcloud/vm/blob/master/nextcloud-startup-script.sh#L65-L132

It's done without being a function as users without internet should be able to run it - which is the case 80 times out of 100 since they don't have the same IP or DHCP enabled. It's been working for years, but maybe it was always buggy?

aaaskew:
I have made 2 changes to nextcloud_install_production.sh that I think solve the issues there

For the INTERNET_DNS variable, this is used in the domain_check_200 function. This function specifically needs to test an Internet DNS server to make sure the Internet IP of Nextcloud VM matches Internet DNS of Nextcloud. If DNS1 was used as before, this might now be 192.168.0.1. The local DNS server would respond with the private IP address of the Nextcloud. This is correct for internal access to Nextcloud as it stops hairpin NAT, but it would fail the external domain check. So we need both an INTERNET_DNS for external checks and DNS that might be the same or might be a private DNS server.

enoch85:
OK, great - thanks!

Feels like this is done now. What do you think?

> Can you please test all the scenarios you can think of and report back if it's working, or fix what's wrong?

aaaskew:
I think this looks good. I am worried about there being a test_connection function and test_connection.sh script as that might confuse the unwary. Explecially as one returns exit codes and the other seems to print Connected/Not Connected with exit code 0. Maybe the test_connection.sh shoud at least exit 0/exit 1 to less likley to cause issues depending on how it is called, but that is just being picky.

enoch85:
I just grepped the repo and I don't think the script is used. Please move it to `/old` then I'll merge this PR.

aaaskew:
FYI, I believe I have tested all of the static_ip.sh scenarios and the initial DHCP but not the nextcloud_install_production.sh script. I pulled the code out of this script to test the logic but I am not sure what base that script should be tested on?

enoch85:
> what base that script should be tested on?

It's really just a matter of setting DNS if you're on a very strict network to be able to run the script, or just have a preference in the default DNS used for the VM.

The released version will always have Quad9.

enoch85:
Have you:
- Pulled the internet cable / removed internet NIC from the VM and run the static IP script? 
- Tried without DHCP
- Tried with faulty DNS to start with?
- Enter the wrong interface (like KJBIB9y9y9 instead of ens32) in the netplan config (most common)
- Etc etc?

PR #1328, opened 2020-07-20
allow automatic saving and restoring of app groups

szaimen:
When using the group settings from the appstore, this is needed, because the nextcloud update disables automatically all apps and this setting gets lost while disabling the apps during the update.. 

So we store that information in an array and restore this setting after the update.

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -573,7 +573,7 @@ do
     then
         occ_command_no_check app:enable $app
     fi
-    occ_command_no_check config:app:set $app enabled --value="${APPSTORAGE[$app]}"
+        occ_command_no_check config:app:set $app enabled --value="${APPSTORAGE[$app]}"
 done
 
 # Remove header for Nextcloud 14 (already in .htaccess)
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -401,7 +401,7 @@ ENABLED_APPS="$(occ_command app:list | sed '/Enabled/d' | sed '/Disabled/,$d' | 
 declare -Ag APPSTORAGE
 for app in $ENABLED_APPS
 do
-    RESULT="$(occ_command config:app:get $app enabled)"
+    RESULT="$(occ_command config:app:get "$app" enabled)"
     if [ "$RESULT" != "yes" ]
     then
         APPSTORAGE[$app]="$RESULT"
@@ -569,11 +569,11 @@ run_script STATIC recover_apps
 print_text_in_color "$ICyan" "Restoring all groups for enabled apps. This can take a while..."
 for app in "${!APPSTORAGE[@]}"
 do 
-    if ! is_app_enabled $app
+    if ! is_app_enabled "$app"
     then
-        occ_command_no_check app:enable $app
+        occ_command_no_check app:enable "$app"
     fi
-    occ_command_no_check config:app:set $app enabled --value="${APPSTORAGE[$app]}"
+    occ_command_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
 done
 
 # Remove header for Nextcloud 14 (already in .htaccess)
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 I've successfully tested updating a test VM to 19.0.1 using this revision of the update script.
So this is ready for review.

nextcloud_update.sh
```diff
@@ -395,6 +395,19 @@ fi
...
 countdown "Backing up files and upgrading to Nextcloud $NCVERSION in 10 seconds... Press CTRL+C to abort." "10"
 
+# Backup app groups
+print_text_in_color "$ICyan" "Getting and backing up all groups for enabled apps for later. This can take a while..."
+ENABLED_APPS="$(occ_command app:list | sed '/Enabled/d' | sed '/Disabled/,$d' | sed 's|^  - ||g' | sed 's|:.*||')"
+declare -Ag APPSTORAGE
+for app in $ENABLED_APPS
+do
+    RESULT="$(occ_command config:app:get "$app" enabled)"
+    if [ "$RESULT" != "yes" ]
```
szaimen, line 405: BTW: We could actually remove that check to save all information inside the array. Also for which apps are disabled and restore those values after the upgrade altogether, since I've already seen, that apps get enabled after an upgrade, although they were disabled before the update.

nextcloud_update.sh
```diff
@@ -552,6 +565,17 @@ fi
...
 run_script STATIC recover_apps
 
+# Restore app groups
+print_text_in_color "$ICyan" "Restoring all groups for enabled apps. This can take a while..."
+for app in "${!APPSTORAGE[@]}"
+do 
+    if ! is_app_enabled "$app"
+    then
+        occ_command_no_check app:enable "$app"
+    fi
```
szaimen, line 575: as a note to myself: if storing all app configuration, this is needed to be removed.

enoch85:
This is only when enabling preview generator, right?

szaimen:
> This is only when enabling preview generator, right?

No, this doesnt have anything to do with the preview generator.


All apps have currently three possible values inside the enabled variable:
1. the app is enabled:
image
image

2. the app is disabled:
image
image

3. the app is enabled and only allowed for specific groups:
image
image

The problem that I am trying to solve is to save the group setting for later to automatically reapply this settings, because when updating the server, all apps get disabled and the group information gets lost.

Hope this explains a bit better what this PR is doing

enoch85:
That the setting is actually lost in the first place would I consider as a bug. I mean shouldn't it be saved in DB so that it's there when you activate the app again?

I think it would be better to actually report this as a bug in the server repo. Don't you think?

szaimen:
> I think it would be better to actually report this as a bug in the server repo. Don't you think?

There is one already existing, but didn't get much attention:
nextcloud/server#4538
Because of that I thought about fixing it this way.

szaimen:
BTW: I am unsure about this line: https://github.com/nextcloud/vm/blob/19486feab1ee1b705393599d78f8cc159ab455e8/static/recover_apps.py#L16
I think this is producing the error, that if I had an app installed but disabled, the app gets automatically enabled again after an update although it was disabled before.

szaimen:
> BTW: I am unsure about this line:
> 
> https://github.com/nextcloud/vm/blob/19486feab1ee1b705393599d78f8cc159ab455e8/static/recover_apps.py#L16
> 
> 
> I think this is producing the error, that if I had an app installed but disabled, the app gets automatically enabled again after an update although it was disabled before.

So what about doing everything in one place?
Please see this PR: #1329

enoch85:
> I think this is producing the error, that if I had an app installed but disabled, the app gets automatically enabled again after an update although it was disabled before.

No that's not the logic. 

It compare Shipped Apps with enabled apps, and only enables those who aren't shipped and exits in ...`/apps`

enoch85:
Please delete the branch if not in use. Thanks.

PR #1329, opened 2020-07-20
allow automatic backup and restore of app status

szaimen:
continuation of #1328

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -401,7 +401,7 @@ ENABLED_APPS="$(occ_command app:list | sed '/Enabled/d' | sed '/Disabled/d' | se
 declare -Ag APPSTORAGE
 for app in $ENABLED_APPS
 do
-    APPSTORAGE[$app]="$(occ_command config:app:get '$app' enabled)"
+    APPSTORAGE[$app]=$(occ_command config:app:get "$app" enabled)
 done
 
 # Stop Apache2
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -398,11 +398,17 @@ countdown "Backing up files and upgrading to Nextcloud $NCVERSION in 10 seconds.
 # Backup app status
 print_text_in_color "$ICyan" "Getting and backing up the status of apps for later. This can take a while..."
 NC_APPS="$(occ_command app:list | awk '{print$2}' | tr -d ':' | sed '/^$/d')"
-declare -Ag APPSTORAGE
-for app in $NC_APPS
-do
-    APPSTORAGE[$app]=$(occ_command_no_check config:app:get "$app" enabled)
-done
+if [ -z "$NC_APPS" ]
+then
+    print_text_in_color "$IRed" "No apps detected, aborting export of app status... Please report this issue to $ISSUES"
+    APPSTORAGE="no-export-done"
+else
+    declare -Ag APPSTORAGE
+    for app in $NC_APPS
+    do
+        APPSTORAGE[$app]=$(occ_command_no_check config:app:get "$app" enabled)
+    done
+fi
 
 # Stop Apache2
 check_command systemctl stop apache2.service
@@ -562,14 +568,17 @@ fi
 run_script STATIC recover_apps
 
 # Restore app status
-print_text_in_color "$ICyan" "Restoring the status of apps. This can take a while..."
-for app in "${!APPSTORAGE[@]}"
-do 
-    if [ -n "${APPSTORAGE[$app]}" ]
-    then
-        occ_command_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
-    fi
-done
+if [ "$APPSTORAGE" != "no-export-done" ]
+then
+    print_text_in_color "$ICyan" "Restoring the status of apps. This can take a while..."
+    for app in "${!APPSTORAGE[@]}"
+    do 
+        if [ -n "${APPSTORAGE[$app]}" ]
+        then
+            occ_command_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+        fi
+    done
+fi
 
 # Remove header for Nextcloud 14 (already in .htaccess)
 if [ -f /etc/apache2/sites-available/"$(hostname -f)".conf ]
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -575,7 +575,19 @@ then
     do 
         if [ -n "${APPSTORAGE[$app]}" ]
         then
-            occ_command_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+            if [ "${APPSTORAGE[$app]}" = "yes" ]
+            then
+                occ_command_no_check app:enable "$app"
+            elif [ "${APPSTORAGE[$app]}" = "no" ]
+            then
+                occ_command_no_check app:disable "$app"
+            elif echo "${APPSTORAGE[deck]}" | grep -q "^\[\".*\"\]$"
+            then
+                if is_app_enabled "$app"
+                then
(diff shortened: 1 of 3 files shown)


szaimen:
@enoch85 I've successfully updated a test VM with this PR to 19.0.1 so this should be ready to review.

nextcloud_update.sh
```diff
@@ -395,6 +395,15 @@ fi
 
 countdown "Backing up files and upgrading to Nextcloud $NCVERSION in 10 seconds... Press CTRL+C to abort." "10"
 
+# Backup app status
+print_text_in_color "$ICyan" "Getting and backing up the status of apps for later. This can take a while..."
+NC_APPS="$(occ_command app:list | sed '/Enabled/d' | sed '/Disabled/d' | sed 's|^  - ||g' | sed 's|:.*||')"
```
enoch85, line 400: That `sed` looks very "seddy". :)

Jokes aside, use this instead: `app:list | sed '/Enabled/d' | sed '/Disabled/d' | cut -d "-" -f2 | cut -d ":" -f1 | tr -d " "`

enoch85: This is even cleaner: `app:list | awk '{print$2}' | tr -d ":" | sed '/^\s*$/d'`

enoch85: And even more clean: `app:list | awk '{print$2}' | tr -d ":" | sed '/^$/d'´

szaimen: I will try out your suggestion later when I am back home.

enoch85: `awk` prints the second "thing", `tr` removes the : and `sed` removes the whitespace.

You could use `cut` instead of `tr` for readability as well. Your choice.

enoch85: Prune for bugs if the output changes though. Maybe ad a check for that?

szaimen: I will see if a check is possible

enoch85: A simple one could be:

```
if [ -z "$NC_APPS" ]
then
    print_text_in_color "$IRed" "No apps detected, aborting export of settings... Please report this issue to $ISSUES"

------SKIP EXPORT
fi

Comment (szaimen) [reply nextcloud_update.sh]: > And even more clean: `app:list | awk '{print$2}' | tr -d ":" | sed '/^$/d'`

This has worked in my testing and was addressed.

Comment (szaimen) [reply nextcloud_update.sh]: > A simple one could be:
> 
> ```
> if [ -z "$NC_APPS" ]
> then
>     print_text_in_color "$IRed" "No apps detected, aborting export of settings... Please report this issue to $ISSUES"
> 
> ------SKIP EXPORT
> $APPSTORAGE=no-export-done
> fi
> ```
> 
> Then with the import:
> 
> ```
> if [ $APPSTORAGE = no-export-done ]
> then
> .................... DON'T IMPORT
> fi
> ```

this was addressed in the latest commit

**Comment by szaimen** [comment]
I've already tested the latest changes again by updating a test-vm to 19.0.1, again. So everything should work.

**Comment by szaimen** [comment]
@enoch85 is there something else you want to see me getting addressed?

**Comment by enoch85** [comment]
> is there anything else you want to see getting addressed?

Not really, just not sure if I want this in the update script since it's the most sensitive script of the all. I need to test this first, and I don't have time right now. Sorry.

**Comment by szaimen** [comment]
> Not really, just not sure if I want this in the update script since it's the most sensitive script of the all. I need to test this first, and I don't have time right now. Sorry.

Thanks for the answer!
Just take your time and let me know.

**Comment by szaimen** [comment]
Need to add a check for yes and no: you can set this value also if an app is not supported on a new NC version.

**Comment by enoch85** [comment]
The thing with the update script is that it should run without any questions, since we do automatic backups.

I like the idea of yes and no, but it doesn't fit here.

**Comment by szaimen** [comment]
> The thing with the update script is that it should run without any questions, since we do automatic backups.
> 
> I like the idea of yes and no, but it doesn't fit here.

I think you misunderstood something: in the value enabled is stored the app status: 
The status can be "yes" for enabled, "no" for disabled or '["group","othergroup"] for group limitations
So I am not asking the admin for input

**Comment by szaimen** [comment]
This is what I came up with:
https://github.com/nextcloud/vm/pull/1329/files#diff-6106caeae41ab8eab27ceea759b9a1daR578-R590

**Comment by szaimen** [comment]
btw: whilte testing, I found out that is_app_enabled is not completely reliable. So I changed it
https://github.com/nextcloud/vm/pull/1329/files#diff-68f7bc5e839823a7568557cf9e87f4e2R788

**Comment by enoch85** [comment]
> btw: whilte testing, I found out that is_app_enabled is not completely reliable. So I changed it

Please make that a separate PR. Also, how did you find out it was not reliable, and what does your fix actually fix?

Thanks!

**Comment by szaimen** [comment]
> Please make that a separate PR. Also, how did you find out it was not reliable, and what does your fix actually fix?

problem is the following:

you have the contactsinteraction app enabled and the contacts app disabled.
With the old is_app_enabled, a check for is_app_enabled contacts would return true, although it is disabled since we don't check for the whole appname but just check if it appears anywhere. Since the contactsinteractionapp contains the string contacts, it returns true. we should fix that.

**Comment by enoch85** [comment]
I would like this PR to only do the export and import if there are any special values set. 

That way we minimize the bugs since the normal behaviour is that no settings are made.

File: nextcloud_update.sh [hunk 2]
```diff
@@ -552,6 +567,31 @@ fi
...
+if [ "${APPSTORAGE[0]}" != "no-export-done" ]
+then
+    print_text_in_color "$ICyan" "Restoring the status of apps. This can take a while..."
+    for app in "${!APPSTORAGE[@]}"
+    do 
+        if [ -n "${APPSTORAGE[$app]}" ]
+        then
+            if [ "${APPSTORAGE[$app]}" = "yes" ]
+            then
+                occ_command_no_check app:enable "$app"
```
enoch85, line 580: Why do you enable or disbale apps? The check should only match already enabled apps, right?

enoch85: The result for me is:

```
root@nextcloud500gb:~# bash test.sh 
Getting and backing up the status of apps for later. This can take a while...
import...
nextcloud_announcements already enabled
firstrunwizard already enabled
settings already enabled
theming already enabled
activity already enabled
contactsinteraction already enabled
files_videoplayer already enabled
sharebymail already enabled
photos already enabled
comments already enabled
support already enabled
password_policy already enabled
systemtags already enabled
provisioning_api already enabled
federation already enabled
files_pdfviewer already enabled
extract already enabled
dav already enabled
files_sharing already enabled
text already enabled
serverinfo already enabled
mail already enabled
workflowengine already enabled
files_trashbin already enabled
calendar already enabled
files already enabled
files_versions already enabled
lookup_server_connector already enabled
files_rightclick already enabled
accessibility already enabled
oauth2 already enabled
twofactor_backupcodes already enabled
admin_audit already enabled
logreader already enabled
deck already enabled
No such app enabled: contacts
survey_client already enabled
federatedfilesharing already enabled
groupfolders already enabled
updatenotification already enabled
viewer already enabled
recommendations already enabled
privacy already enabled
notifications already enabled
cloud_federation_api already enabled
root@nextcloud500gb:~# 
```

Doesn't look good, imho.

enoch85:
I'm fine with this. Any comments? @szaimen

szaimen:
> I'm fine with this. Any comments? @szaimen

No I am also fine with this. 👍

enoch85:
@szaimen Please delete all your unmerged branches that are not in use.

Thanks!

szaimen:
Done 👍
BTW: what is the matter with the postgresql branch?

enoch85:
>  what is the matter with the postgresql branch?

It's there for backwards compatibility. https://github.com/nextcloud/vm/blob/postgresql/nextcloud_update.sh

szaimen:
Ah okay 👍

PR #1334, opened 2020-07-25
Re-enable FullTextSearch app

### LINKED ISSUE #1324: Can't install FullTextSearch on with latest NC (19.01)
## Steps to reproduce
1. Run main_menu, select additional apps, then FullTextSearch
2. Notice the failure:  ```Error: App "Full text search" cannot be installed because it is not compatible with this version of the server.```

## Expected behaviour
FullTextSearch should install. If FullTextSearch is known to be incompatible with my currently installed NC, `fulltextsearch.sh` should check for that before pulling docker images and trying to install it only to fail. Maybe a check similar to `lowest_compatible_nc`?

## Actual behaviour
Docker images are pulled, containers started, many many log lines go by, only to be given the error message in the `Steps to Reproduce` above

## Server configuration

**Nextcloud server version:**  Installed with NC 19.0, used upgrade scripts to upgrade 19.01. Same failure on both.

**Server OS** Default Ubuntu in OVA

**How did you install the VM?** Downloaded OVA

## Network
**Do you use DHCP?** Yes

**Is port 80 and/or 443 open?** Yes

## Logs / Screenshots
https://0bin.net/paste/2bIQalxLNZQ1LFaG#L77nH6IWFVp+Vz-A2W9ri6h8/g73v5Aw4mk7yJ2gGyC

kellybyrd:
NC 19 is supported as of the most recent FTS, FTS-Files,
and  FTS-ES releases.

TESTING DONE:
From a NC 19.01 VM, with a failed FTS install from an
earlier version of the install script, I ran this install.
After uninstall, re-install, I saw it indexing files.
After indexing, I could use the NC Web UI to search for
words I knew were in some of my PDFs.

NOTE:
I can't get search results from external files on CIFS
shares, but I believe this is a separate issue in FTS.

Resolves #1324

### DIFF
--- a/apps/fulltextsearch.sh
+++ b/apps/fulltextsearch.sh
@@ -23,19 +23,6 @@ root_check
 # Nextcloud 19 is required.
 lowest_compatible_nc 19
 
-# Won't work info
-msg_box "Full Text Search needs to be updated by the developer for it to work. 
-Currently there are issues that blocks it from working in NC 19.
-
-You can find the repos here:
-https://github.com/nextcloud/fulltextsearch
-https://github.com/nextcloud/fulltextsearch_elasticsearch
-https://github.com/nextcloud/files_fulltextsearch
-https://github.com/nextcloud/bookmarks_fulltextsearch
-
-When this is fixed, we will allow installation again."
-exit 1
-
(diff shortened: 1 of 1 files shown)


enoch85:
Thanks for testing and for the PR!

PR #1337, opened 2020-07-27
performance improvements to smbmount

szaimen:
In my testing provide these small changes huge performance upgrades especially when playing videos and scrolling through them.

### DIFF
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -107,7 +107,7 @@ do
     if ! grep -q "$SMBSHARES/$count " /etc/fstab
     then 
         # Write to /etc/fstab and mount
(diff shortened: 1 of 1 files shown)


enoch85:
Why change from the latest SMB version?

SMB in general are known to have several flaws, at least the older versions. So, speed or security - what is most important?

szaimen:
> Why change from the latest SMB version?
> 
> SMB in general are known to have several flaws, at least the older versions. So, speed or security - what is most important?

I've remove the vers tag since it automatically uses the highest version that both client and server support.
e.g. On my server this leads the removal to the usage of 3.1.1 instead of 3.0
I added the vers=3.0 in the begginning, since I wasn't sure if Linux would automatically use the highest available version, which it does.

PR #1349, opened 2020-08-09
Make Webmin optional and move some scripts

### LINKED ISSUE #1290: Scripts that should move to "addons"
- automatic_updates.sh
- cookielifetime.sh
- locales.sh
- locate_mirror.sh
- prune_zfs_snaphots.sh
- zfs-prune-snapshots.sh
- redis-server-ubuntu.sh
- security.sh

@szaimen Makes sense?

enoch85:
Fixes #1290 

cc @szaimen Feel free to continue.

We need to have backwards compatibility for some of the script that are run in the startup script. The rest is fine as they are not static in the releases.

I've been thinking if we could make the backwards compatibility a function, but I haven't figured out how yet... Maybe you have some ideas?


enoch85:
I think this is a good structure:

- [ ] Menus
- [ ] Configuration
- [ ] Network
- [ ] Addons

enoch85:
Scripts that can be removed from `/static` in like 3 releases:

```
-rw-r--r--  1 root root 3106 aug  9 17:41 additional_apps.sh
-rw-r--r--  1 root root  527 aug  9 17:52 apps.sh
-rw-r--r--  1 root root  541 aug  9 18:03 configuration.sh
-rw-r--r--  1 root root 1595 aug  9 18:08 main_menu.sh
-rw-r--r--  1 root root  515 aug  9 17:51 menu.sh
-rw-r--r--  1 root root 3140 aug  9 17:56 nextcloud_configuration.sh
-rw-r--r--  1 root root 2936 aug  9 17:41 server_configuration.sh
```
```
-rw-r--r--  1 root root 8115 aug  9 14:16 static_ip.sh
-rw-r--r--  1 root root  866 aug  9 18:11 trusted.sh
-rw-r--r--  1 root root 1228 aug  9 14:16 update-config.php
```
I think this would be OK to merge in the current state. Need to test first though.

enoch85:
`apps.sh` is not needed as it's not used anywhere. Better to use `menu.sh`

menu/server_configuration.sh
```diff
@@ -0,0 +1,91 @@
...
+"Activate TLS" "(Enable HTTPS with Let's Encrypt)" ON \
+"Security" "(Add extra security based on this http://goo.gl/gEJHi7)" OFF \
+"Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
+"Disk Check" "(Check for S.M.A.R.T errors on your disks every week on Mondays)" OFF \
+"Automatic updates" "(Automatically update your server every week on Sundays)" OFF 3>&1 1>&2 2>&3)
+
+case "$choice" in
+    *"Disk Check"*)
+        clear
+        run_script ADDONS smartctl
```
enoch85, line 45: it's correct in git cli, but wrong here...

enoch85: fixed now, probably a caching issue

enoch85:
@szaimen You don't need to test everything (would be super nice if you did though), but just have a quick look if you see something that's off.

I'm pretty confident this will be OK to merge as all the scrips that are run from the startup script are kept in their original folder.

nextcloud-startup-script.sh
```diff
@@ -239,15 +239,21 @@ print_text_in_color "$ICyan" "Getting scripts from GitHub to be able to run the
...
+download_script STATIC trusted
+# download_script NETWORK trusted
 download_script STATIC menu
+# download_script MENU menu
 download_script STATIC server_configuration
+# download_script MENU server_configuration
 download_script STATIC nextcloud_configuration
+# download_script MENU nextcloud_configuration
 download_script APP additional_apps
+# download_script MENU additional_apps
```
szaimen, line 256: Wouldn't it make sense to uncomment the new version and comment the old?

szaimen:
> I've been thinking if we could make the backwards compatibility a function, but I haven't figured out how yet... Maybe you have some ideas?

Actually yes:
Since the only scripts, that have to remain backwards compatible are the menu scripts, I would add this to the beginning of the download_script function:

```
    if [ "$2" = "STATIC" ]
    then
        if [ "$3" = "main_menu" ]
        then
            download_script MENU main_menu && return
        elif [ "$3" = "server_configuration" ]
        then
            download_script MENU main_menu && return
        elif [ "$3" = "nextcloud_configuration" ]
        then
            download_script MENU main_menu && return
        fi
    elif [ "$2" = "APPS" ] && [ "$3" "nextcloud_configuration" ]
    then
        download_script MENU additional_apps && return
    fi
```

the complete download_script function would then look like this:
```
download_script() {
    if [ "$2" = "STATIC" ]
    then
        if [ "$3" = "main_menu" ]
        then
            download_script MENU main_menu && return
        elif [ "$3" = "server_configuration" ]
        then
            download_script MENU main_menu && return
        elif [ "$3" = "nextcloud_configuration" ]
        then
            download_script MENU main_menu && return
        fi
    elif [ "$2" = "APPS" ] && [ "$3" "nextcloud_configuration" ]
    then
        download_script MENU additional_apps && return
    fi
    if ! { curl_to_dir "${!1}" "${2}.sh" "$SCRIPTS" || curl_to_dir "${!1}" "${2}.php" "$SCRIPTS" || curl_to_dir "${!1}" "${2}.py" "$SCRIPTS"; }
    then
        exit 1
    fi
}

```
But haven't tested this...

nextcloud-startup-script.sh
enoch85: Yes, it would. :smile: Or maybe the functions wants to be nostalgic about where they came from? Haha :)

enoch85:
Why would only the menu scripts have to be backwards compatible?

nextcloud-startup-script.sh
enoch85: Duuh, yes it would. Thought that I was smart here adding backwards compatibility.

Thing is that the download function exits if the script doesn't exist, so I've got several emails regarding test_connection and activate-tls since those are "missing" in the old VMs (the startup script).

That was what I was trying to solve here, thinking that we could use the old references just until we make the new release.

But you're right, it doesn't make any difference right now. What actually matter is  that the scripts are at the old locations.

enoch85:
@szaimen 

Aah, I get what you mean now. But rather than bloating the function, we can just keep double copies of the scripts like I've done now.

Merging this when travis is fixed.

enoch85:
@szaimen Tested on an old VM and it worked. :fire:

szaimen:
> @szaimen
> 
> Aah, I get what you mean now. But rather than bloating the function, we can just keep double copies of the scripts like I've done now.
> 
> Merging this when travis is fixed.

However you like it more  👍

> Tested on an old VM and it worked. 🔥

Great! 🚀

enoch85:
One thing bugged out, the smartctl script as it uses the old lib file in the released version. I can live with that though.

Issue #1297, opened 2020-06-19
Add option to Bitwarden script to disable registration to prevent unauthorized creation of accounts


e-alfred:
## Steps to reproduce
1. Install Bitwarden through the "additional apps" option
2. Register users

## Expected behaviour
After all necessary users are registered, the script for Bitwarden should allow to disable user creation through the web interface to prevent creation of unauthorized accounts.

## Actual behaviour
Bitwarden by default allows everyone to create an account, especially if it is open on the internet this is an issue.

## Server configuration

**Nextcloud server version:** (see your admin page) 19

**Server OS** (Ubuntu server is default) Ubuntu 20.04

**How did you install the VM?** Script on custom VM

## Network
**Do you use DHCP?** No

**Is port 80 and/or 443 open?** Yes

enoch85:
Thanks for your suggestion. Maybe it's better to bring this up with Bitwarden directly which made the original install script (which the VM runs).

You can find their repo here: https://github.com/bitwarden/server

Leaving this open in case we fix a workaround for it anyway.

e-alfred:
Well, this is a bit of a security risk/nuisance if Bitwarden is accessible over the public internet because everyone could register a user maybe even resulting in a DOS attack of some kind.

A more secure workflow would be this:

1. Admin installs Bitwarden using the script.
2. Admin sets up all users using the Bitwarden web interface.
3. Admin disables registration after setting up all users through the script (should be added to the "Server Configuration" section).
4. Admin can enable/disable registration if a new user should be added to Bitwarden.

e-alfred:
Well, this is a bit of a security risk/nuisance if Bitwarden is accessible over the public internet because everyone could register a user maybe even resulting in a DOS attack of some kind.

A more secure workflow would be this:

1. Admin installs Bitwarden using the script.
2. Admin sets up all users using the Bitwarden web interface.
3. Admin disables registration after setting up all users through the script (should be added to the "Server Configuration" section).
4. Admin can enable/disable registration if a new user should be added to Bitwarden.

The right options should be these: https://github.com/bitwarden/server/issues/631

enoch85:
@e-alfred You're welcome to contribute! :+1:

enoch85:
@e-alfred Please test the PR, it's now ready for review.

Thanks!

PR #1350, opened 2020-08-09
Make user Registration configurable

enoch85:
Fix https://github.com/nextcloud/vm/issues/1297

cc @e-alfred


apps/bitwarden-registration.sh
```diff
@@ -0,0 +1,73 @@
...
+        msg_box "It seems like 'Bitwarden' isn't installed.\n\nYou cannot run this script."
+        exit 1
+    fi
+else
+    msg_box "It seems like 'Bitwarden' isn't installed.\n\nYou cannot run this script."
+    exit 1
+fi
+
+# Yes or No?
+choice=$(whiptail --title "Bitwarden Registration" --checklist "Do you want to disable Bitwarden User Registration?\nSelect by pressing the spacebar\nYou can view this menu later by running 'sudo bash $SCRIPTS/menu.sh'" "$WT_HEIGHT" "$WT_WIDTH" 4 \
```
szaimen, line 42: Wouldn't make a radiollist here more sense?

apps/tmbitwarden.sh
```diff
@@ -84,7 +84,10 @@ check_command ./bitwarden.sh rebuild
 check_command ./bitwarden.sh start
 if check_command ./bitwarden.sh updatedb
 then
-msg_box "Bitwarden was sucessfully installed! Please visit $(grep 'url:' /root/bwdata/config.yml | awk '{print$2}'):8443 to setup your account."
+msg_box "Bitwarden was sucessfully installed! Please visit $(grep 'url:' /root/bwdata/config.yml | awk '{print$2}'):8443 to setup your account.
+
+After the account it setup, please disable user registration by running sudo bash $SCRIPTS/menu.sh and choose:
+Server Configuration --> Bitwarden Registration"
```
szaimen, line 90: I am unsure if it fits that well in server_configuration, since it is configuring the bitwarden app.
Are you sure with that?

static/server_configuration.sh
```diff
@@ -37,9 +37,14 @@ choice=$(whiptail --title "Server configurations" --checklist "Choose what you w
...
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
 "Disk Check" "(Check for S.M.A.R.T errors on your disks every week on Mondays)" OFF \
+"Bitwarden Registration" "(Enable or disable public user registration for Bitwarden)" OFF \
 "Automatic updates" "(Automatically update your server every week on Sundays)" OFF 3>&1 1>&2 2>&3)
 
 case "$choice" in
+    *"Bitwarden Registration"*)
+        clear
+        run_script APP bitwarden-registration
+    ;;&
```
szaimen, line 47: We should think about the best order executing those scripts: I am not sure if when all options get chosen in server_configuration, the Bitwarden Registration gets executed first and the Disk Check second, and so on...

apps/tmbitwarden.sh
szaimen: Also this would be an option before installing any apps if executed during the startup script, isn't that a problem?

enoch85: > I am unsure if it fits that well in server_configuration

OK, so where would you put it?

> Also this would be an option before installing any apps if executed during the startup script, isn't that a problem?

I don't get this question, sorry. I see it as a more polite thing that like "hey, you maybe would like to disable Public User Registration once we're done here"

static/server_configuration.sh
enoch85: Please test it and make proper changes.

And yes, radiolist would fit better. :+1: Was quick and dirty yesterday. :)

apps/tmbitwarden.sh
szaimen: > OK, so where would you put it?

Hm, maybe as an additional option for an apps_script?
or maybe the bitwarden option in additional_apps itself could be a menu, where you can choose between the installer/reinstaller/uninstall script and the bitwarden configuration script?

> I don't get this question, sorry. I see it as a more polite thing that like "hey, you maybe would like to disable Public User Registration once we're done here"

What I mean is, that if you run the startup script, you first see the server_configuration menu with the option to configure bitwarden. The thing is, that you will see the additional_apps menu afterwards where you have the first chance to install bitwarden.

enoch85: True!

Let's change it to the apps menu instead. Do you have time? I'm in the middle of a big migration.

Please also change the script itself to a radiolist as you proposed.

Thanks! :+1:

static/server_configuration.sh
szaimen: So my idea is to use this order here:
```
"Activate TLS" "(Enable HTTPS with Let's Encrypt)" ON \
"Automatic updates" "(Automatically update your server every week on Sundays)" OFF \
"Disk Check" "(Check for S.M.A.R.T errors on your disks every week on Mondays)" OFF \
"Bitwarden Registration" "(Enable or disable public user registration for Bitwarden)" OFF 3>&1 1>&2 2>&3)
```
and execute it in the same order (apart from activate-tls, which should probably get executed always at the end):

```
case "$choice" in
    *"Security"*)
        clear
        run_script ADDONS security
    ;;&
    *"Static IP"*)
        clear
        run_script NETWORK static_ip
    ;;&
    *"Automatic updates"*)
        clear
        run_script ADDONS automatic_updates
    ;;&
    *"Disk Check"*)
        clear
        run_script DISK smartctl
    ;;&
    *"Bitwarden Registration"*)
        clear
        run_script APP bitwarden-registration
    ;;&
    *"Activate TLS"*)
        clear
msg_box "The following script will install a trusted
TLS certificate through Let's Encrypt.
It's recommended to use TLS (https) together with Nextcloud.
Please open port 80 and 443 to this servers IP before you continue.
More information can be found here:
https://www.techandme.se/open-port-80-443/"

        if [[ "yes" == $(ask_yes_or_no "Do you want to install TLS?") ]]
        then
            if [ -f $SCRIPTS/activate-tls.sh ]
            then
                bash $SCRIPTS/activate-tls.sh
            else
                download_script LETS_ENC activate-tls
                bash $SCRIPTS/activate-tls.sh
            fi
        else
            echo
            print_text_in_color "$ICyan" "OK, but if you want to run it later, just type: sudo bash $SCRIPTS/activate-tls.sh"
            any_key "Press any key to continue..."
        fi

        # Just make sure they are gone
        rm -f "$SCRIPTS/test-new-config.sh"
        rm -f "$SCRIPTS/activate-tls.sh"
        clear
    ;;&
    *)
    ;;
esac
```

apps/bitwarden-registration.sh
szaimen: done

apps/tmbitwarden.sh
szaimen: I've changed it to the apps menu and changed the checklist to a radiolist

static/server_configuration.sh
szaimen: I've moved that proposal to https://github.com/nextcloud/vm/pull/1351

apps/bitwarden-registration.sh
```diff
@@ -0,0 +1,73 @@
...
+# Check if Bitwarden is already installed
+print_text_in_color "$ICyan" "Checking if Bitwarden is already installed..."
+if is_docker_running
+then
+    if docker ps -a --format '{{.Names}}' | grep -Eq "bitwarden";
+    then
+        if [ -d /root/bwdata ]
+        then
+            sleep 0.1
+         else
```
szaimen, line 28: instead of sleep 0.1 you could simply use 
```
if [ ! -d /root/bwdata ]
then 
    msg_box "It seems like 'Bitwarden' isn't installed.\n\nYou cannot run this script."
    exit 1
fi
```

enoch85: Duhh, was tired yesterday. Thanks!

enoch85:
Is this approved now? @szaimen

szaimen:
> Is this approved now? @szaimen

Theoratically yes, but haven't had the chance to test the script itself.
Did you?

enoch85:
Let's merge! :D

PR #1352, opened 2020-08-10
restore main_menu to its original place

szaimen:
else you won't be able to start the main_menu on certain installations

I've just encountered this as I wanted to acces the main_menu.

### DIFF
--- a/static/main_menu.sh
+++ b/static/main_menu.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
+
+# shellcheck disable=2034,2059
+true
+# shellcheck source=lib.sh
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
+# Check for errors + debug code and abort if something isn't right
+# 1 = ON
+# 0 = OFF
+DEBUG=0
+debug_mode
+
+# Must be root
+root_check
+
+# Main menu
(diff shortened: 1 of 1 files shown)


enoch85:
> else you won't be able to start the main_menu on certain installations

Well, `main_menu.sh` is never used - we only run `menu.sh` right? I mean `menu.sh` get's the latest script...

szaimen:
> > else you won't be able to start the main_menu on certain installations
> 
> Well, `main_menu.sh` is never used - we only run `menu.sh` right? I mean `menu.sh` get's the latest script...

Yes, but all installations have an already downloaded menu.sh. So if you execute the local file it tries to download the hardcoded /static/main_menu.sh. That is the reason why it fails on my installation...

enoch85:
In each update the latest script is downloaded. Since the last thing that happens in the setup is updating, everyone using the old VM will get the latest script, and everyone running the update script will also get the latest version.

szaimen:
> In each update the latest script is downloaded. Since the last thing that happens in the setup is updating, everyone using the old VM will get the latest script, and everyone running the update script will also get the latest version.
> 
> https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L84

Ah okay but not if you have a running VM and try to access the main_menu without running the update.sh first...

enoch85:
OK, we can merge this - keep it for some versions, and then remove it when we do the other cleanup of leftover scripts for backwards compatibility. Fine with me.

PR #1353, opened 2020-08-10
Update fail2ban

szaimen:
I've just tried to install fail2ban on my productive instance and it tried to search through all my files which are mounted via smb although I have just the standard configuration and hence this is unnecessary and needs a LOT of time.

Since it it is much faster to look for the default config first, I would propose to change it to this method.
Was successfully tested on my system.

### DIFF
--- a/apps/fail2ban.sh
+++ b/apps/fail2ban.sh
@@ -61,23 +61,25 @@ mkdir -p "$VMLOGS"
 print_text_in_color "$ICyan" "Finding nextcloud.log..."
 while :
 do
-    NCLOG=$(find / -type f -name "nextcloud.log" 2> /dev/null)
-    if [ "$NCLOG" != "$VMLOGS/nextcloud.log" ]
+    if [ "$(occ_command_no_check config:system:get logfile)" = "$VMLOGS/nextcloud.log" ]
     then
-        # Might enter here if no OR multiple logs already exist, tidy up any existing logs and set the correct path
-        print_text_in_color "$ICyan" "Unexpected or non-existent logging configuration - deleting any discovered nextcloud.log files and creating a new one at $VMLOGS/nextcloud.log..."
-        xargs rm -f <<< "$NCLOG"
+        break
+    elif [ "$(occ_command_no_check config:system:get logfile)" != "" ]
+    then
         # Set logging
         occ_command config:system:set log_type --value=file
         occ_command config:system:set logfile --value="$VMLOGS/nextcloud.log"
         occ_command config:system:set loglevel --value=2
         touch "$VMLOGS/nextcloud.log"
         chown www-data:www-data "$VMLOGS/nextcloud.log"
+        break
     else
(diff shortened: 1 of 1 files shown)


enoch85:
The reason that we search through the whole system is that for some reason nextcloud.log tends to be logged in the default datadir if found there. That's also the case if it exists in two places, e.g. `/mnt/ncdata/nextcloud.log` and `$VMLOGS/nextcloud.log` - then it will log to `/mnt/ncdata/nextcloud.log` even if it's set in config.php to another path. Strange, but that's what happended during my testing at least.

Also, not to confuse users, we remove everything that's not in the default path set in the script.

szaimen:
> The reason that we search through the whole system is that for some reason nextcloud.log tends to be logged in the default datadir if found there. That's also the case if it exists in two places, e.g. `/mnt/ncdata/nextcloud.log` and `$VMLOGS/nextcloud.log` - then it will log to `/mnt/ncdata/nextcloud.log` even if it's set in config.php to another path. Strange, but that's what happended during my testing at least.
> 
> Also, not to confuse users, we remove everything that's not in the default path set in the script.

This sounds like a bug on the nextcloud side.
What you are saying sounds like it is using the default config always if the file exists, which would be /mnt/ncdata/nextcloud.log
image
```
So I propose just adding 
rm -f $NCDATA/nextcloud.log 
```
to the script.
Which was done with a69ee88

enoch85:
What I'm saying is that we need `xargs rm -f <<< "$NCLOG"` and for that we first need to look for all the places where the log exists. It could be more than two places...

So, sorry, but rather have it done correctly than quick. And yes, I've been thinking about this for months, every time I install fail2ban...

szaimen:
> What I'm saying is that we need `xargs rm -f <<< "$NCLOG"` and for that we first need to look for all the places where the log exists. It could be more than two places...

What you are saying is, that it also uses nextcloud.log files that are not in the default path and not in the configured path.
Is that the case here? 
For me it just sounds like if it is using the default config, if the file exists no matter what you have configured in the config.php.
So my proposal is to remove that file in NCDATA and configure nextcloud to use the file in /var/log/nextcloud
Which shold cover all cases.
Does that make sense to you?

enoch85:
Yes, it makes sense, but (there's always a but) I would like to do the cleanup anyway.

This script should run on any Nextcloud installation (well, almost) and they might have choose to set logging else where. 

It might be picky of me, sure. I know about the `else`, but what if they come from outside and used the update script which created the `/var/log/nextcloud` path or something else we used in the past, then there might just be a log there without a function. To me it's all about not confusing the user and to make it stable.

enoch85:
What we could do though is exluding some standard paths like `/mnt/smb` or similar (in the find command).

enoch85:
Don't close this though. I will think about ut some more.

enoch85:
Actually, just tested this on a NC 15.0.0 VM from December 2018, it works as expected.

Let's merge this.

PR #1361, opened 2020-08-11
first attempt to allow bitwarden on a subdomain

### LINKED ISSUE #1336: Use a Reverse Proxy for Bitwarden and generate certs with our function
Hey, 

The standard bitwarden implementation is quite heavy on ressources.
Do you think it is feasable to add an option to install [bitwarden_rs](https://github.com/dani-garcia/bitwarden_rs) as alternative to the normal installation or as an additional option?
If we could add a vhost and proxy the request like in the the collabora script, it would then also be possible to install it on a subdomain while using the normal ports 80 and 443.
Do you think it makes sense and would you accept a pull-request?
If yes, I would like to work on it.

BTW: where did you get all the information that is needed to make a virtualhost like this: https://github.com/nextcloud/vm/blob/master/old/onlyoffice.sh#L126-L164
It is much more complicated than the short apache examples in the bitwarden_rs wiki:
https://github.com/dani-garcia/bitwarden_rs/wiki/Proxy-examples
Do you think it would be possible to create a working proxy host template here, too?

for later: this is a guide for nginx: https://dennisnotes.com/note/20181112-bitwarden-server/
https://github.com/bitwarden/server/issues/314

szaimen:
attempt to fix #1336

### DIFF
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -185,10 +185,10 @@ sed -i "s|https_port.*|https_port: 5179|g" /root/bwdata/config.yml
 sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /root/bwdata/config.yml
 sed -i 's|http://|https://|g' /root/bwdata/env/global.override.env
 
-# Symlink TLS certificates
-ln -s $CERTFILES/$SUBDOMAIN/fullchain.pem /root/bwdata/ssl/$SUBDOMAIN/certificate.crt
-ln -s $CERTFILES/$SUBDOMAIN/privkey.pem /root/bwdata/ssl/$SUBDOMAIN/privkey.crt
-ln -s $CERTFILES/$SUBDOMAIN/dhparam.pem /root/bwdata/ssl/$SUBDOMAIN/dhparam.pem
+# add certificate files to config.yml
+sed -i "s|^ssl_certificate_path:.*|ssl_certificate_path: $CERTFILES/$SUBDOMAIN/fullchain.pem|g" /root/bwdata/config.yml
+sed -i "s|^ssl_key_path:.*|ssl_key_path: $CERTFILES/$SUBDOMAIN/privkey.pem|g" /root/bwdata/config.yml
+sed -i "s|^ssl_diffie_hellman_path:.*|ssl_diffie_hellman_path: $CERTFILES/$SUBDOMAIN/dhparam.pem|g" /root/bwdata/config.yml
 
 # Build and start
 check_command ./bitwarden.sh rebuild
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -18,7 +18,7 @@ root_check
 
 # Test RAM size (3 GB min) + CPUs (min 2)
 ram_check 3 Bitwarden
-cpu_check 3 Bitwarden
+cpu_check 2 Bitwarden
 
 # Check if Bitwarden is already installed
 print_text_in_color "$ICyan" "Checking if Bitwarden is already installed..."
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -5,7 +5,8 @@
 # shellcheck disable=2034,2059
 true
 # shellcheck source=lib.sh
-. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+BITWARDEN=1 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+unset BITWARDEN
 
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
@@ -62,6 +63,14 @@ To run this script again, execute $SCRIPTS/menu.sh and choose Additional Apps --
     exit
 fi
 
+# Check if $SUBDOMAIN exists and is reachable
+print_text_in_color "$ICyan" "Checking if $SUBDOMAIN exists and is reachable..."
+domain_check_200 "$SUBDOMAIN"
+
+# Check open ports with NMAP
+check_open_port 80 "$SUBDOMAIN"
+check_open_port 443 "$SUBDOMAIN"
+
 msg_box "ATTENTION!
 
 You will see in the next step a setup which is provided by bitwarden.
@@ -77,42 +86,7 @@ And please don't worry: we will create a valid SSL certificate afterwards oursel
 5. You will be asked if you have a SSL certificate which you want to use; Please type in no: 'n'
 6. You will be asked if you want to generate a self-signed SSL certificate; Please type in no: 'n'"
 
-# Install Docker
-install_docker
-install_if_not docker-compose
-
-# Stop Apache to not conflict when LE is run
-check_command systemctl stop apache2.service
-
-# Install Bitwarden 
-install_if_not curl
-cd /root
-curl_to_dir "https://raw.githubusercontent.com/bitwarden/core/master/scripts" "bitwarden.sh" "/root"
-chmod +x /root/bitwarden.sh
-check_command ./bitwarden.sh install
-sed -i "s|http_port.*|http_port: 5178|g" /root/bwdata/config.yml
-sed -i "s|https_port.*|https_port: 5179|g" /root/bwdata/config.yml
-# Get Subdomain from config.yml and change it to https
-SUBDOMAIN=$(grep ^url /root/bwdata/config.yml)
-SUBDOMAIN=${SUBDOMAIN##*url: http://}
-sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /root/bwdata/config.yml
-sed -i 's|http://|https://|g' /root/bwdata/env/global.override.env
-check_command ./bitwarden.sh rebuild
-check_command ./bitwarden.sh start
-check_command ./bitwarden.sh updatedb
-
 # Produce reverse-proxy config and get lets-encrypt certificate
-
-# Curl the lib another time to get the correct HTTPS_CONF
-. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
-
-# Check if $SUBDOMAIN exists and is reachable
-print_text_in_color "$ICyan" "Checking if $SUBDOMAIN exists and is reachable..."
-domain_check_200 "$SUBDOMAIN"
-
-# Check open ports with NMAP
-check_open_port 80 "$SUBDOMAIN"
-check_open_port 443 "$SUBDOMAIN"
 
 # Install Apache2
 install_if_not apache2
@@ -194,6 +168,32 @@ then
 else
     last_fail_tls "$SCRIPTS"/apps/tmbitwarden.sh
 fi
+
+# Install Docker
+install_docker
+install_if_not docker-compose
+
+# Install Bitwarden 
+install_if_not curl
+cd /root
+curl_to_dir "https://raw.githubusercontent.com/bitwarden/core/master/scripts" "bitwarden.sh" "/root"
+export DOMAIN="$SUBDOMAIN"
+chmod +x /root/bitwarden.sh
+check_command ./bitwarden.sh install
+sed -i "s|http_port.*|http_port: 5178|g" /root/bwdata/config.yml
+sed -i "s|https_port.*|https_port: 5179|g" /root/bwdata/config.yml
+sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /root/bwdata/config.yml
+sed -i 's|http://|https://|g' /root/bwdata/env/global.override.env
+
+# Symlink TLS certificates
+ln -s $CERTFILES/$SUBDOMAIN/fullchain.pem /root/bwdata/ssl/$SUBDOMAIN/certificate.crt
+ln -s $CERTFILES/$SUBDOMAIN/privkey.pem /root/bwdata/ssl/$SUBDOMAIN/privkey.crt
+ln -s $CERTFILES/$SUBDOMAIN/dhparam.pem /root/bwdata/ssl/$SUBDOMAIN/dhparam.pem
+
+# Build and start
+check_command ./bitwarden.sh rebuild
+check_command ./bitwarden.sh start
+check_command ./bitwarden.sh updatedb
 
 # Add prune command
 {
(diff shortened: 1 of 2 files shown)


szaimen:
@enoch85 I could need some feedback on this. Do you have some suggestions?
I just copied the reverse proxy file from the onlyoffice contaner and made some small changes to it (changed the port and ipaddress)

enoch85:
> @enoch85 I could need some feedback on this. Do you have some suggestions?
I just copied the reverse proxy file from the onlyoffice contaner and made some small changes to it (changed the port and ipaddress)

I don't have any suggestions now I'm afraid. What I would do is setup a VPS/VM and test, debug, test, debug until it's working. 

I don't have time exactly right now as I'm in a middle of another migration (ownCloud --> Nextcloud - yay) but I'll see if I get some time over later tonight.

enoch85:
Reading their docs again, it might be a better idea to use their Let's Encrypt after all?

Basically there are two options:
1. Use their LE
2. Use our LE, but symlink the certs folder to theirs: https://bitwarden.com/help/article/install-on-premise/#install-bitwarden

> If you already have your own SSL certificate you can place the necessary files in the ./bwdata/ssl/your.domain.com directory. File paths for certificate assets are configurable from the ./bwdata/config.yml file if you wish to change the default paths generated by the installer. 

The we would only have to solve the proxy to be able to use 443 and 80.

szaimen:
> I don't have time exactly right now as I'm in a middle of another migration (ownCloud --> Nextcloud - yay) but I'll see if I get some time over later tonight.

Gladly the installation has already worked one time on a local testmachine and I was able to access bitwarden under the entered subdomain. But I don't think that it is already stable enough push it to master...

szaimen:
> 1. Use their LE

I don't think that this is possible since it would need to run on port 443 and 80...

szaimen:
> 2\. Use our LE, but symlink the certs folder to theirs: https://bitwarden.com/help/article/install-on-premise/#install-bitwarden
> 
> > If you already have your own SSL certificate you can place the necessary files in the ./bwdata/ssl/your.domain.com directory. File paths for certificate assets are configurable from the ./bwdata/config.yml file if you wish to change the default paths generated by the installer.
> 
> The we would only have to solve the proxy to be able to use 443 and 80.

Actually, I think there is a 3rd way: I've found it here:
https://dennisnotes.com/note/20181112-bitwarden-server/
and followed it except that I configure the reverse proxy in the end.

In this solution, I am running bitwarden locally without certificate but apache2 proxies the result and also has access to the certificate. You should probably just try the script yourself ;)

enoch85:
Actually, it says in their docs that it's better to generate TLS and all that before their script is run, so I would reverse the order here.

enoch85:
> In this solution, I am running bitwarden locally without certificate but apache2 proxies the result and also has access to the certificate. You should probably just try the script yourself ;)

Haha yeah sure.

That guide seems great! Would it be possible to stay on apache you think? I have nothing against Nginx, but I don't like the idea of mixing two webservers - also we use Apache everywhere else here.

enoch85:
While were add it, this is what I got from a user:

```
You will want to configure the Bitwarden Server to use a `bitwarden` service account. $USER=bitwarden You will want to have your installation owned by the bitwarden service account, and you should be logged in as bitwarden. After those are verified, you will want to make sure the UID and GID in the /bwdata/env/uid.env file are set to your bitwarden service account id numbers in Linux. When using the bitwarden service account you will also need to follow these steps:

sudo groupadd docker

sudo usermod -aG docker $USER

sudo vi bitwarden.service

[Unit]
Description=Bitwarden
Requires=docker.service
After=docker.service

[Service]
Type=oneshot
User=bitwarden
Group=bitwarden
RemainAfterExit=true

[Install]
WantedBy=multi-user.target

systemctl daemon-reload

```

He said he had talked to Bitwarden support.

szaimen:
> Actually, it says in their docs that it's better to generate TLS and all that before their script is run, so I would reverse the order here.

We could of course do it this way, but then the user would need to enter the domain twice. I just wanted to make it as simple as possible. (which is the way I came up with)

szaimen:
> That guide seems great! Would it be possible to stay on apache you think? I have nothing against Nginx, but I don't like the idea of mixing two webservers - also we use Apache everywhere else here.

I haven't used nginx for this, just used the onlyoffice reverse proxy for bitwarden and at least the log in screen comes up when I access the site now.

szaimen:
> While were add it, this is what I got from a user:
> 
> ```
> You will want to configure the Bitwarden Server to use a `bitwarden` service account. $USER=bitwarden You will want to have your installation owned by the bitwarden service account, and you should be logged in as bitwarden. After those are verified, you will want to make sure the UID and GID in the /bwdata/env/uid.env file are set to your bitwarden service account id numbers in Linux. When using the bitwarden service account you will also need to follow these steps:
> 
> 1. Make sure the docker group has been created.
> sudo groupadd docker
> 
> 2. Add the bitwarden account to the docker group
> sudo usermod -aG docker $USER
> 
> 3. Create the bitwarden service file (may want to store it with your bitwarden installation)
> sudo vi bitwarden.service
> 
> [Unit]
> Description=Bitwarden
> Requires=docker.service
> After=docker.service
> 
> [Service]
> Type=oneshot
> User=bitwarden
> Group=bitwarden
> ExecStart={INSTALL_DIR}/bitwarden.sh start
> RemainAfterExit=true
> ExecStop= {INSTALL_DIR}/bitwarden.sh stop
> 
> [Install]
> WantedBy=multi-user.target
> 
> 4. Copy the bitwarden service file to systemd.
> sudo cp bitwarden.service /etc/systemd/system/bitwarden.service
> 
> 5. Set permissions on bitwarden service file under systemd.
> sudo chmod 644 /etc/systemd/system/bitwarden.service
> 
> 6. Optional (reload for testing)
> systemctl daemon-reload
> 
> 7. Add service to start with system boot.
> sudo systemctl enable bitwarden.service
> ```
> 
> He said he had talked to Bitwarden support.

What does that mean? Why is it necessary?

szaimen:
> I haven't used nginx for this, just used the onlyoffice reverse proxy for bitwarden and at least the log in screen comes up when I access the site now.

But actually here is another guide: with an obviusly working configuration:
bitwarden/server#314
Maybe we should think about implementing this.

Also although I can reach the welcome and login site, I cannot create an account, which is hopefully because I am running the test-vm with only 1.5 GB RAM which is probably just not enough.

enoch85:
> We could of course do it this way, but then the user would need to enter the domain twice. I just wanted to make it as simple as possible. (which is the way I came up with)

Sure, let's do it like that.

> What does that mean? Why is it necessary?

Basically, setup an unprivileged user for Bitwarden and don't put everything in root.

enoch85:
I'm setting up a test VPS as we speak.

szaimen:
> > What does that mean? Why is it necessary?
> 
> Basically, setup an unprivileged user for Bitwarden and don't put everything in root.

ah okay, sound good 👍

szaimen:
> I'm setting up a test VPS as we speak.

Nice! 👍

szaimen:
> Actually, I think there is a 3rd way: I've found it here:
> https://dennisnotes.com/note/20181112-bitwarden-server/
> and followed it except that I configure the reverse proxy in the end. (and use apache2 instead of nginx)

Actually, this guide is not very clear whether you should enter the domain you want to use during the setup of bitwarden or just localhost. So I tried the domain. Which seems to kind of work...

enoch85:
I looked into how bitwarden.sh works, and basically it just fetches https://github.com/bitwarden/server/blob/master/scripts/run.sh.

We could make it a bit advanced and:
1. Download the run.sh script to our repo
2. Edit the script to suit our needs, e.g. remove all the questions in the beginning and only ask for domain, and/or export the $SUBDOMAIN to $DOMAIN and use that in run.sh...

I will experiment a bit with this...

szaimen:
> I looked into how bitwarden.sh works, and basically it just fetches https://github.com/bitwarden/server/blob/master/scripts/run.sh.
> 
> We could make it a bit advanced and:
> 
> 1. Download the run.sh script to our repo
> 2. Edit the script to suit our needs, e.g. remove all the questions in the beginning and only ask for domain, and/or export the $SUBDOMAIN to $DOMAIN and use that in run.sh...
> 
> I will experiment a bit with this...

okay sounds good! I thought about this, too but I encountered the problem, that all questions that you can see below, starting at the installation id, aren't inside the run script but inside other files, that are probably much more complicated to change...
image

szaimen:
BTW: have you tried the script on an appropriately equipped VM? Can you create an account afterwards?

enoch85:
OK, so we can export SUBDOMAIN as DOMAIN and make it work with the internal TLS like this:

```
COLLABORA_INSTALL=1 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
unset COLLABORA_INSTALL

export -p DOMAIN="$SUBDOMAIN"
check_command /root/bitwarden.sh install
unset DOMAIN
```

But, to do this we have to put this at line 61 in `bitwarden.sh`: `sed -i "s|read DOMAIN||g" $SCRIPTS_DIR/run.sh` so that it doesn't read the DOMAIN name, but uses the DOMAIN variable from what's entered in SUBDOMAIN. Very hacky, I know.

Even if we would get it to work, the result would be that the user don't have to enter the domain name twice.

I'll see if there's a better solution. Please help here.

enoch85:
So I'm close to something here. If I can get the symlink to work properly we can use our own certs in the setup and skip a few questions.

Downside is that you still have to enter the domain twice, but I'll work on that too.

enoch85:
I give up. It's better to do it your way with all the questions. Please revert.

Conclusion: You can't symlink into the docker from the real Let's Encrypt path, it doesn't find the files. You have to copy real files in the .`/ssl` dir otherwise it won't work. But, that won't work either as the files are renewd every time LE renews.

enoch85:
JFTR:

```
COLLABORA_INSTALL=1 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
unset COLLABORA_INSTALL

WORKS
cp -f $CERTFILES/$SUBDOMAIN/fullchain.pem /root/bwdata/ssl/$SUBDOMAIN/certificate.crt
cp -f $CERTFILES/$SUBDOMAIN/privkey.pem /root/bwdata/ssl/$SUBDOMAIN/private.key
cp -f $CERTFILES/$SUBDOMAIN/dhparam.pem /root/bwdata/ssl/$SUBDOMAIN/dhparam.pem

DOESN'T WORK
# Symlink TLS certificates
ln -sfv $CERTFILES/$SUBDOMAIN/fullchain.pem /root/bwdata/ssl/$SUBDOMAIN/certificate.crt
ln -sfv $CERTFILES/$SUBDOMAIN/privkey.pem /root/bwdata/ssl/$SUBDOMAIN/private.key
ln -sfv $CERTFILES/$SUBDOMAIN/dhparam.pem /root/bwdata/ssl/$SUBDOMAIN/dhparam.pem
```

enoch85:
Actually @szaimen Let's close this one and continue to work here instead: https://github.com/morph027/vm/blob/master/apps/talk_signaling.sh

szaimen:
Hm... have you seen the options inside the config.yml file of bitwarden?
Maybe we can make use of it to link to the correct path inside the VM:
```
# Configure Nginx for SSL.
ssl: false
#
# SSL versions used by Nginx (ssl_protocols). Leave empty for recommended default.
# Learn more: https://wiki.mozilla.org/Security/Server_Side_TLS
ssl_versions:
#
# SSL ciphersuites used by Nginx (ssl_ciphers). Leave empty for recommended default.
# Learn more: https://wiki.mozilla.org/Security/Server_Side_TLS
ssl_ciphersuites:
#
# Installation uses a managed Let's Encrypt certificate.
ssl_managed_lets_encrypt: false
#
# The actual certificate. (Required if using SSL without managed Let's Encrypt)
# Note: Path uses the container's ssl directory. The `./ssl` host directory is mapped to
# `/etc/ssl` within the container.
ssl_certificate_path:
#
# The certificate's private key. (Required if using SSL without managed Let's Encrypt)
# Note: Path uses the container's ssl directory. The `./ssl` host directory is mapped to
# `/etc/ssl` within the container.
ssl_key_path:
#
# If the certificate is trusted by a CA, you should provide the CA's certificate.
# Note: Path uses the container's ssl directory. The `./ssl` host directory is mapped to
# `/etc/ssl` within the container.
ssl_ca_path:
#
# Diffie Hellman ephemeral parameters
# Learn more: https://security.stackexchange.com/q/94390/79072
# Note: Path uses the container's ssl directory. The `./ssl` host directory is mapped to
# `/etc/ssl` within the container.
ssl_diffie_hellman_path:
```
But I actually don't know which we need to enter. Do you?

apps/tmbitwarden.sh
```diff
@@ -61,39 +61,154 @@ msg_box "OK, please do the necessary preparations before you run this script and
...
-sed -i "s|https_port.*|https_port: 8443|g" /root/bwdata/config.yml
+sed -i "s|http_port.*|http_port: 5178|g" /root/bwdata/config.yml
+sed -i "s|https_port.*|https_port: 5179|g" /root/bwdata/config.yml
+sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /root/bwdata/config.yml
+sed -i 's|http://|https://|g' /root/bwdata/env/global.override.env
+
+# add certificate files to config.yml
+sed -i "s|^ssl_certificate_path:.*|ssl_certificate_path: $CERTFILES/$SUBDOMAIN/fullchain.pem|g" /root/bwdata/config.yml
+sed -i "s|^ssl_key_path:.*|ssl_key_path: $CERTFILES/$SUBDOMAIN/privkey.pem|g" /root/bwdata/config.yml
+sed -i "s|^ssl_diffie_hellman_path:.*|ssl_diffie_hellman_path: $CERTFILES/$SUBDOMAIN/dhparam.pem|g" /root/bwdata/config.yml
```
szaimen, line 191: This should theoratically work now, but I think we don't need this at all since we are using the apache2 reverse proxy. 
It probably never gets into action because of our reverse proxy.
So I don't get why you are trying so hard to get this right?

szaimen:
So after thinking about this a bit longer, I think my first attempt was better since you don't need to enter the domain twice, without having to change something in run.sh and bitwarden.sh. which should make everything more stable and easier.
If you aggree, I can revert all not needed changes.

And btw: have you tried to run the initial script?

szaimen:
I've started a new attempt here: #1362
since this PR is already a bit cluttered and not going into the right direction, ihmo.

apps/tmbitwarden.sh
enoch85: It hit me later than it would never work as the redirection wouldn't work anyway.

Let's continue in the other PR as you proposed.

PR #1362, opened 2020-08-11
update bitwarden to allow subdomains

### LINKED ISSUE #1336: Use a Reverse Proxy for Bitwarden and generate certs with our function

szaimen:
second attempt to fix #1336

What works is the proxy: 
- you have access to bitwarden after executing the script on your subdomain with tls

What doesn't work on my local test machine with not enough specs: 
- creating a bitwarden account.

What we should do next, imo:
1. test it on a VPS with enough specs
2. if creating an admin account doesn't work there, neither: try to implement the proxy settings from here: bitwarden/server#314
3. If creating an admin account doesn't work after step 2, try to use localhost as the domain during the bitwarden setup instead of the real subdomain and don't change anything in config.yml and globaloverride to https://.
Additionally, ask for a subdomain first.
4. Now it should definitely work.
5. create some tests to make it more reliable and reset everything to the default, if anything goes wrong

### DIFF
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -94,7 +94,8 @@ check_command ./bitwarden.sh install
 if grep ^url /root/bwdata/config.yml | grep -q https || grep ^url /root/bwdata/config.yml | grep -q localhost
 then
     message "It seems like you have entered some wrong settings. We will remove bitwarden now again so that you can start over again."
-    ./bitwarden stop && docker system prune -af
+    check_command ./bitwarden.sh install
+    docker system prune -af
     rm -rf /root/bwdata
     exit 1
 fi
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -41,8 +41,7 @@ print_text_in_color "$ICyan" "Installing Bitwarden password manager..."
 
 msg_box "Bitwarden is a password manager that is seperate from Nextcloud, though we provide this service because it's self hosted and secure.
 
-If you just want to run Bitwarden locally (not connecting your smartphone) then you can use 'localhost' as domain.
-If you on the other hand want to run this on a domain, then please create a DNS record and point it to this server.
+If you on the other hand want to run this on a subdomain, then please create a DNS record and point it to this server.
 In the process of setting up Bitwarden you will be asked to generate an TLS cert with Let's Enrypt so no need to get your own prior to this setup.
 
 The script is based on this documentation: https://help.bitwarden.com/article/install-on-premise/
@@ -51,8 +50,8 @@ It's a good idea to read that before you start this script.
 Please also report any issues regarding this script setup to $ISSUES"
 
 msg_box "The necessary preparations to run expose Bitwarden to the internet are:
-1. The HTTP proxy and HTTPS ports for Bitwarden are 8080 and 8443, please open those ports before running this script.
-2. Please create a DNS record and point that to this server.
+1. Please open port 443 and 80 and point to this server.
+2. Please create a DNS record for your subdomain and point that to this server.
 3. Raise the amount of RAM to this server to at least 3 GB."
 
 if [[ "no" == $(ask_yes_or_no "Have you made the necessary preparations?") ]]
@@ -61,9 +60,22 @@ msg_box "OK, please do the necessary preparations before you run this script and
 
 To run this script again, execute $SCRIPTS/menu.sh and choose Additional Apps --> Bitwarden"
     exit
-else
-    sleep 0.1
 fi
+
+msg_box "ATTENTION!
+
+You will see in the next step a setup which is provided by bitwarden.
+You have to make your choices based on the following instructions, 
+otherwise things can go wrong since we have no influence on those scripts.
+Best is if you take a picture of those instructions so that you remember them later on.
+And please don't worry: we will create a valid SSL certificate afterwards ourselves.
+
+1. Enter your subdomain like e.g. 'bitwarden.example.com'
+2. You will be asked if you want to use 'lets encrypt'; Please type in no: 'n'
+3. Enter your installation ID, which you must get on https://bitwarden.com/host
+4. Enter your installation key, which you will also get there
+5. You will be asked if you have a SSL certificate which you want to use; Please type in no: 'n'
+6. You will be asked if you want to generate a self-signed SSL certificate; Please type in no: 'n'"
 
 # Install Docker
 install_docker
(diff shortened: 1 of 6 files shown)


szaimen:
@enoch85 I mean we could bypass all of this easily when using bitwarden_rs which would be much simpler to implement incl. some examples for apache2 reverse proxies.

enoch85:
First attempt:

```
Migrating database.
Unhandled exception. System.Data.SqlClient.SqlException (0x80131904): Login failed for user 'sa'.
   at Bit.Setup.Program.MigrateDatabase(Int32 attempt) in /home/runner/work/server/server/util/Setup/Program.cs:line 193
   at Bit.Setup.Program.Update() in /home/runner/work/server/server/util/Setup/Program.cs:line 140
   at Bit.Setup.Program.Main(String[] args) in /home/runner/work/server/server/util/Setup/Program.cs:line 65
ClientConnectionId:82aa6fc8-efa0-46ee-9327-ba1457305bb7
Error Number:18456,State:1,Class:14
./bitwarden.sh updatedb failed
```

enoch85:
Second attempt (no changes made): 
Screenshot from 2020-08-12 09:42:49

So it seems to work, but we have to tidy up the prompts to make them easier to  read, and give better instructions. I can take care of that.

Also, Iit would be nice to implement a whiptail for reinstall and uninstall. Would you mind? :)

enoch85:
> bitwarden_rs

Reading from their docs: 

> warningIMPORTANTwarning: When using this server, please report any bugs or suggestions to us directly (look at the bottom of this page for ways to get in touch), regardless of whatever clients you are using (mobile, desktop, browser...). DO NOT use the official support channels.

Thanks, but no thanks. :)

enoch85:
A little bonus, the Bitwarden Registration function works. :)

szaimen:
> A little bonus, the Bitwarden Registration function works. :)

Okay great, so it seems that 1.5GB for the NcVM are really too less when trying to install Bitwarden 😅

szaimen:
> So it seems to work, but we have to tidy up the prompts to make them easier to read, and give better instructions. I can take care of that.

Cool!

> Also, Iit would be nice to implement a whiptail for reinstall and uninstall. Would you mind? :)

Hm, theoratically yes but we still cannot install it when the files already exist since the user would loose all files.
So I guess this is out of scope unless we can automate 
https://community.bitwarden.com/t/restoring-bitwarden-installation-to-another-server/1195 which we cannot, I fear.

apps/tmbitwarden.sh
```diff
@@ -78,22 +92,126 @@ cd /root
...
 check_command ./bitwarden.sh rebuild
 check_command ./bitwarden.sh start
-if check_command ./bitwarden.sh updatedb
+check_command ./bitwarden.sh updatedb
+
+# Produce reverse-proxy config and get lets-encrypt certificate
+msg_box "We will not setup the Apache Proxy that will act as TLS front for your Bitwarden installation."
+
+# Curl the lib another time to get the correct HTTPS_CONF
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
```
szaimen, line 110: I remember that you've said, that there is an alternative to this notation. What was it? I currently don't remember.
Travis doesn't seem to like this notation...

szaimen:
> > Also, Iit would be nice to implement a whiptail for reinstall and uninstall. Would you mind? :)
> 
> Hm, theoratically yes but we still cannot install it when the files already exist since the user would loose all files.
> So I guess this is out of scope unless we can automate
> https://community.bitwarden.com/t/restoring-bitwarden-installation-to-another-server/1195 which we cannot, I fear.

We could only do this if we would allow the user to delete his old data, which shouldn't be the case?

szaimen:
I've just added some checks/cleanup which should make the whole script more stable.

apps/tmbitwarden.sh
```diff
@@ -78,22 +92,145 @@ cd /root
...
+sed -i "s|https_port.*|https_port: 5179|g" /root/bwdata/config.yml
+# Get Subdomain from config.yml and change it to https
+SUBDOMAIN=$(grep ^url /root/bwdata/config.yml)
+SUBDOMAIN=${SUBDOMAIN##*url: http://}
+sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /root/bwdata/config.yml
+sed -i 's|http://|https://|g' /root/bwdata/env/global.override.env
 check_command ./bitwarden.sh rebuild
 check_command ./bitwarden.sh start
-if check_command ./bitwarden.sh updatedb
+check_command ./bitwarden.sh updatedb
```
szaimen, line 115: BTW: if you really don't like that we already start bitwarden here, although the ssl certificate is not existing, we could also do that later. 

But maybe it isn't that important since the proxy to the correct port happens later when proxying it 😅

szaimen: What I mean is thtat we could move these three lines to the complete end of the script if we wanted ;)

enoch85:
> Hm strange, I got the same error when I was trying to install it (because of that I removed the check_command locally). Could you track it down why it has worked the second time and not the first time? Did you e.g. allocate more ressources?

I think it's due to that it was installed before. I run 3 GB on my VPS, so there should be enough RAM. Bitwarden themselves says that 1 GB for standalone is sufficient - which I can confirm.

> We could only do this if we would allow the user to delete his old data, which shouldn't be the case?

Now when thinking about it, no - maybe it's not a good idea. I was thinking like: Uninstall --> Are you sure --> Are you really sure? --> All your data will be gone, are you sure? ---> Fine, let's delete it. :laughing:

apps/tmbitwarden.sh
enoch85: Idk, but Bitwarden doesn't have to start until the end anyway, but it might be a good idea to start if before we begin with the proxy so that we don't clutter the system when not needed.

Also, now when thinking about it, if the Bitwarden installation fails, we actually should remove `/bwdata` so that the script can be run again, since no passwords are lost anyway. Right?

apps/tmbitwarden.sh
```diff
@@ -78,22 +92,149 @@ cd /root
...
 check_command ./bitwarden.sh install
-sed -i "s|http_port.*|http_port: 8080|g" /root/bwdata/config.yml
-sed -i "s|https_port.*|https_port: 8443|g" /root/bwdata/config.yml
+
+# Check if all ssl settings were entered correctly
+if grep ^url /root/bwdata/config.yml | grep -q https || grep ^url /root/bwdata/config.yml | grep -q localhost
+then
+    message "It seems like you have entered some wrong settings. We will remove bitwarden now again so that you can start over again."
+    check_command docker rm bitwarden-nginx bitwarden-admin bitwarden-events bitwarden-attachments \
+    bitwarden-identity bitwarden-api bitwarden-web bitwarden-icons bitwarden-notifications bitwarden-mssql
```
enoch85, line 101: It would be better to run `./bitwarden stop && docker system prune -af`

apps/tmbitwarden.sh
enoch85: Aah, just saw that you did that now. Great!

apps/tmbitwarden.sh
```diff
@@ -78,22 +92,149 @@ cd /root
...
+    rm -rf /root/bwdata
+    exit 1
+fi
+
+# Continue with the installation
+sed -i "s|http_port.*|http_port: 5178|g" /root/bwdata/config.yml
+sed -i "s|https_port.*|https_port: 5179|g" /root/bwdata/config.yml
```
enoch85, line 108: Why these ports? Are they Bitwarden default?

szaimen: no, we only cannot use prots 80 and 443 and also not the ports that onlyoffice, collabroa & co use (have to be special)
Those are the ports that the guide that I followed (dennisnotes...) has used...

apps/tmbitwarden.sh
szaimen: okay

apps/tmbitwarden.sh
enoch85: OK :+1:

enoch85: Please add that to the NONO_PORTS in the TURN_INSTALL in lib.sh.

apps/tmbitwarden.sh
szaimen: fixed

apps/tmbitwarden.sh
szaimen: Hm can you guide me where I find those NONO_PORTS? a search didn't find them and it is also not in TURN_INSTALL if I looked correctly...

apps/tmbitwarden.sh
szaimen: yup

szaimen:
> > Hm strange, I got the same error when I was trying to install it (because of that I removed the check_command locally). Could you track it down why it has worked the second time and not the first time? Did you e.g. allocate more ressources?
> 
> I think it's due to that it was installed before. I run 3 GB on my VPS, so there should be enough RAM. Bitwarden themselves says that 1 GB for standalone is sufficient - which I can confirm.

Hm... how can we work around this? just rebuild -> start -> updatedb a second time or add a sleep command in between those commands?

szaimen:
> > We could only do this if we would allow the user to delete his old data, which shouldn't be the case?
> 
> Now when thinking about it, no - maybe it's not a good idea. I was thinking like: Uninstall --> Are you sure --> Are you really sure? --> All your data will be gone, are you sure? ---> Fine, let's delete it. 😆

Yah, this truly doesn't sound like a good idea 😄

apps/tmbitwarden.sh
szaimen: was it `source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)` ?

apps/tmbitwarden.sh
szaimen: ah I found them: are in talk.sh

szaimen: fixed

apps/tmbitwarden.sh
enoch85: Maybe, but let's leave it as is. :)

enoch85:
I've tested this so many times now that I have to change subomdain due to LE restrictions. :rofl:

szaimen:
> I've tested this so many times now that I have to change subomdain due to LE restrictions. 🤣

That happened to me yesterday, too 😅

enoch85:
Just tested on a new subdomain and everyting works as expected.

Stuff left:

1. A better more clear IMGUR screenshot with removed host keys
2. Try the suggestion to add a seperate user for Bitwarden since their support  thought that was the best soulution. I suggest `ncbitwarden`.

```
You will want to configure the Bitwarden Server to use a `bitwarden` service account. $USER=bitwarden You will want to have your installation owned by the bitwarden service account, and you should be logged in as bitwarden. After those are verified, you will want to make sure the UID and GID in the /bwdata/env/uid.env file are set to your bitwarden service account id numbers in Linux. When using the bitwarden service account you will also need to follow these steps:

sudo groupadd docker

sudo usermod -aG docker $USER

sudo vi bitwarden.service

[Unit]
Description=Bitwarden
Requires=docker.service
After=docker.service

[Service]
Type=oneshot
User=bitwarden
Group=bitwarden
RemainAfterExit=true

[Install]
WantedBy=multi-user.target

systemctl daemon-reload

```

I have to go now, maybe you could have a look at that? Thanks!

enoch85:
But, let's do that in another PR btw, I'll merge this now if you are fine with it?

szaimen:
> But, let's do that in another PR btw, I'll merge this now if you are fine with it?

Yes I am fine with merging. 👍

Doing the rest in a new PR sounds like a good idea!

apps/tmbitwarden.sh
```diff
@@ -61,39 +60,165 @@ msg_box "OK, please do the necessary preparations before you run this script and
...
+# Get Subdomain from config.yml and change it to https
+SUBDOMAIN=$(grep ^url /root/bwdata/config.yml)
+SUBDOMAIN=${SUBDOMAIN##*url: http://}
+sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /root/bwdata/config.yml
+sed -i 's|http://|https://|g' /root/bwdata/env/global.override.env
 check_command ./bitwarden.sh rebuild
 check_command ./bitwarden.sh start
-if check_command ./bitwarden.sh updatedb
+check_command ./bitwarden.sh updatedb
```
szaimen, line 113: @enoch85 so this seems to make still problems. As I read in the bitwarden issue...

BTW: Why is it needed at all?
I couldn't find the reason why this is needed in the bitwarden install manual...

szaimen: Hm, Maybe this could help:
https://github.com/bitwarden/server/issues/155#issuecomment-381177257

enoch85: Read this: investigate and make a PR. Sorry, I'm busy now.

Thanks!

PR #1364, opened 2020-08-12
run Bitwarden as a non-privileged user

### LINKED ISSUE #1363: Improve Bitwarden even more
@szaimen As discussed.

Stuff left:

1. A better more clear IMGUR screenshot with removed host keys
2. Try the suggestion to add a seperate user for Bitwarden since their support  thought that was the best soulution. I suggest `ncbitwarden` but maybe it has to be `bitwarden`?

`You will want to configure the Bitwarden Server to use a **bitwarden** service account. $USER=bitwarden You will want to have your installation owned by the bitwarden service account, and you should be logged in as bitwarden. After those are verified, you will want to make sure the UID and GID in the /bwdata/env/uid.env file are set to your bitwarden service account id numbers in Linux. `
```
When using the bitwarden service account you will also need to follow these steps:

sudo groupadd docker

sudo usermod -aG docker $USER

sudo vi bitwarden.service

[Unit]
Description=Bitwarden
Requires=docker.service
After=docker.service

[Service]
Type=oneshot
User=ncbitwarden
Group=ncbitwarden
RemainAfterExit=true

[Install]
WantedBy=multi-user.target

systemctl daemon-reload

```

I have to go now, maybe you could have a look at that? Thanks!

szaimen:
Things left to do: 
- [x] bitwarden-registration has to also get changed to feature the new path.
- [x] nextcloud_update also have to get updated.

fix #1363

### DIFF
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -107,6 +107,11 @@ then
 fi
 
 # Create the service
+msg_box "We will now create a systemd service for Bitwarden.
+
+To start Bitwarden, simply execute: systemctl start bitwarden
+To stop Bitwarden, simply execute: systemctl stop bitwarden"
+
 cat << BITWARDEN_SERVICE > /etc/systemd/system/bitwarden.service
 [Unit]
 Description=Bitwarden
@@ -165,6 +170,7 @@ SUBDOMAIN=${SUBDOMAIN##*url: http://}
 sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /home/"$BITWARDEN_USER"/bwdata/config.yml
 sed -i 's|http://|https://|g' /home/"$BITWARDEN_USER"/bwdata/env/global.override.env
 check_command sudo -u "$BITWARDEN_USER" ./bitwarden.sh rebuild
+print_text_in_color "$ICyan" "Starting Bitwarden for the first time, please be patient..."
 check_command systemctl start bitwarden
 check_command sudo -u "$BITWARDEN_USER" ./bitwarden.sh updatedb
 
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -92,13 +92,16 @@ https://i.imgur.com/YPynDAf.png"
 if ! id "$BITWARDEN_USER" >/dev/null 2>&1
 then
     useradd -r "$BITWARDEN_USER"
+else
+    userdel "$BITWARDEN_USER"
+    rm -rf /home/"$BITWARDEN_USER"
 fi
 groupadd docker >/dev/null 2>&1
 sudo usermod -aG docker "$BITWARDEN_USER"
 if [ ! -d /home/"$BITWARDEN_USER" ]
 then
     mkdir -p /home/"$BITWARDEN_USER"
-    chmod -R "$BITWARDEN_USER"n:docker /home/"$BITWARDEN_USER"
+    chown -R "$BITWARDEN_USER"n:docker /home/"$BITWARDEN_USER"
 fi
 
 # Create the service
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -101,7 +101,8 @@ then
     chmod -R "$BITWARDEN_USER"n:docker /home/"$BITWARDEN_USER"
 fi
 
-{
+# Create the service
+cat << BITWARDEN_SERVICE > /etc/systemd/system/bitwarden.service
 [Unit]
 Description=Bitwarden
 Requires=docker.service
@@ -113,11 +114,11 @@ User="$BITWARDEN_USER"
 Group="$BITWARDEN_USER"
 ExecStart=/home/"$BITWARDEN_USER"/bitwarden.sh start
 RemainAfterExit=true
-ExecStop=/home/"$BITWARDEN_USER"/bitwarden.sh stop
+ExecStop= /home/"$BITWARDEN_USER"/bitwarden.sh stop
 
 [Install]
 WantedBy=multi-user.target
-} > /etc/systemd/system/bitwarden.service
+BITWARDEN_SERVICE
 
 sudo chmod 644 /etc/systemd/system/bitwarden.service
 systemctl daemon-reload
--- a/apps/tmbitwarden.sh
+++ b/apps/tmbitwarden.sh
@@ -28,7 +28,7 @@ then
     then
         if is_this_installed apache2
         then
-            if [ -d /root/bwdata ]
+            if [ -d /root/bwdata ] || [ -d /home/ncbitwarden/bwdata ]
             then
                 msg_box "It seems like 'Bitwarden' is already installed.\n\nYou cannot run this script twice, because you would loose all your passwords."
                 exit 1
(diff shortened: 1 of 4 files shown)


szaimen:
BTW: I wonder if installing has to get executed as user ncbitwarden since it will always create the files as the executed user?

apps/tmbitwarden.sh
```diff
@@ -79,38 +82,80 @@ Basically:
...
+RemainAfterExit=true
+ExecStop=/home/"$BITWARDEN_USER"/bitwarden.sh stop
+
+[Install]
+WantedBy=multi-user.target
+BITWARDEN_SERVICE
+
+sudo chmod 644 /etc/systemd/system/bitwarden.service
+systemctl daemon-reload
+sudo systemctl enable bitwarden.service
```
szaimen, line 125: btw: this cannot work here since we have not installed bitwarden here, yet

enoch85: Please move it. Thanks!

apps/tmbitwarden.sh
```diff
@@ -79,38 +82,83 @@ Basically:
...
-SUBDOMAIN=$(grep ^url /root/bwdata/config.yml)
+SUBDOMAIN=$(grep ^url /home/"$BITWARDEN_USER"/bwdata/config.yml)
 SUBDOMAIN=${SUBDOMAIN##*url: http://}
-sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /root/bwdata/config.yml
-sed -i 's|http://|https://|g' /root/bwdata/env/global.override.env
-check_command ./bitwarden.sh rebuild
-check_command ./bitwarden.sh start
-check_command ./bitwarden.sh updatedb
+sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" /home/"$BITWARDEN_USER"/bwdata/config.yml
+sed -i 's|http://|https://|g' /home/"$BITWARDEN_USER"/bwdata/env/global.override.env
```
szaimen, line 158: This is missing here:

sed -i "s|LOCAL_UID=.*|LOCAL_UID=$USERID|g" /home/"$BITWARDEN_USER"/bwdata/env/uid.env
sed -i "s|LOCAL_GID=.*|LOCAL_GID=$USERGROUPID|g" /home/"$BITWARDEN_USER"/bwdata/env/uid.env

enoch85: Aah, thanks. Just adding stuff as fast as possible so I can test now :D

enoch85: Please add btw :)

szaimen: done

apps/tmbitwarden.sh
szaimen: done

szaimen:
> BTW: I wonder if installing has to get executed as user ncbitwarden since it will always create the files as the executed user?

Do you have already an answer to this?

enoch85:
I was thinking we could use `sudo -u $BITWARDEN_USER bitwarden.sh install` have to try that though.

enoch85:
It seems to work. A few glitches that needs fixing though...

szaimen:
I have to do now. Maybe I'll join again later.
Please don't forget the checklist on top of the PR

enoch85:
OK, so the setup works until the service. I tired to run the commands manually afterwards as the `ncbitwarden` user and it worked.

So, If we can fix the service we're good to go I think.

enoch85:
The script itself works, now let's concentrate on the rest. :rocket:

enoch85:
@szaimen Please test this before merge, including the registration and nextcloud_update script.

Thanks!

enoch85:
OK, so I've tested this now and everything seems to work as expected.

I'm too tired to merge though, maybe I missed something. @szaimen Please test this, thanks!

apps/tmbitwarden.sh
```diff
@@ -79,38 +79,92 @@ Basically:
...
+    mkdir -p "$BITWARDEN_HOME"
+    chown -R "$BITWARDEN_USER":"$BITWARDEN_USER" "$BITWARDEN_HOME"
+fi
+
+# Create the service
+msg_box "We will now create a systemd service for Bitwarden.
+
+START Bitwarden, simply execute: systemctl start bitwarden
+STOP Bitwarden, simply execute: systemctl stop bitwarden
+RESTART Bitwarden, simply execute: systemctl restart bitwarden"
```
szaimen, line 105: Hm, I think I would put this to the end of the script...
There are already too many msg_boxes at the beginning of the script

nextcloud_update.sh
```diff
@@ -544,9 +544,20 @@ then
...
                     sleep 2
-                    bash /root/bitwarden.sh updateself
-                    bash /root/bitwarden.sh update
+                    yes no | bash /root/bitwarden.sh updateself
+                    yes no | bash /root/bitwarden.sh update
                 fi
+            elif [ -d "$BITWARDEN_HOME"/bwdata ]
+                then
+                    curl_to_dir "https://raw.githubusercontent.com/bitwarden/server/master/scripts" "bitwarden.sh" "$BITWARDEN_HOME"
+                    chown "$BITWARDEN_USER":"$BITWARDEN_USER" "$BITWARDEN_HOME"
```
szaimen, line 553: I would only chown the downloaded script so:
`chown "$BITWARDEN_USER":"$BITWARDEN_USER" "$BITWARDEN_HOME"/bitwarden.sh`
or we could also chown the whole folder like this:
`chown -R "$BITWARDEN_USER":"$BITWARDEN_USER" "$BITWARDEN_HOME"`

szaimen: Additionally, I would make it exectuable with 
`chmod +x "$BITWARDEN_HOME"/bitwarden.sh`

apps/tmbitwarden.sh
```diff
@@ -218,7 +268,12 @@ add_dockerprune
...
 
-After the account it setup, please disable user registration by running sudo bash $SCRIPTS/menu.sh and choose:
-Additional Apps --> Bitwarden Registration"
+After the account is registered, please disable user registration by running sudo bash $SCRIPTS/menu.sh and choose:
+Additional Apps --> Bitwarden Registration
+
+Some notes to the bitwarden service:
+to START Bitwarden, simply execute: 'systemctl start bitwarden'
+to STOP Bitwarden, simply execute: 'systemctl stop bitwarden'
+to RESTART Bitwarden, simply execute: 'systemctl restart bitwarden'"
```
szaimen, line 277: BTW: are you really sure, that restart works? since we haven't configured this in the service file...

apps/tmbitwarden.sh
szaimen: I've moved it to the buttom.

nextcloud_update.sh
szaimen: this is fixed

szaimen:
BTW: Maybe some words concerning complete testing of big PRs from myself:

The thing with extensive testing for me is, that I first have to create a VPS, make DNS-entries, setup a NcVM from scratch, etc, which is relatively timeexpensive, imho.

I could also test this locally on a test-VM but my laptop only has 4GB of RAM which is not enough for using more than 1.5GB for a VM which is really not enough for this, also my productive machine runs on the same network (with the same public ipaddress) and I would have to shut it down for testing and change the port forwarding, etc.

Also I don't want to have the responsibility if something doesn't work. So I believe you if you say that everything works :)

Maybe that explains a bit better my problem with extensive testing of big PRs...

So if it is a smaller thing: sure, testing is no problem, because I can test it locally on a small VM but especially if TLS has to get used or if I need more than 1.5GB of RAM for the VM, this won't work.

nextcloud_update.sh
```diff
@@ -540,13 +540,26 @@ then
...
+                    chown "$BITWARDEN_USER":"$BITWARDEN_USER" "$BITWARDEN_HOME"/bitwarden.sh
+                    chmod +x "$BITWARDEN_HOME"/bitwarden.sh
+                    if [ -f "$BITWARDEN_HOME"/bitwarden.sh ]
+                    then
+                        print_text_in_color "$IGreen" "Upgrading Bitwarden..."
+                        sleep 2
+                        yes no | sudo -u "$BITWARDEN_USER" bash "$BITWARDEN_HOME"/bitwarden.sh updateself
+                        yes no | sudo -u "$BITWARDEN_USER" bash "$BITWARDEN_HOME"/bitwarden.sh update
+                    fi
             fi
```
szaimen, line 563: hm... isn't here a `fi` lacking? It looks to me like this...

szaimen: no, was only the indentation. is fixed.

enoch85:
Let's do this!

Tested restart and it works.

PR #1365, opened 2020-08-12
use logrotate for nextcloud logs

### LINKED ISSUE #1280: Use logrotate to store logs
## Steps to reproduce
1. Use Nextcloud
2. Check logfiles at `/var/log/nextcloud`
3. Two files are kept: `nextcloud.log`and `nextcloud.log.1`

## Expected behaviour
More than one log file should be saved for archive

## Actual behaviour
One file is kept.

## Solution
Added nextcloud.log to logrotate using webmin and save e.g. 30 log files.

## Server configuration

**Nextcloud server version:** 19.0.0

**How did you install the VM?** Bought with device

#### Screenshoots
Bildschirmfoto 2020-06-07 um 14 01 30

Bildschirmfoto 2020-06-07 um 14 01 51

szaimen:
first attempt to fix #1280

for testing probably use `logrotate --force  /etc/logrotate.d/nextcloud.log.conf`
(https://stackoverflow.com/questions/2117771/is-it-possible-to-run-one-logrotate-check-manually)

How logrotate works
The system runs logrotate on a schedule, usually daily. On most distributions, the script that runs logrotate daily is located at /etc/cron.daily/logrotate.

Some distributions use a variation. For example, on Gentoo the logrotate script is located at /etc/cron.daily/logrotate.cron.

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -427,8 +427,15 @@ occ_command config:system:set mail_smtpmode --value="smtp"
 # Forget login/session after 30 minutes
 occ_command config:system:set remember_login_cookie_lifetime --value="1800"
 
-# Set logrotate (max 10 MB)
-occ_command config:system:set log_rotate_size --value="10485760"
+# Set logrotate (without size restriction)
+occ_command config:system:set log_rotate_size --value=0
+
+# Configure logrotate to rotate logs for us (max 10, every day a new one)
+touch /etc/logrotate.d/nextcloud.log.conf
+"/var/log/nextcloud/nextcloud.log {
+daily
+rotate 10
+}" > /etc/logrotate.d/nextcloud.log.conf
 
 # Set trashbin retention obligation (save it in trahbin for 6 months or delete when space is needed)
 occ_command config:system:set trashbin_retention_obligation --value="auto, 180"
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -432,7 +432,7 @@ occ_command config:system:set log_rotate_size --value=0
 
 # Configure logrotate to rotate logs for us (max 10, every day a new one)
 touch /etc/logrotate.d/nextcloud.log.conf
-"/var/log/nextcloud/nextcloud.log {
+"$VMLOGS/nextcloud.log {
 daily
 rotate 10
 }" > /etc/logrotate.d/nextcloud.log.conf
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -427,16 +427,8 @@ occ_command config:system:set mail_smtpmode --value="smtp"
 # Forget login/session after 30 minutes
 occ_command config:system:set remember_login_cookie_lifetime --value="1800"
 
-# Set logrotate (without size restriction)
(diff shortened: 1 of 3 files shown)


nextcloud_install_production.sh
```diff
@@ -427,8 +427,15 @@ occ_command config:system:set mail_smtpmode --value="smtp"
...
 occ_command config:system:set remember_login_cookie_lifetime --value="1800"
 
-# Set logrotate (max 10 MB)
-occ_command config:system:set log_rotate_size --value="10485760"
+# Set logrotate (without size restriction)
+occ_command config:system:set log_rotate_size --value=0
+
+# Configure logrotate to rotate logs for us (max 10, every day a new one)
+touch /etc/logrotate.d/nextcloud.log.conf
+"/var/log/nextcloud/nextcloud.log {
```
enoch85, line 435: Please use `$VMLOGS/nextcloud.log` here

enoch85:
@rakekniven Can you please test this?

nextcloud_install_production.sh
szaimen: done

rakekniven:
> @rakekniven Can you please test this?

Hello, I have no test environment here and for me it is not clear how to up a testinstance including this change.

szaimen:
With my changes everything seems to work, although I get this error at the beginning of the new log each time logrotate runs:
image
But I don't think that it is a problem.

I tested forcing 10 log files to create (it really doesn't exceed 10 files) and it always moves the old log to a new file until 10 are reached. Then it starts to delete the oldest log.
Also the creation of the nextcloud.log.conf, etc (this PR) works.
To make it work, I had to change the settings in my installation for the /var/log/nextcloud folder (already included in this PR).

What I couldn't test is if it really works to create a new daily conf file since I don't know if logrotate runs always daily or if we have to configure it somehow...
@enoch85 @rakekniven any idea?

szaimen:
What about this: 
I am already sure enough that it doesn't destroy anything:
@rakekniven You could execute the following script on your NcVM or a test-NCVM and report if it works as expected? (it should create every day a new log...)
[logrotate.zip](https://github.com/nextcloud/vm/files/5068747/logrotate.zip)

enoch85:
Please remember that the OS disk is only 40 GB in the VM.

szaimen:
> Please remember that the OS disk is only 40 GB in the VM.

Yes, I do. 
Do you think that it is possible that logs are bigger than e.g. 100MB on one day? Because if "only" 100MB a day, it would together fill max 1GB of space which is okay, imo.
Realistically I would say it will be less than 10MB per day which will result in max 100MB for all logs together

szaimen:
We could of course alternatively set the log_rotate size of nextcloud to something like 100MB which should never occur but limit the max size ...

enoch85:
> fill max 1GB of space which is okay, imo.

I think that sounds good, but since @rakekniven wanted the function in the first place, what does he say?

Also, I think this should be an option and nothing default. Put it in Nextcloud configuration.

Thanks!

rakekniven:
Works for me :-)

szaimen:
I've already tested the latest changes. should be ready to go now.

enoch85:
> I've already tested

OK, let's do it!

PR #1366, opened 2020-08-13
attempt to fix bitwarden

szaimen:


szaimen:
If that dowsn't work at the first time, we could also raise the ram size to 4GB or enable docker container for mssql in config.yml

enoch85:
Not tested but at least the functions are correct now. 

I don't think that BW needs to be started at all for it to rebuild though, to me that order seems strange

szaimen:
Actually I am pretty unsure about this, too and don't think that this will fix the issue, hence I'll close this PR. 
This bug is so strange that it can really be anything. Maybe too less RAM, maybe too less CPU power, maybe a wrong storage maybe anything...

enoch85:
Have you tested these changes?

Anyway, it's really just if you re-install that the issue appears. 

Maybe we should put the msql in a docker as well as that seemed to work?

PR #1370, opened 2020-08-15
Bitwarden mailconfig

szaimen:


szaimen:
Some comments:
- What does `adminSettings__admins=` do?
- is it really enough to just restart bitwarden after writing those information to the file?
- Those need to get configured:
```
globalSettings__mail__smtp__host=REPLACE
globalSettings__mail__smtp__port=587
globalSettings__mail__smtp__ssl=false
globalSettings__mail__smtp__username=REPLACE
globalSettings__mail__smtp__password=REPLACE
```

szaimen:
I haven't tested this but it should already work.

apps/tmbitwarden.sh
```diff
@@ -165,6 +165,11 @@ SUBDOMAIN=$(grep ^url "$BITWARDEN_HOME"/bwdata/config.yml)
 SUBDOMAIN=${SUBDOMAIN##*url: http://}
 sed -i "s|^url: .*|url: https://$SUBDOMAIN|g" "$BITWARDEN_HOME"/bwdata/config.yml
 sed -i 's|http://|https://|g' "$BITWARDEN_HOME"/bwdata/env/global.override.env
+# Insert globalSettings__mail__smtp__trustServer to global.override
+if ! grep -q "^globalSettings__mail__smtp__trustServer=" "$BITWARDEN_HOME"/bwdata/env/global.override.env
+then
+    echo "globalSettings__mail__smtp__trustServer=" >> "$BITWARDEN_HOME"/bwdata/env/global.override.env
```
enoch85, line 171: IMHO, this should only be set if you run the SMTP script, not otherwise.

menu/additional_apps.sh
```diff
@@ -34,6 +33,11 @@ choice=$(whiptail --title "Which apps do you want to install?" --checklist "Auto
 "SMB-mount" "(Connect to SMB-shares from your local network)" OFF 3>&1 1>&2 2>&3)
 
 case "$choice" in
+    *"Bitwarden "*)
+        clear
+        print_text_in_color "$ICyan" "Downloading Bitwarden_Menu.sh..."
```
enoch85, line 38: Actually, I think we should change the info here. It would be better if it said "Downloading script for Bitwarden..."

enoch85:
Generally looks promising! Thank you!

apps/tmbitwarden.sh
szaimen: Done

menu/additional_apps.sh
szaimen: Done

apps/bitwarden-mailconfig.sh
```diff
@@ -0,0 +1,218 @@
...
+while true
+do
+    USE_SSL=$(whiptail --inputbox "Do you want to use SSL for your Mailserver? If yes: please type in 'yes'.\nPlease note: if your Mailserver only supports Starttls, you cannot use SSL - so please answer in this case here 'no'.\nIf you don't want to change the SSL-setting, that is already configured inside the global.override.env-file, just leave the box empty." "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    if [[ "no" == $(ask_yes_or_no "Are your sure about your answer?") ]]
+    then
+        msg_box "It seems like your weren't satisfied by the answer you entered. Please try again."
+    else
+        if [ "$USE_SSL" = "yes" ]
+        then
+            DEFAULT_PORT=465
```
enoch85, line 73: I think we should add 587 which is TLS and most common in SMTP.

apps/bitwarden-mailconfig.sh
```diff
@@ -0,0 +1,218 @@
...
+        msg_box "It seems like your weren't satisfied by the Mailserver you entered. Please try again."
+    else
+        break
+    fi
+done
+
+# Enter if you want to use ssl
+while true
+do
+    USE_SSL=$(whiptail --inputbox "Do you want to use SSL for your Mailserver? If yes: please type in 'yes'.\nPlease note: if your Mailserver only supports Starttls, you cannot use SSL - so please answer in this case here 'no'.\nIf you don't want to change the SSL-setting, that is already configured inside the global.override.env-file, just leave the box empty." "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
```
enoch85, line 66: SSL here is port 465, that should be clear.

szaimen: Hm, I am not sure about this. Isn't there a possibility where you can enable this on another Port?

apps/bitwarden-mailconfig.sh
szaimen: Hm, but is it also default in most cases together with SSL?

enoch85: SSL is older than TLS, less secure. 

I think we should provide SSL and TLS (587)

apps/bitwarden-mailconfig.sh
enoch85: Isn't 465 hardcoded?

szaimen: No, it isn't. 
I just use the Default_port variable to suggest a Port based on if he wants to use ssl or not...

apps/bitwarden-mailconfig.sh
szaimen: it looks like there is an additional setting for starttls:
globalSettings__mail__smtp__startTls=true
https://github.com/bitwarden/server/issues/451#issuecomment-481676589

szaimen:
Since there is obviously an additional setting for starttls:
globalSettings__mail__smtp__startTls=true
https://github.com/bitwarden/server/issues/451#issuecomment-481676589
I would suggest to add this...

apps/bitwarden-mailconfig.sh
enoch85: Aaah, to tired now. :smile: Well, I think 587 (TLS) should be the default.

apps/bitwarden-mailconfig.sh
enoch85: Yup, that's it!

enoch85:
> I would suggest to add this...

Set it as default, and make it possible for the user to change port if wanted.

apps/bitwarden-mailconfig.sh
szaimen: You can choose the Port already independently from the used encryption protocol.

szaimen:
> > I would suggest to add this...
> 
> Set it as default, and make it possible for the user to change port if wanted.

Already done by the latest commit 👍

apps/bitwarden-mailconfig.sh
```diff
@@ -63,22 +69,26 @@ done
 # Enter if you want to use ssl
 while true
 do
-    USE_SSL=$(whiptail --inputbox "Do you want to use SSL for your Mailserver? If yes: please type in 'yes'.\nPlease note: if your Mailserver only supports Starttls, you cannot use SSL - so please answer in this case here 'no'.\nIf you don't want to change the SSL-setting, that is already configured inside the global.override.env-file, just leave the box empty." "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    PROTOCOL=$(whiptail --inputbox "Please type in the encryption Protocol for your mailserver.\nThe available options are 'SSL', 'STARTTLS' or 'none'.\n\nIf you don't want to change the Protocol-setting, that are already configured inside the global.override.env-file, just leave the box empty." "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
```
enoch85, line 72: Perfect solution!

enoch85:
Are we ready to merge this?

enoch85:
The text needs to be looked at, but I'm too tired now. 

Code wise - looks good
Menu wise - looks good
Grammar - needs to be looked at

Thanks @szaimen! :1st_place_medal:

enoch85:
> Actually, a rebuild is not needed based on @mamama1's experience.

Done.

Haven't tested, but it looks good.

Would you say we can merge? Have you tested?

apps/bitwarden-mailconfig.sh
```diff
@@ -0,0 +1,235 @@
...
+if [ -n "$ADMIN_ACCOUNT" ]
+then
+    check_command sed -i "s|^adminSettings__admins=.*|adminSettings__admins=$ADMIN_ACCOUNT|g" "$BITWARDEN_HOME"/bwdata/env/global.override.env
+fi
+
+# Start Bitwarden
+systemctl start bitwarden
+msg_box "Your Bitwarden mailserver settings should be successfully changed by now.
+
+If you experiance any issue, please report them to $ISSUES"
```
szaimen, line 234: typo?

szaimen:
I've tested it on a test-VM with non-working bitwarden, so cannot say for sure, that everything works. 
But at least the input and write process works. 👍

apps/bitwarden-mailconfig.sh
```diff
@@ -0,0 +1,235 @@
...
+    echo "globalSettings__mail__smtp__startTls=false" >> "$BITWARDEN_HOME"/bwdata/env/global.override.env
+fi
+
+# Enter mailserver
+while true
+do
+    MAIL_SERVER=$(whiptail --inputbox "Please enter the mailserver URL that you want to use.\nE.g. smtp.mail.de\nIf you don't want to change the mailserver, that is already configured inside the global.override.env-file, just leave the box empty." "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    if [[ "no" == $(ask_yes_or_no "Is this correct? $MAIL_SERVER") ]]
+    then
+        print_text_in_color "$ICyan" "OK, please try again."
```
szaimen, line 63: btw: if you don't want a msg_box here, I would suggest to add a `sleep 2` after this, else you won't be able to read it.

szaimen: This of course is meant for all places where you have done that.

szaimen: `sleep 1` is probably enough

PR #1377, opened 2020-08-19
fix problem with static_ip

### LINKED ISSUE #1376: No option to set static IP address when DHCP & Internet Fully operational
## Steps to reproduce
Mount ova in VMWare Player and let it  run through its startup, choose to set static IP address from the menu

## Expected behaviour
per the instructions:
"4.1 TLDR;
  
The startup-script needs a working internet connection to run. It will check in the following order:

DHCP OK? --> Yes --> Can we reach internet? ---> Yes --> Continue to run the script --> Ask to set static IP later as an option.
DHCP NOT OK? --> Ask to set static IP --> Can we reach internet? ---> Yes --> Continue to run the script.
DHCP NOT ACTIVATED? --> Ask to set a static address right from the get go.
SET STATIC IP if all of the above fails.
So all and all, it will only ask to set static IP if if can't reach internet with DHCP. You will get the option to set static IP later though."

## Actual behaviour
No option was offered to set a static IP address

## Server configuration

Nextcloud 19.0.1

Open ova file with VMWare Player

## Network
We do use DHCP, but wanted to use a static IP address for this server.

Yes

```
log file is empty

```

```
Insert the CLI output here
```

szaimen:
attempt to fix #1376

### DIFF
--- a/menu/server_configuration.sh
+++ b/menu/server_configuration.sh
@@ -2,20 +2,20 @@
 
 # T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
 
-# Use local lib file in case there is no internet connection
-if [ -f /var/scripts/lib.sh ]
-then
-# shellcheck disable=2034,2059
-true
-# shellcheck source=lib.sh
-source /var/scripts/lib.sh
- # If we have internet, then use the latest variables from the lib remote file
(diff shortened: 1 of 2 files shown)


szaimen:
Was already successfully tested

enoch85:
OK, this is working. Just tested on an old VM.

PR #1380, opened 2020-08-20
Fixed reference to script name change

ZifanPronouncedTony:
The name of the script change-to-zfs-mount-generator was changed to zfs-prune-snapshots.sh in commit f9c9e0d in /vm/disk. This breaks the install process as it fails to download. This PR changes the name to the correct name. Honestly this is my first time on github and I'm not too sure if this is the right way to do this but I couldn't find any other references to that script so hopefully it's correct. Apologies for any inconvenience!

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -184,13 +184,13 @@ case "$choice" in
     "2 Disks Auto")
         run_script DISK format-sdb
         # Change to zfs-mount-generator
-        run_script DISK zfs-prune-snapshots.sh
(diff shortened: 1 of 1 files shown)


ZifanPronouncedTony:
forgot to remove the .sh

enoch85:
Sorry, this is not a valid fix. We made a lot of changes to scripts the last days, and this is a bug we missed.

Thanks for your efforts though!

I will fix it right now.

enoch85:
It's now fixed correctly: https://github.com/nextcloud/vm/pull/1381

PR #1383, opened 2020-08-22
Create bitwarden_rs.sh

szaimen:


apps/talk_signaling.sh
```diff
@@ -104,7 +104,7 @@ else
 fi
 
 # Let the user choose port. TURN_PORT in msg_box is taken from lib.sh and later changed if user decides to.
-NONO_PORTS=(22 25 53 80 443 3306 5178 5179 5432 7983 8983 10000 8081)
+NONO_PORTS=(22 25 53 80 443 1024 3012 3306 5178 5179 5432 7983 8983 10000 8081)
```
enoch85, line 107: Great catch! Please add that to the original script as well.

@morph027 Can you please list all the ports in use by the Signaling server?

szaimen: they were already added to the original talk script 👍

szaimen:
@enoch85 I've tested this sucessfully. Is anthing else needed to get this in?

szaimen:
The only question that remains from my side is, if we should provide an additional script for bitwarden_rs to disable (or enable) the admin panel completely. But I would to this in a followup-PR, if needed.

enoch85:
This is your script, I won't maintain it ;)

If you think it's good, then I'll make a few adjustments when I have time and then merge. :+1:

szaimen:
> If you think it's good, then I'll make a few adjustments when I have time and then merge. 👍

Yes, I think it is good 👍

PR #1387, opened 2020-08-24
further improvements to Talk Signaling

enoch85:
@morph027 @szaimen 

Still internal server error, but Nextcloud reports OK, and curl command works.

**Apache2 log**
```
[Mon Aug 24 22:57:51.555540 2020] [proxy:warn] [pid 64118:tid 140500374681344] [client 5.243.70.111:60706] AH01144: No protocol handler was valid for the URL /favicon.ico (scheme 'ws'). If you are using a DSO version of mod_proxy, make sure the proxy submodules are included in the configuration using LoadModule.
```

Apache also needs access to the logs dir.

### DIFF
--- a/apps/talk_signaling.sh
+++ b/apps/talk_signaling.sh
@@ -369,6 +369,7 @@ then
     ProxyPass /error/ !
     ErrorDocument 404 /error/404_proxy.html
     # Enable proxying Websocket requests to the standalone signaling server.
+    # https://httpd.apache.org/docs/2.4/mod/mod_proxy_wstunnel.html
     ProxyPass / "http://127.0.0.1:8081/"
     RewriteEngine on
     RewriteCond %{HTTP:Upgrade} websocket [NC]
--- a/apps/talk_signaling.sh
+++ b/apps/talk_signaling.sh
@@ -226,6 +226,12 @@ then
     exit 1
 fi
 
+SUBDOMAIN=$(whiptail --title "T&M Hansson IT - Talk Signaling Server" --inputbox "Talk Signaling Server subdomain eg: talk.yourdomain.com\n\nNOTE: This domain must be different than your Nextcloud domain. They can however be hosted on the same server, but would require seperate DNS entries." "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+
+# curl the lib another time to get the correct https_conf
+# shellcheck source=lib.sh
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
 # Check if $SUBDOMAIN exists and is reachable
 print_text_in_color "$ICyan" "Checking if $SUBDOMAIN exists and is reachable..."
 domain_check_200 "$SUBDOMAIN"
--- a/apps/talk_signaling.sh
+++ b/apps/talk_signaling.sh
@@ -259,7 +259,7 @@ sed -i "s|#turn_rest_api_key\s*=.*|$JANUS_API_KEY|" /etc/janus/janus.jcfg
 sed -i "s|#full_trickle|full_trickle|g" /etc/janus/janus.jcfg
 sed -i 's|#interface.*|interface = "lo"|g' /etc/janus/janus.transport.websockets.jcfg
 sed -i 's|#ws_interface.*|ws_interface = "lo"|g' /etc/janus/janus.transport.websockets.jcfg
-check_command systemctl restart janus
+start_if_stopped janus
 check_command systemctl enable janus
 
 # HPB
@@ -295,10 +295,10 @@ url = ws://127.0.0.1:8188
 [turn]
 apikey = ${JANUS_API_KEY}
 secret = ${TURN_SECRET}
-servers = turn:$TURN_DOMAIN:$TURN_PORT?transport=tcp
+servers = turn:$TURN_DOMAIN:$TURN_PORT?transport=tcp,turn:$TURN_DOMAIN:$TURN_PORT?transport=udp
 SIGNALING_CONF_CREATE
 fi
-check_command systemctl restart signaling
+start_if_stopped signaling
 check_command systemctl enable signaling
 
 ### PROXY ###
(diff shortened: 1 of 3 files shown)


enoch85:
Funny, adding *whatever* to the `ProxyPass` removes the Internal Server Error.

```
    ProxyPass "/whatever"  "ws://127.0.0.1:8081/"
    ProxyPassReverse "/whatever"  "ws://127.0.0.1:8081/"
```

@morph027 Why are we not using "signaling-server" as in the Janus docs?

morph027:
The initial idea was to use a separate subdomain so no path was necessary. I'm using the sub-path as well so let's change it to this.

enoch85:
> separate subdomain

I don't understand? We are using a separate subdomain in this setup already.

morph027:
Yes, but as we agreed on a subdomain (apache setup is much easier), we also agreed to not use a subpath. However, doesn't matter, let's use the subdomain +  path,  like signaling.example.com/signaling-server/

I will fix the script and let you know for further testing.

szaimen:
> subdomain + path, like signaling.example.com/signaling-server/

I don't get why we need that subpath? I mean it works already by using the main domain...
What I mean is, that https://"$SUBDOMAIN"/api/v1/welcome returns the correct code...

morph027:
> > subdomain + path, like signaling.example.com/signaling-server/
> 
> I don't get why we need that subpath?
> I mean it works already by using the main domain...
> What I mean is, that https://"$SUBDOMAIN"/api/v1/welcome returns the correct code...

You can't route `/` of the main domain to the signaling server websocket as it requires php for NC.
Also, without a subdomain, we need to add the signaling proxy into the main apache config, which gets written in another script.

Obviously, a plain `/`` for the subdomain does not work either due to bogus apache restrictions (leading to the dso error).

szaimen:
> You can't route `/` of the main domain to the signaling server websocket as it requires php for NC.
> Also, without a subdomain, we need to add the signaling proxy into the main apache config, which gets written in another script.
> 
> Obviously, a plain `/`` for the subdomain does not work either due to bogus apache restrictions (leading to the dso error).

Yes, I get this. What I don't get is why we are now changing this to use the nextcloud-domain although the nextcloud-signaling-server already works with the subdomain...

What I am talking about is: although apache reports internal server error, the nextcloud-signalilng-server works now with this subdomain.

morph027:
> What I am talking about is: although apache reports internal server error, the nextcloud-signalilng-server works now with this subdomain.

Even with `/` ? So the talk app in browser works when joining a room and signal server is logging connections?

Anyway, I agree, there's no need to change the subdomain to the nextcloud domain.

szaimen:
> > What I am talking about is: although apache reports internal server error, the nextcloud-signalilng-server works now with this subdomain.
> 
> Even with `/` ? So the talk app in browser works when joining a room and signal server is logging connections?
> 
> Anyway, I agree, there's no need to change the subdomain to the nextcloud domain.

Yes, I think so. But I actually haven't looked in the logs.
But in the admin settings of talk, it reports, that the connection works...

enoch85:
OK, let's just keep the internal server error. We can disallow index alltogether and make it forbidden instead. At least that's cleaner.

enoch85:
Turns out we can't disallow index on the internal server error as there's no index. 

Any suggestions?

szaimen:
Yes, but maybe first should be comfirmed, that it really works?

enoch85:
> Yes, but maybe first should be comfirmed, that it really works?

Please login with your admin account and join the call. I'm waiting. :)

szaimen:
one moment please

enoch85:
JFTR, it works. :1st_place_medal:

szaimen:
okay nice, so it is definitely working, which the test confirmed. 🚀

morph027:
Awesome :rocket:

szaimen:
@morph027 do you have an idea maybe how to hide the internal server error message?

morph027:
server logs of the test vps shows in the logs:

```
Aug 24 21:28:22 ... signaling[55260]: 2020/08/24 21:28:22.910213 mcu_janus.go:303: WARNING: Full-Trickle is NOT enabled in Janus!
```

Will doublecheck the config

enoch85:
> do you have an idea maybe how to hide the internal server error message?

I have tried:

- Custom error page - Didn't work since there's an internal server error from the beginning
- Disallow Indexes - There's no index, so nothing to disallow.

szaimen:
I think I found a solution

szaimen:
I've changed:
```
    ProxyPass "/"  "ws://127.0.0.1:8081/"
    ProxyPassReverse "/"  "ws://127.0.0.1:8081/"
```

to the follwing:
```
    ProxyPass "/"  "http://127.0.0.1:8081/"
    ProxyPassReverse "/"  "http://127.0.0.1:8081/"
```

Now the internal server error is gone
and a custom index.html file should theoratically work

enoch85:
@szaimen Please join a new call for testing.

enoch85:
Nope, doesn't work.

szaimen:
> Nope, doesn't work.

Hm :/ maybe we should reload all services?
The admin panel still says that the connection works...

enoch85:
Thing is, it needs to proxy **w**eb**s**ocket, not http.

szaimen:
> Thing is, it needs to proxy websocket, not http.

Then this is of course the problem why we cannot show a html page if we just rewrite to websocket which is not http...

enoch85:
https://httpd.apache.org/docs/2.4/mod/mod_proxy_wstunnel.html

> Proxying both HTTP and websockets at the same time, with a specific set of URL's being websocket-only, can be done by specifying the websockets ProxyPass directive before the HTTP directive:

> ProxyPassMatch ^/(myApp/ws)$  ws://example.com:9080/$1
> ProxyPass / http://example.com:9080/

So, it would be possible to proxy http and ws at the same time. Question is, what are the paths to the `ws`? @morph027

enoch85:
@szaimen I think I solved it. 

Please join another call.

morph027:
> So, it would be possible to proxy http and ws at the same time. Question is, what are the paths to the `ws`?

The signaling server only does websockets, so the apache ws proxy needs to point to `ws://127.0.0.1:8081/`

enoch85:
@morph027 We just tested and it works.

@szaimen One last call to confirm the fix works. Thanks!

enoch85:
Talk test call is here: https://cloud.lindnercloud.com/call/j65ht5wt I'm waiting.

enoch85:
Credits to https://www.itsfullofstars.de/2017/03/custom-503-error-page-for-apache/ for the error document handling.

enoch85:
@morph027 

> The signaling server only does websockets, so the apache ws proxy needs to point to ws://127.0.0.1:8081/

The thing is that this config proxies all `ws` on `127.0.0.1:8081` and at the same time allows `http`.

Where are the logs for everything so that I can check if it actually works or not?

enoch85:
New link: https://cloud.lindnercloud.com/call/6qpcwsw9

This is just reinstalled with this PR script.

@morph027 @szaimen Please join.

morph027:
> Where are the logs for everything so that I can check if it actually works or not?

Signaling server logs can be viewed using `journalctl -lfu signaling` When everything works, there are a lot of lines showing established sessions, creating rooms and so on.

If it won't work, the talk app will show an error message anyways ;)

enoch85:
> If it won't work, the talk app will show an error message anyways ;)

As I said, it will only show if the signalling is installed and the proxy config is active. I'm fine with that.

I will try to put the logs in $VMLOGS now, after that I think we're done here. Do you have anything else?

morph027 (approved):
lgtm and test worked.

enoch85:
OK all done. Will merge this when travis thinks it looks good.

Thanks guys! Really really appreciated! :1st_place_medal:

enoch85:
@szaimen You can remove the domains now. 

Everything did some final tidying up, and everything is working like charm. :)

szaimen:
> @szaimen You can remove the domains now.

okay, both? (cloud.lindnercloud.com & test.lindnercloud.com)

enoch85:
Yes, if not in use ofc. :)

szaimen:
done 👍

enoch85:
https://www.techandme.se/its-your-turn-to-try-nextcloud-talk/

szaimen:
@enoch85 I've just tested it on my productive machine and there was a bug:
I head to set the domain of my signaling-server manually in the talk-admin-site.
To fix this, we have to move those two lines 
https://github.com/nextcloud/vm/blob/master/apps/talk_signaling.sh#L196
https://github.com/nextcloud/vm/blob/master/apps/talk_signaling.sh#L205
after this line:
https://github.com/nextcloud/vm/blob/master/apps/talk_signaling.sh#L233
Otherwise it worked perfectly 👍

enoch85:
Could you please make a PR? I'll merge it right away. Thanks!

enoch85:
Please fix the typo here as well: https://github.com/nextcloud/vm/blob/master/apps/talk_signaling.sh#L431

PR #1392, opened 2020-08-29
update domain_check_200 to allow CNAME tags

szaimen:


enoch85:
Thanks for the PR! 

It would be appreciated with a description. I have no idea what this means or does. Why is it even needed?

enoch85:
Don't use `echo` if not necessary, that's why we made the function. :smile:

szaimen:
Sry, I though the title was clear enough about this.

The problem is, that this command `dig +short "${1}" @resolver1.opendns.com` returns not only the ipaddress which should be stored inside the A-record but also the Domain, that it is referring to, if I use a CNAME DNS-record on a subdomain.
so e.g. if I have created a CNAME-tag for test.example.com and point to example.com, the command returns example.com and the ip-address of example.com and not only the ip-address of example.com. 

hope that makes it clear

szaimen:
> Don't use `echo` if not necessary, that's why we made the function. 😄

yes, but are you sure, that this works?
`if print_text_in_color "$IWhite" "$DIG" | grep -q "$WANIP4"`

because I am sure, that this works: (I am not really printing the return value of $DIG into the command line)
`if echo "$DIG" | grep -q "$WANIP4"`

another working possibility would be:
`if dig +short "${1}" @resolver1.opendns.com | grep -q "$WANIP4"`

enoch85:
> if dig +short "${1}" @resolver1.opendns.com | grep -q "$WANIP4"

That's the cleanest option of them all. Tested the print function, and it seemed to work though.

szaimen:
> > if dig +short "${1}" @resolver1.opendns.com | grep -q "$WANIP4"
> 
> That's the cleanest option of them all.

I've changed it to this 👍

PR #1394, opened 2020-08-30
add option to disable the admin-panel for bitwarden_rs

szaimen:
This is ready for review. It was already sucessfully tested.


enoch85:
Is this specific to RS?

I know that the official release also got an admin panel.

szaimen:
> Is this specific to RS?
> 
> I know that the official release also got an admin panel.

Yes, this is specific to bitwarden_rs.

For official bitwarden, if you leave the global-config adminSettings__admins empty, which it is by default, nobody will have access to the its admin panel. (you can change this already in the mailconfig script)

enoch85:
Aah ok, makes sense.

Didn't test my changes, but I think it looks cleaner.

If you can confirm it work, I'll merge. :+1:

szaimen:
> Didn't test my changes, but I think it looks cleaner.

your changes unfortunately don't work, but I found a working solution, that I just pushed to the branch.
So it is ready to go 👍

enoch85:
Strange since `" "` is used. But ok, it still looks better than before. :+1:

szaimen:
> Strange since `" "` is used. 

i can try to explain it to you, if you want

enoch85:
No worries. :)

PR #1396, opened 2020-09-01
introduce title-system, yesno_box, input_box, update smbmount

szaimen:
This PR does a few things:
1. allow a customizable title-system
2. introduce a yesno_box
3. create an input_box function
4. update smbmount to feature all points above

All changes are implemented so that they don't break anything anywhere else.
It was already tested successfully and should be ready to go.

### DIFF
--- a/apps/smbmount.sh
+++ b/apps/smbmount.sh
@@ -83,7 +83,7 @@ fi
 # Enter SMB-server and Share-name
 while true
 do
-    SERVER_SHARE_NAME=$(whiptail --title "$TITLE - $SUBTITLE" --inputbox "Please enter the server and Share-name like this:\n//Server/Share\nor\n//IP-address/Share" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    SERVER_SHARE_NAME=$(input_box "Please enter the server and Share-name like this:\n//Server/Share\nor\n//IP-address/Share" "$SUBTITLE")
     if ! yesno_box "Is this correct? $SERVER_SHARE_NAME" "$SUBTITLE"
     then
         msg_box "It seems like your weren't satisfied by the PATH you entered. Please try again." "$SUBTITLE"
@@ -96,7 +96,7 @@ done
 # Enter the SMB-user
 while true
 do
-    SMB_USER=$(whiptail --title "$TITLE - $SUBTITLE" --inputbox "Please enter the username of the SMB-user" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
+    SMB_USER=$(input_box "Please enter the username of the SMB-user" "$SUBTITLE")
     if ! yesno_box "Is this correct? $SMB_USER" "$SUBTITLE"
     then
         msg_box "It seems like your weren't satisfied by the SMB-user you entered. Please try again." "$SUBTITLE"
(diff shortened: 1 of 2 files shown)


szaimen:
@enoch85 do you have some feedback regarding this?
I am already quite satisfied with this PR...

enoch85:
I didn't have time to look at it yet. I thought about the same thing actually, so great move!

Just have to test it and so on...

enoch85:
First quick test, running the new function `msg_box` on anything that hasn't the new variables, will result in two messages.

IMHO this should be backwards compatible so that everything old works just as usual as well.

enoch85:
Sorry, my bad. Faulty testing.

Actually, there's no title at all...

szaimen:
> Actually, there's no title at all...

He?
The title variable is defined inside the lib here:
https://github.com/nextcloud/vm/pull/1396/files#diff-68f7bc5e839823a7568557cf9e87f4e2R77
and should automatically get inserted here:
https://github.com/nextcloud/vm/pull/1396/files#diff-68f7bc5e839823a7568557cf9e87f4e2R233

Did you maybe forgot to insert the title variable during testing?

enoch85:
> Did you maybe forget to insert the title variable during testing?

Well, there is no title variable in the old `msg_box`es. Or did I miss something?

szaimen:
> Well, there is no title variable in the old `msg_box`es. Or did I miss something?

Yes, there has been a title in the old msg_boxes (the title was` Nextcloud VM - $(date +"%Y")`):
https://github.com/nextcloud/vm/pull/1396/files#diff-68f7bc5e839823a7568557cf9e87f4e2L230

I just made the title a global variable so that we can use it in all places (like the new yesno_box and input_box but also in custom whiptails)

So if you haven't seen a title in the msg_boxes, something might be wrong?

enoch85:
Let me be more specific:

```
msg_box() {
    whiptail --title "$TITLE$SUBTITLE" --msgbox "$1" "$WT_HEIGHT" "$WT_WIDTH"
}

msg_box "You can't do this"
```

Results in this: https://i.imgur.com/05cAN9g.png

enoch85:
AAAH! I rest my case, I forgot to add the variable TILE which I now see you added. Sorry :)

enoch85:
Anyway, I think this PR is done, when all the occurrences is replaced with this new function. Not just the SMB script.

Or do you want to do that in a separate PR?

szaimen:
> Anyway, I think this PR is done, when all the occurrences is replaced with this new function. Not just the SMB script.

Acutally, I am not quite sure what you mean with replacing all occurrences, since everything should work like before. Maybe you could specify it a bit?

> Or do you want to do that in a separate PR?

Yes, a seperare PR would be better for this, imho.

apps/smbmount.sh
```diff
@@ -5,6 +5,7 @@
 
 # shellcheck disable=2034,2059
 true
+SCRIPT_NAME="SMB-mount"
```
enoch85, line 8: This is what I mean for the other scripts. We would need to add that.

apps/smbmount.sh
```diff
@@ -35,7 +36,7 @@ fi
 
 # Inform the user
 msg_box "This script automates mounting SMB-shares locally in your system and adds them automatically as external storage to your Nextcloud."
-if [[ "no" == $(ask_yes_or_no "Do you want to continue?") ]]
+if ! yesno_box "Do you want to proceed with this script?"
```
enoch85, line 39: This is also something that we would change. The new implementation is better imo

apps/smbmount.sh
```diff
@@ -89,10 +96,10 @@ done
 # Enter the SMB-user
 while true
 do
-    SMB_USER=$(whiptail --inputbox "Please enter the username of the SMB-user" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
-    if [[ "no" == $(ask_yes_or_no "Is this correct? $SMB_USER") ]]
+    SMB_USER=$(input_box "Please enter the username of the SMB-user" "$SUBTITLE")
```
enoch85, line 99: And imputbox function is also better...

So in other words, change were we use the old method. We don't want mixed code, remember? ;) :+1:

enoch85:
OK, let's do a separate PR.

Great work, thanks!

PR #1403, opened 2020-09-03
create input_box for TURN_PORT

szaimen:
This is just one detail, I wanted to fix since it has bothered me for a while.

I know, that there are also some inputs in static_ip.sh that we could probably switch to input_box, too.
Shall this be a part of this PR or do you want to leave static_ip.sh as it is?

### DIFF
--- a/apps/talk.sh
+++ b/apps/talk.sh
@@ -95,14 +95,7 @@ then
     while true
     do
     # Ask for port
-cat << ENTERDOMAIN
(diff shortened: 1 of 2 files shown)


enoch85:
Please fix the other occurances as well.

I plan to release today so it would be great if those could be fixed, like right now :D

enoch85:
Let's merge this so that it gets into the new release.

We can do the other inputboxes in separate PRs.

PR #1408, opened 2020-09-04
change all whiptail-radiolists to whiptail-menus

### LINKED ISSUE #1399: Change some implementations after #1396
Just as a reminder to change some implementations after #1396

- [x] 1. SCRIPT_NAME https://github.com/nextcloud/vm/pull/1396#discussion_r483057386 #1400
- [x] 2. yesno_box https://github.com/nextcloud/vm/pull/1396#discussion_r483057684 #1401
- [x] 3. input_box https://github.com/nextcloud/vm/pull/1396#discussion_r483058440 #1402

szaimen:
fix #1399

### DIFF
--- a/apps/adminer.sh
+++ b/apps/adminer.sh
@@ -21,9 +21,9 @@ root_check
 print_text_in_color "$ICyan" "Checking if Adminer is already installed..."
 if is_this_installed adminer
 then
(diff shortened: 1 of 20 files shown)


enoch85:
I think it's a bad idea for radiolist containing more options than one.

With this change you can only choose one at the time. Think the startup script, should the user not be able to choose several apps or configurations to install?

It's a good fit for the reinstall menu though.

szaimen:
> With this change you can only choose one at the time. Think the startup script, should the user not be able to choose several apps or configurations to install?

As I said, I am only changing the radiolists not the checklists.
and with radiolists, you weren't able to select more than one item before, so nothing changes in this regard.

enoch85:
Sorry my bad, I looked more closely now.

Issue #1405, opened 2020-09-04
Improve the way domain name is entered for Documentserver


enoch85:
As it is right now, you need to add cloud\.yourdomain\.com in the inputbox. Before the change you could just copy what was already written, or only change the parts that needed changing. Now there's a big risk of failure, even I failed as the check thought that I didn't use TLS on the main domain.

We should therefore make it possible to only type the domain itself, and codewise fix the \ \ \ stuff.

The user should only have to type the real domain.

We should also add a `yesno_box` to ask if the domain is correct.

enoch85:
Btw, the `yesno_box` also removes the "just hit enter through the whole installation to choose the defaults" since "yes" is always chosen. 

Is it possible to change that?

szaimen:
> Btw, the `yesno_box` also removes the "just hit enter through the whole installation to choose the defaults" since "yes" is always chosen.
> 
> Is it possible to change that?

Yes, with the `--defaultno` tag.
The question is, how we can insert this into the function.
BTW: I think it would be helpful, if all places, where no is the default would be linked here.

szaimen:
> Yes, with the `--defaultno` tag.
> The question is, how we can insert this into the function.

The only good solution that comes to my mind is if we add an additional function:
```
yesno_box_defaultno() {
    [ -n "$2" ] && local SUBTITLE=" - $2"
    if (whiptail --title "$TITLE$SUBTITLE" --defaultno --yesno "$1" "$WT_HEIGHT" "$WT_WIDTH")
    then
        return 0
    else
        return 1
    fi
}
```

enoch85:
Not tested, but what about this?

yesno_box() {
    [ -n "$3" ] && local DEFNO=" $3"
    [ -n "$2" ] && local SUBTITLE=" - $2"
    if (whiptail --title "$TITLE$SUBTITLE" $DEFNO --yesno "$1" "$WT_HEIGHT" "$WT_WIDTH")
    then
        return 0
    else
        return 1
    fi
}

szaimen:
> Not tested, but what about this?
> 
> ```
> yesno_box() {
>     [ -n "$3" ] && local DEFNO=" $3"
>     [ -n "$2" ] && local SUBTITLE=" - $2"
>     if (whiptail --title "$TITLE$SUBTITLE" $DEFNO --yesno "$1" "$WT_HEIGHT" "$WT_WIDTH")
>     then
>         return 0
>     else
>         return 1
>     fi
> }
> ```

[...]

PR #1409, opened 2020-09-04
add yesno_box_no and yesno_box_yes

szaimen:
POC, fix part of #1405
The function works as expected (was already tested) but the cases where it is needed are still missing (I don't know where it was no by default before)

### DIFF
--- a/apps/onlyoffice_docker.sh
+++ b/apps/onlyoffice_docker.sh
@@ -146,7 +146,7 @@ fi
 if [ "$(apache2ctl -M | grep evasive)" != "" ]
 then
     msg_box "We noticed that 'mod_evasive' is installed which is the DDOS protection for webservices. It has comptibility issues with OnlyOffice and you can now choose to disable it."
-    if ! yesno_box "Do you want to disable DDOS protection?"
+    if ! yesno_box_yes "Do you want to disable DDOS protection?"
     then
         print_text_in_color "$ICyan" "Keeping mod_evasive active."
     else
--- a/apps/onlyoffice_docker.sh
+++ b/apps/onlyoffice_docker.sh
@@ -146,7 +146,7 @@ fi
 if [ "$(apache2ctl -M | grep evasive)" != "" ]
 then
     msg_box "We noticed that 'mod_evasive' is installed which is the DDOS protection for webservices. It has comptibility issues with OnlyOffice and you can now choose to disable it."
-    if ! yesno_box "Do you want to disable DDOS protection?"
+    if ! yesno_box_defaultno "Do you want to disable DDOS protection?"
     then
         print_text_in_color "$ICyan" "Keeping mod_evasive active."
     else
--- a/apps/onlyoffice_docker.sh
+++ b/apps/onlyoffice_docker.sh
@@ -146,7 +146,7 @@ fi
 if [ "$(apache2ctl -M | grep evasive)" != "" ]
 then
     msg_box "We noticed that 'mod_evasive' is installed which is the DDOS protection for webservices. It has comptibility issues with OnlyOffice and you can now choose to disable it."
-    if ! yesno_box_defaultno "Do you want to disable DDOS protection?"
+    if ! yesno_box "Do you want to disable DDOS protection?"
     then
         print_text_in_color "$ICyan" "Keeping mod_evasive active."
     else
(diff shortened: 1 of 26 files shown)


enoch85:
You can check the defaults here: https://github.com/nextcloud/vm/pull/1401/files

szaimen:
> You can check the defaults here: https://github.com/nextcloud/vm/pull/1401/files

I've gone through the list and added the new function to all places, where I would say, that it makes sense.

szaimen:
Please wait before merging

szaimen:
I think it is ready now.

apps/onlyoffice_docker.sh
```diff
@@ -146,7 +146,7 @@ fi
 if [ "$(apache2ctl -M | grep evasive)" != "" ]
 then
     msg_box "We noticed that 'mod_evasive' is installed which is the DDOS protection for webservices. It has comptibility issues with OnlyOffice and you can now choose to disable it."
-    if ! yesno_box "Do you want to disable DDOS protection?"
+    if ! yesno_box_defaultno "Do you want to disable DDOS protection?"
```
enoch85, line 149: I think default should be yes here.

apps/onlyoffice_integrated.sh
```diff
@@ -172,7 +172,7 @@ fi
 if [ "$(apache2ctl -M | grep evasive)" != "" ]
 then
     msg_box "We noticed that 'mod_evasive' is installed which is the DDOS protection for webservices. It has comptibility issues with OnlyOffice and you can now choose to disable it."
-    if ! yesno_box "Do you want to disable DDOS protection?"
+    if ! yesno_box_defaultno "Do you want to disable DDOS protection?"
```
enoch85, line 175: Yes here as well

enoch85:
```
yesno_box() {

    # Check if yes or no
    if [ "$3" == "yes" ]
    then
        YESNO="--yesno"
    elif [ "$3" == "no" ]
    then
        YESNO="--defaultno --yesno"
    else
        YESNO="--yesno"
    fi

    # Final command
    if (whiptail --title "$TITLE$SUBTITLE" $YESNO "$1" "$WT_HEIGHT" "$WT_WIDTH")
    then
        return 0
    else
        return 1
    fi
}
```

I still don't get where `$2` comes from?

Travis will probably fail on the above since `$YESNO` can't be with `" "`, but we can work around it. It would be more neat to have everything in one function.

szaimen:
> I still don't get where `$2` comes from?

It is already used inside the smb-mount script and e.g. here:
https://github.com/nextcloud/vm/blob/2cc0f8a17a656ac3b2713d336fbaefd3a88d8609/apps/smbmount.sh#L87
The idea was to be able to add a custom Subtitle per msg_box, input_box or yesno_box in order create the possibility to customize the title further.

> Travis will probably fail on the above since `$YESNO` can't be with `" "`, but we can work around it. It would be more neat to have everything in one function.

This is what I meant when I thought about if this will work.

And BTW: I still think, that this:
`if yesno_box_defaultno "That is the message"`
looks much better than this
`if yesno_box "That is the message" "" "no"`

Or a bit differently: Why don't you want to add another short function and instead make the existing one more complicated?

apps/onlyoffice_docker.sh
szaimen: done

apps/onlyoffice_integrated.sh
szaimen: done

szaimen:
A different idea: what about changing the name of` yesno_box_defaultno` to `yesno_box_no` ? 
In order to be consistent, we could rename `yesno_box` then to `yesno_box_yes`
What do you think about that?

enoch85:

That's better, but can't be merged until next release as the `static_ip` use the old function and the `lib` file is also containing the old function.

But yes, I think that's better.

So, let's change `defaultno` to `no` for now, and change the name of the regular `yesno_box_yes` later.

szaimen:
I will work on this tomorrow. Don't have time today

szaimen:
I think this is ready now.
I have renamed yesno_box_defaultno to yesno_box_no.
Additionally, I have created yesno_box_yes and renamed all appearances of yesno_box to this.
But there should be no problem since I have not deleted the current function yesno_box and we can just delete it with NC20 or so.

szaimen:
Is there anything missing here or is there still something to discuss?

enoch85:
> Is there anything missing here or is there still something to discuss?

Nope, I think it looks good. :+1:

enoch85:
Tested the current VM (19.0.2) without internet, and all is good. It doesn't bug out after this merge.

PR #1410, opened 2020-09-04
standardize checklist-instructions and menu-instructions

szaimen:
we can use the CHECKLIST_GUIDE variable now whenever we need checklist-instructions
and the new MENU_GUIDE can be used for every MENU script.


enoch85:
LGTM! :+1:

PR #1411, opened 2020-09-06
get the correct NCDOMAIN from overwrite.cli.url

### LINKED ISSUE #1405: Improve the way domain name is entered for Documentserver

szaimen:
should fix #1405

### DIFF
--- a/apps/collabora_docker.sh
+++ b/apps/collabora_docker.sh
@@ -157,10 +157,10 @@ apt update -q4 & spinner_loading
 
 # Check if Nextcloud is installed
 print_text_in_color "$ICyan" "Checking if Nextcloud is installed..."
-if ! curl -s https://"${NCDOMAIN//\\/}"/status.php | grep -q 'installed":true'
+if ! curl -s https://"$NCDOMAIN"/status.php | grep -q 'installed":true'
 then
 msg_box "It seems like Nextcloud is not installed or that you don't use https on:
-${NCDOMAIN//\\/}.
+$NCDOMAIN.
 Please install Nextcloud and make sure your domain is reachable, or activate TLS
 on your domain to be able to run this script.
 
@@ -184,7 +184,7 @@ install_docker
 
 # Install Collabora docker
 docker pull collabora/code:latest
-docker run -t -d -p 127.0.0.1:9980:9980 -e "domain=$NCDOMAIN" --restart always --name code --cap-add MKNOD collabora/code
+docker run -t -d -p 127.0.0.1:9980:9980 -e "domain=$NCDOMAIN_ESCAPED" --restart always --name code --cap-add MKNOD collabora/code
 
 # Install Apache2
 install_if_not apache2
@@ -237,7 +237,7 @@ then
   Header set X-XSS-Protection "1; mode=block"
   Header set Strict-Transport-Security "max-age=31536000; includeSubDomains; preload"
   Header set X-Content-Type-Options nosniff
-  Header set Content-Security-Policy "frame-ancestors 'self' ${NCDOMAIN//\\/}"
+  Header set Content-Security-Policy "frame-ancestors 'self' $NCDOMAIN"
 
   # keep the host
   ProxyPreserveHost On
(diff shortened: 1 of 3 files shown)


szaimen:
@enoch85 I haven't tested the whole scripts but it should already work.

enoch85:
We should do it like we do in OnlyOffice for the variable (https://github.com/nextcloud/vm/blob/master/apps/onlyoffice_docker.sh#L163) and remove from lib.

apps/onlyoffice_docker.sh
```diff
@@ -160,7 +160,7 @@ fi
 # OnlyOffice URL (onlyoffice.sh)
 SUBDOMAIN=$(input_box "OnlyOffice subdomain e.g: office.yourdomain.com\n\nNOTE: This domain must be different than your Nextcloud domain. They can however be hosted on the same server, but would require seperate DNS entries.")
 # Nextcloud Main Domain (onlyoffice.sh)
-NCDOMAIN=$(input_box "Nextcloud domain, make sure it looks like this: cloud\\.yourdomain\\.com")
+NCDOMAIN=$(occ_command_no_check config:system:get overwrite.cli.url | sed 's#https://##;s#/##')
```
enoch85, line 163: What if that value isn't set? Better to hardcode it if you ask me.

szaimen: The overwrite.cli.url must be set, else Nextcloud doesn't work correctly. So I am sure that this works.

enoch85: If the user haven't run the LE script the domain is 'localhost'. Have you tested that as well so that it doesn't bug out if nothing is set to something not being a domain?

szaimen: No, I haven't tested this detail. But my testing in the past has showed that the curl command always fails if self singed certificates are used which is the case here.

szaimen: I've tested this just now and can confirm my previous comment

szaimen:
> We should do it like we do in OnlyOffice for the variable (https://github.com/nextcloud/vm/blob/master/apps/onlyoffice_docker.sh#L163) and remove from lib.

done 👍

enoch85:
OK, so is this ready to merge?

szaimen:
I would say yes, but still haven't tested the whole scripts.

enoch85:
Just tested the varaiable, and it doesn't work since it needs to be `\\` and not `\`.

apps/collabora_docker.sh
```diff
@@ -142,12 +142,16 @@ then
     occ_command app:remove onlyoffice
 fi
 
-# shellcheck disable=2034,2059
-true
+# Collabora Docker URL (collabora.sh
+SUBDOMAIN=$(input_box "Collabora subdomain eg: office.yourdomain.com\n\nNOTE: This domain must be different than your Nextcloud domain. They can however be hosted on the same server, but would require seperate DNS entries.")
+# Nextcloud Main Domain (collabora.sh)
+NCDOMAIN=$(occ_command_no_check config:system:get overwrite.cli.url | sed 's#https://##;s#/##')
```
enoch85, line 148: Please use | instead of # as we do that everywhere else.

apps/onlyoffice_docker.sh
enoch85, line 163: | instead of # please

enoch85:
When comments are fixed, I will merge.

apps/collabora_docker.sh
szaimen: fixed

apps/onlyoffice_docker.sh
szaimen: fixed

PR #1412, opened 2020-09-06
fix first internet check on startup-script and static_ip and remove it from server_configuration

szaimen:
This fixes 2 points:
- The network check for server_configuration is not needed, so I removed it. 
If I am wrong with that statement, please tell me why.
- The first network check for static_ip and startup-script was wrong because it would always use the lib if it was existing and thus never use the newest lib also not if you had a working internet-connection from the beginning.

### DIFF
--- a/menu/server_configuration.sh
+++ b/menu/server_configuration.sh
@@ -2,6 +2,12 @@
 
 # T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
 
+# shellcheck disable=2034,2059
+true
+SCRIPT_NAME="Server Configuration Menu"
+# shellcheck source=lib.sh
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
 # 0 = OFF
--- a/menu/server_configuration.sh
+++ b/menu/server_configuration.sh
@@ -2,27 +2,6 @@
 
 # T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
 
-# If we have internet, then use the latest variables from the lib remote file
-if printf "Testing internet connection..." && ping github.com -c 2
-then
-# shellcheck disable=2034,2059
-true
-SCRIPT_NAME="Server Configuration Menu"
-# shellcheck source=lib.sh
-. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
-# Use local lib file in case there is no internet connection
-elif [ -f /var/scripts/lib.sh ]
-then
-# shellcheck disable=2034,2059
-true
-# shellcheck source=lib.sh
-source /var/scripts/lib.sh
-else
-    printf "You don't seem to have a working internet connection, and /var/scripts/lib.sh is missing so you can't run this script."
-    printf "Please report this to https://github.com/nextcloud/vm/issues/"
-    exit 1
-fi
-
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
 # 0 = OFF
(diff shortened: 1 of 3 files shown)


szaimen:
I have a short question concerning the startup-script.
It might be a bit off-topic but I think it fits kind of:
https://github.com/nextcloud/vm/blob/e255051d962de5be642ce8066281da83d6417745/nextcloud-startup-script.sh#L170-L171
Is this reboot really only needed to apply the new keyboard settings as the explanation says or does this do more?

enoch85:
We need that in the server configuration menu as static_ip is included there and that should be possible to run without internet.

enoch85:
Regarding the current setup in the startup script I think that's safer. We can expect that 99% of the users doesn't have an internet connection during the first run since:
1. The `iface` is wrong
2. DHCP is not enabled in most networks

enoch85:
But sure, as we test the internet connection with ping, I'd say this is the proper way of doing it.

enoch85:
Please revert the change sin `server_configuration.sh` and I'll review and merge.

szaimen:
If the internet really doesn't work, you get the possibility to execute the static_ip script directly [here](https://github.com/nextcloud/vm/blob/e255051d962de5be642ce8066281da83d6417745/nextcloud-startup-script.sh#L95)
without having to go over the server_condiguration script, no?

enoch85:
> no?

It would actually be better to add the same logic in all the menu scripts as you did for `static_ip.sh`

I've been thinking if we should make the whole thing a separate lib file and include that when running the installer. The add something like `source /var/scripts/fetch_libs.sh` in top of each script instead of adding the logic to each script.

`/var/scripts/fetch_libs.sh` would the contain this and only this:

```
# Use local lib file in case there is no internet connection
if [ -f /var/scripts/lib.sh ]
then
true
SCRIPT_NAME="Nextcloud First Startup Script"
# shellcheck source=lib.sh
NCDB
=1 && FIRST_IFACE=1 source /var/scripts/lib.sh
unset NCDB
unset FIRST_IFACE
 # If we have internet, then use the latest variables from the lib remote file
elif printf "Testing internet connection..." && ping github.com -c 2
then
true
# shellcheck source=lib.sh
NCDB=1 && FIRST_IFACE=1 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
unset FIRST_IFACE
unset NCDB
else
    printf "You don't seem to have a working internet connection, and /var/scripts/lib.sh is missing so you can't run this script."
    printf "Please report this to https://github.com/nextcloud/vm/issues/"
    exit 1
fi
```
Where tags like `NCDB` and `IFACE` would be replaced with `$1` `$2` or similar. Not thought through yet, but you get the point. 

Maybe we could add the `NCDB` and `IFACE` in two places as well - this lib and the default one, or maybe only in this?

Then we would only have to change the logic in one file and always be sure that the lib file is working since we would either use the local or the online one.

szaimen:
In my opinion this is better for another PR then handling this here. I also don't get what this shall do currently. We can discuss this then in a new issue. I can create one, if you want?

BTW: the logic of executing the static_ip script via server_configuration.sh can also not work since the server_configration.sh script is just downloaded during the startup-script here: https://github.com/nextcloud/vm/blob/e255051d962de5be642ce8066281da83d6417745/nextcloud-startup-script.sh#L249
So I still think that my logic is correct. and the better way is executing static_ip directly if the internet is really not working.

szaimen:
And do you have an answer for this?
https://github.com/nextcloud/vm/pull/1412#issuecomment-687838324

enoch85:
> And do you have an answer for this?

I actually never tested myself.

We used setupcon before, but people where still complaining about that the keyboard layout was wrong, so I just added reboot to be sure. Since then, no further complaints regarding that.

enoch85:
OK, let's do the other stuff for a different PR.

I'll merge this and see what response we get.

szaimen:
> We used setupcon before, but people where still complaining about that the keyboard layout was wrong, so I just added reboot to be sure. Since then, no further complaints regarding that.

Okay, I've already suspected something like this and I have a solution which doesn't necessarily requires a reboot ;)

Issue #1413, opened 2020-09-06
cache the lib for half an hour?


szaimen:
I don't know if it makes sense but this is a working POC how we could cache the lib for have an hour, if we run this at the beginning of every script (and don't have to download the whole lib again for every script for have an hour).

```
if ! [ -f /var/scripts/lib.sh ] || test "`find /var/scripts/lib.sh -mmin +30`"
then
    rm -f /var/scripts/lib.sh
    curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh
fi
```
We could then always use this inside the script itself:
```
# shellcheck source=lib.sh
source /var/scripts/lib.sh
```

szaimen:
I am closing this since it apparently isn't something that we want to do.

PR #1414, opened 2020-09-06
chain libraries

szaimen:
I've just found out that we can chain libaries.
This implements #1413

We could just always use the /var/scripts/get.sh file in all scripts as local lib then.
```
# shellcheck source=get.sh
source /var/scripts/get.sh
```
I am still not sure if it makes sense but it is kind of cool and could bring advantages like having a small header
One problem is that I don't know how we can migrate the users to this new local get.sh...


nextcloud-startup-script.sh
```diff
@@ -246,7 +246,6 @@ download_script STATIC setup_secure_permissions_nextcloud
 download_script STATIC change_db_pass
 download_script STATIC nextcloud
 download_script NETWORK update-config
-download_script NETWORK trusted
 download_script MENU menu
 download_script MENU server_configuration
 download_script MENU nextcloud_configuration
```
szaimen, line 251: Could we somehow work around the initial downloading? then we wouldn't have to track which script is already downloaded (and execute the downloaded one) but instead always run the newest script from the repo...
(which should theoratically be no problem since when the user is at this point the internet of the VM is working)

enoch85: That would make the startup script more prune to errors. Is that what we want?

szaimen: Probably not...

szaimen:
> Is this what I was talking about with fetch_lib?

Yes, I guess so. We could just rename it to` fetch_lib`

enoch85:
> fetch_lib

Please do. :+1:

szaimen:
the script was renamed.
The question is if we do it than that way and just use 
```
# shellcheck source=get.sh
source /var/scripts/get.sh
```
everywhere?

szaimen:
> The question is if we do it than that way and just use the following everywhere?
> 
> ```
> # shellcheck source=get.sh
> source /var/scripts/get.sh
> ```

Though If I think about this, this might be the better approach (to cover the cases where the local fetch_lib isn't available):
```
# shellcheck source=lib.sh
source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/lib.sh) &>/dev/null
```

enoch85:
Great job, though I'm still wondering about all the `WHATEVER=1` stuff?

As I wrote earlier, if we call `fetch_lib` instead, we need to make sure that all the called variables still work specific to each script.

szaimen:
> Great job, though I'm still wondering about all the `WHATEVER=1` stuff?
> 
> As I wrote earlier, if we call `fetch_lib` instead, we need to make sure that all the called variables still work specific to each script.

I've already tested this and since we are chaining the source files, it works as expected :D

But actually before merging this PR, I would still change all the WHATEVER=1 stuff to functions. And this time I am confident that it will work since we we have already abandoned all problematic parts from the library. This will be the next PR.

enoch85:
Now when all WHATEVER=1 are functions, this could be merged in all files.

So, please replace the `curl` and `source` to this fetch_lib instead.

enoch85:
Please do it like you proposed:

```
# shellcheck source=fetch_lib.sh
source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/lib.sh) &>/dev/null
```

That way we can merge it now and remove the curl command once a few releases has passed.

static/fetch_lib.sh
```diff
@@ -0,0 +1,42 @@
...
+        mkdir -p /var/scripts
+        if ! curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh
+        then
+            lib_error_message
+        fi
+    fi
+}
+
+# Run the script
+if [ -f /var/scripts/nextcloud-startup-script.sh ] && ! [ -f "$SCRIPTS/you-can-not-run-the-startup-script-several-times" ]
```
szaimen, line 26: @enoch85 does that make sense to you? (I just added the second condition)

static/fetch_lib.sh
```diff
@@ -0,0 +1,42 @@
...
+        mkdir -p /var/scripts
+        if ! curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh
+        then
+            lib_error_message
+        fi
+    fi
+}
+
+# Run the script
+if [ -f /var/scripts/nextcloud-startup-script.sh ] && ! [ -f /var/scripts/you-can-not-run-the-startup-script-several-times ]
```
szaimen, line 26: Does that make sense to you? (I just added the second condition)

enoch85: I don't understand what that does. Why do we need to check if the startup script?

fetch_lib should only fetch either the local lib or curl if we have internet.

szaimen: because we have to differentiate between the startup-script and all other cases.
Or do you just want to check always if the internet works before continuing with any script?
(doesn't make really sense, imo since many scripts get downloaded before they get executed.)

szaimen: maybe we should talk about what it should do:
Currently it tests the internet connection for every script during the startup-script, as long as the you-can-not-run-the-startup-script-twice file isn't there.
If one of these cases aren't met, it always downloads the new library from github, if the local library doesn't exist or if the local library is older than 30 minutes. If one of those cases is met, it will first remove the local lib and will try to download the new lib, and will report and exit if this download hasn't succeeded.

enoch85: The idea here is to avoid having to type like 20 lines

```
# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/

# If we have internet, then use the latest variables from the lib remote file
if printf "Testing internet connection..." && ping github.com -c 2
then
true
SCRIPT_NAME="Server configuration"
# shellcheck source=lib.sh
# Use local lib file in case there is no internet connection
elif [ -f /var/scripts/lib.sh ]
then
true
# shellcheck source=lib.sh
source /var/scripts/lib.sh
else
    printf "You don't seem to have a working internet connection, and /var/scripts/lib.sh is missing so you can't run this script."
    printf "Please report this to https://github.com/nextcloud/vm/issues/"
    exit 1
fi
```

And only do:

```
# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/

true
# shellcheck source=fetch_lib.sh
```

In the startup script we use something totally different, so leave that out of `fetch_lib` from that one and do it everywhere else as the logic for the startup script only happens there.

enoch85: Also, why do we want to cache the lib file? Don't we always want to use the latest and greatest?

szaimen: okay, my idea was to also cover these lines, so I guess this is where our solutions diverged 😅

szaimen: > Also, why do we want to cache the lib file? Don't we always want to use the latest and greatest?

I though about if it would make sense to cache the lib for 30min?
I am not sure if that really makes sense, though.

szaimen: So am I right, that you just want the following code in fetch_lib:
```
#!/bin/bash
true
# see https://github.com/koalaman/shellcheck/wiki/Directive

# shellcheck source=lib.sh
source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/lib.sh) &>/dev/null
```

szaimen: > # T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
> 
> # If we have internet, then use the latest variables from the lib remote file
> if printf "Testing internet connection..." && ping github.com -c 2
> then
> # shellcheck disable=2034,2059
> true
> SCRIPT_NAME="Server configuration"
> # shellcheck source=lib.sh
> . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
> # Use local lib file in case there is no internet connection
> elif [ -f /var/scripts/lib.sh ]
> then
> # shellcheck disable=2034,2059
> true
> # shellcheck source=lib.sh
> source /var/scripts/lib.sh
> else
>     printf "You don't seem to have a working internet connection, and /var/scripts/lib.sh is missing so you can't run this script."
>     printf "Please report this to https://github.com/nextcloud/vm/issues/"
>     exit 1
> fi

This is exactly what we do in the startup script, no?

szaimen:
@enoch85 maybe you could create an example file in this branch where you show what you imagine should be in the fetch_lib file?

static/fetch_lib.sh
szaimen: Sry, I apparently don't get it, yet.

enoch85:
> maybe you could create an example file in this branch where you show what you imagine should be in the fetch_lib file?

Done.

So, we should use one lib to fetch either the local lib or the online lib with curl.

szaimen:
Great! Thanks for showing your vision. 👍

Question is, if we should update the local lib somehow and if we really want to test the internet connection each time any script runs? Is that what you want?

I think my version is already a bit more advanced but I in conclusion I think our goal is the same.

What do you think?

static/fetch_lib.sh
```diff
@@ -0,0 +1,42 @@
...
+    exit 1
+}
+
+download_cache_lib() {
+    # Cache the lib for half an hour
+    if ! [ -f /var/scripts/lib.sh ] || test "$(find /var/scripts/lib.sh -mmin +30)"
+    then
+        rm -f /var/scripts/lib.sh
+        mkdir -p /var/scripts
+        if ! curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh
```
szaimen, line 18: question is, it is possible that the curl command succeeds, if the file already exists to which I want to curl another file?
If yes, I think this does already everything we want.

szaimen: Yes, it works. I've just tested this.

static/fetch_lib.sh
```diff
@@ -0,0 +1,21 @@
...
+        printf "The local lib isn't available and you don't have internet access. You cannot run this script"
+        exit 1
+    fi
+elif test "$(find /var/scripts/lib.sh -mmin +30)"
+then
+    curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh
+fi
+
+# shellcheck source=lib.sh
+source /var/scripts/lib.sh &>/dev/null
```
szaimen, line 21: @enoch85 I just redid my whole version. I hope it is better understandable now.

enoch85: We pushed at the same time. Check the latest.

szaimen: I think this does the same but is much more simplistic which I really like. :)

szaimen: @enoch85 I think the biggest advantage, is that we don't have to check the internet each time. We just try to fetch it and if it doesnt work, the internet isn't working correctly. If the file exists, we just try to fetch a new one (if it is older than 30mins in this case to have the file always up do date. If that doesnt work here, it is no problem since we just use the local lob either way

menu/main_menu.sh
```diff
@@ -2,16 +2,25 @@
...
 then
     curl -so /var/scripts/fetch_lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/static/fetch_lib.sh
 fi
-SCRIPT_NAME="Main Menu"
+
+###########################################################################
+# shellcheck disable=2034,2059
+true
+# shellcheck source=fetch_lib.sh
+if ! source /var/scripts/fetch_lib.sh >/dev/null 2>&1
```
szaimen, line 18: this won't work
and I really think, that this does look better:
```
# shellcheck source=fetch_lib.sh
source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/lib.sh) &>/dev/null
```

addons/locales.sh
```diff
@@ -2,16 +2,25 @@
...
+###########################################################################
+# shellcheck disable=2034,2059
+true
+# shellcheck source=fetch_lib.sh
+if ! source /var/scripts/fetch_lib.sh >/dev/null 2>&1
+then
 # shellcheck source=lib.sh
 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+fi
+###########################################################################
```
szaimen: @enoch85 can we please discuss this one time before doing this to all scripts?

menu/main_menu.sh
enoch85: OK, please change

addons/locales.sh
enoch85: Sure, https://cloud.hanssonit.se/call/3ejfaazp

enoch85:
Let's merge it for those two scripts to begin with, all the other changes are fine I think.

That way we download the fetch_lib file on every menu run as well, and not only when updating which will make more people have it.

Since the file is very important, maybe we should hide it somewhere were it's not obvious i.e not in /var/scripts, what do you think?

static/fetch_lib.sh
enoch85: Please join the chat: https://cloud.hanssonit.se/call/3ejfaazp

enoch85: ping @szaimen

szaimen:
> Since the file is very important, maybe we should hide it somewhere were it's not obvious i.e not in /var/scripts, what do you think?

No, I don't think that this is necessary. They should never delete manually anything in /var/scripts So we just don't start with hiding files.

And can we please discuss, whether we can use this method, because I still think, that it makes more sense this way (not testing the internet every time, I also removed the caching; we could of course enhance this version with a msg_box and print_text_in_color)
```
#!/bin/bash
true
# see https://github.com/koalaman/shellcheck/wiki/Directive


mkdir -p /var/scripts
if ! [ -f /var/scripts/lib.sh ]
then
    if ! curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh 
    then
        printf "The local lib isn't available and you don't have internet access. You cannot run this script"
        exit 1
    fi
else
    curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh
fi

# shellcheck source=lib.sh
source /var/scripts/lib.sh &>/dev/null
```

szaimen:
I will now fix travis

enoch85:
@szaimen I tried pinging you three times now for the talk session: https://cloud.hanssonit.se/call/3ejfaazp ;)

szaimen:
> @szaimen I tried pinging you three times now for the talk session: https://cloud.hanssonit.se/call/3ejfaazp ;)

okay, I am in

szaimen:
@enoch85 I would start here and improve it
```
#!/bin/bash
true
# see https://github.com/koalaman/shellcheck/wiki/Directive

mkdir -p /var/scripts
if ! [ -f /var/scripts/lib.sh ]
then
    if ! curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh 
    then
        printf "You don't seem to have an internet connection and the local lib isn't available. Hence you cannot run this script."
        exit 1
    fi
else
    printf "Updating lib..."
    curl -so /var/scripts/lib.sh https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh
fi

# shellcheck source=lib.sh
source /var/scripts/lib.sh &>/dev/null
```

szaimen:
So what about caching and please tell what you would change regarding informing the user. Because I think we don't need special print_text_in_color or msg_box here. printf should be enough, no?

enoch85:
> So what about caching

No caching now, let's test this first for a while.

Ready to merge of you ask me.

szaimen:
@enoch85 yes, looks much better, thank you.

szaimen (commented):
For me this is good enough for now, if travis is fixed.
(we can still change some small details later)

PR #1416, opened 2020-09-07
fix network_ok

szaimen:
This should harden the check here:
https://github.com/nextcloud/vm/blob/cbc20d1881f1341db2b8b231b83acfdc611278bb/nextcloud-startup-script.sh#L69

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -824,7 +824,10 @@ then
     then
         # sleep 40 seconds so that some slow networks have time to restart
         countdown 'Not online yet, waiting a bit more...' 40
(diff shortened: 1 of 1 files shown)


szaimen:
It seems like this isn't neede since it just returns the value from site_200

enoch85:
@szaimen What was this?

enoch85:
Please remove the branch if you don't plan to do anything further with this.

szaimen:
> Please remove the branch if you don't plan to do anything further with this.

done

PR #1418, opened 2020-09-07
ask if the domain is correct

enoch85:
@szaimen I tested this and it worked, but the first time I entered the wrong domain and the script failed ofc. We should ask if it's correct before that happens

### DIFF
--- a/apps/bitwarden-rs.sh
+++ b/apps/bitwarden-rs.sh
@@ -57,7 +57,15 @@ then
     exit
 fi
 
-SUBDOMAIN=$(input_box "Please enter the Domain that you want to use for Bitwarden RS.")
+# Ask for domain
+while :
+do
+    SUBDOMAIN=$(input_box "Please enter the Domain that you want to use for Bitwarden RS.")
+    if yesno_box_yes "Is this correct? $SUBDOMAIN"
+    then
+        break
+    fi
+done
 
 # curl the lib another time to get the correct https_conf
 # shellcheck source=lib.sh
--- a/apps/bitwarden_mailconfig.sh
+++ b/apps/bitwarden_mailconfig.sh
@@ -56,7 +56,7 @@ then
 fi
 
 # Enter mailserver
-while true
+while :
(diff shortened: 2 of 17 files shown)


szaimen:
@enoch85 yes, you are right. Shouldn't this also get changed in other places where we ask for a subdomain?

enoch85:
Oops, I merged the other change to master. :D

enoch85:
Here: https://github.com/nextcloud/vm/commit/20dd9543c3d9dbe0b43e52a76a708e4022d09bd2

enoch85:
@szaimen Anything I missed?

enoch85:
We could do the same thing for the LE Revoke stuff, but that feels a bit much... Or what do you think?

szaimen:
> We could do the same thing for the LE Revoke stuff, but that feels a bit much... Or what do you think?

No, I think this is a good idea

szaimen:
> @szaimen Anything I missed?

I think there is another one in bitwarden_rs...

enoch85:
I'll fix Collabora and OnlyOffice, can you take care of the rest?

enoch85:
I can fix everyting, just a sec.

szaimen:
> I can fix everyting, just a sec.

Nice 🚀

enoch85:
OK done, do you think it looks good @szaimen?

szaimen:
I think it looks good but for revoking is the question still missing. (I think in all documentserver-scripts)
I would suggest seaching for "SUDOMAIN" to find all places

enoch85:
Aaah, forgot about integrated.

szaimen:
Yes, integrated was missing but there are still additional places in the other documentserver scripts, too.
(We ask around 3 times in docker and 2 times in integrated if I remember correctly)

szaimen:
@enoch85 shall I open a new issue so that we don't forget this?

enoch85:
Please do, or make a PR directly ;)

PR #1419, opened 2020-09-07
fix yesno on some questions

enoch85:
cc @szaimen FYI


enoch85:
@szaimen We can't use `print_text_in_color` as that won't be shown before the `yesno_box`.

I don't have time to check more occurrences now, but keep an eye on this please.

PR #1421, opened 2020-09-07
fix last subdomain occurences and remove print_text_in_color before a yesno_box

### LINKED ISSUE #1420: fix the missing subdomains and ask for confirmation
Yes, integrated was missing but there are still additional places in the other documentserver scripts, too.
(We ask around 3 times in docker and 2 times in integrated if I remember correctly)

_Originally posted by @szaimen in https://github.com/nextcloud/vm/pull/1418#issuecomment-688349162_

szaimen:
fix #1420 by fixing the last subdomain occurences
and remove the only the only print_text_in_color occurences before a yesno_box that I could still find


enoch85:
Thanks!

PR #1422, opened 2020-09-07
change order of server_configuration and automatically set the activate_tls_switch

szaimen:
one of the successors to #1417
it does what the title says
I think this is a better order


menu/server_configuration.sh
```diff
@@ -82,6 +80,16 @@ https://www.techandme.se/open-port-80-443/"
         rm -f "$SCRIPTS/activate-tls.sh"
```
szaimen, line 80: Is this removal needed? I would just remove the file, if tls was activated, no?

szaimen: It gets removed here:
https://github.com/nextcloud/vm/blob/4a6402ebfd7c5e642091643094988db216c40fa2/lets-encrypt/test-new-config.sh#L78
So I think the removal isn't necessary at this  place before the actual activation of tls

enoch85: It would ensure to always get the latest script, but at the same time it's actually not me who wrote it ;)

To me it doesn't matter really, we could do run_script instead, but then it would remove the logic in the startup script as we want all scripts to be pre-feteched to ensure a complete run.

enoch85: Aah, I remember now why they are there. Sometimes it bugs out and the scripts aren't removed. 

It's just there for cleaning purposes. Please leave it as is.

menu/server_configuration.sh
```diff
@@ -17,41 +17,39 @@ debug_mode
...
 root_check
 
+# Set the correct switch for activate_tls
+if [ -f $SCRIPTS/activate-tls.sh ]
+then
+    ACTIVATE_TLS_SWITCH="ON"
+else
+    ACTIVATE_TLS_SWITCH="OFF"
+fi
+
```
enoch85, line 27: Why do we need a switch?

szaimen: Because it was on by default but only really makes sense if tls wasn't already activated.

Because of that check (if the activate-tls file is still there) I also don't want to remove the activate-tls file if tls wasn't successfully activated

szaimen: I mean I could of course also come up with a better check. What do you think?

menu/server_configuration.sh
szaimen: Please see the discussion below. It is related

menu/server_configuration.sh
enoch85: Aah I get it, good idea. 👍🏻

menu/server_configuration.sh
```diff
@@ -76,12 +74,18 @@ https://www.techandme.se/open-port-80-443/"
             print_text_in_color "$ICyan" "OK, but if you want to run it later, just type: sudo bash $SCRIPTS/activate-tls.sh"
             any_key "Press any key to continue..."
         fi
-
-        # Just make sure they are gone
-        rm -f "$SCRIPTS/test-new-config.sh"
```
enoch85, line 81: Please add back rm for the test script.

Thanks!

szaimen: done 👍

PR #1423, opened 2020-09-07
create startup_configuration.sh

szaimen:
This is the second successor to #1417

Motivation: I think it is nicer to put all those startup-options into another menu.
We could then also think about putting this as an additional option into main_menu.

Disclaimer: I haven't tested the complete PR but at least the new menu works in my testing.


szaimen:
@enoch85 BTW 😉
image

menu/startup_configuration.sh
```diff
@@ -0,0 +1,108 @@
...
+else
+    KEYBOARD_LAYOUT_SWITCH="OFF"
+fi
+
+# Get the correct locate mirror switch
+if [ -z "$REPO" ]
+then
+    REPO=$(apt update -q4 && apt-cache policy | grep http | tail -1 | awk '{print $2}')
+fi
+if [ "$REPO" = "http://se.archive.ubuntu.com/ubuntu" ]
```
enoch85, line 33: Actually, in the newer releases we use archive.ubuntu.com.

And also, from a stability perspective, I'd say that it's better to actually *not* run that script. It's just in edge cases where you live like on the other side of the world from the mirrors on slow hardware or a bad WAN. 

As it said in the description before the question, it might not work and we can't guarantee that the mirrors will stay up for the lifetime of the server.

szaimen: > Actually, in the newer releases we use archive.ubuntu.com.

I've tested it with a new VM release (19.0.1) and the server I am checking for was the result.

I only activated this option because you will presented with the question if you really want to run this script which is no by default:
https://github.com/nextcloud/vm/blob/3e5947b755a43c12d40c694cacf25fc9a6cdf382/addons/locate_mirror.sh#L28

But I mean  I can just set this option always to 'No' if you like that more.

szaimen: I've removed that switch and set it to OFF. also I made it the last option

szaimen:
> We could then also think about putting this as an additional option into main_menu.

What do you think about that?

enoch85:
I see one major issue with this PR - what if setupcon only works for the current session (being this script)?

That would mean that the next script still have the old setting - that's also why I rebooted as default in the previous setup.

enoch85:

Sure, if this gets merged.

szaimen:
> what if setupcon only works for the current session (being this script)

I will try that one out.

szaimen:
> > what if setupcon only works for the current session (being this script)
> 
> I will try that one out.

I've tested this and it works as expected. Apparently `setupcon` seems to set the keyboard layout globally and not only for one script.

szaimen:
> > > We could then also think about putting this as an additional option into main_menu.
> >
> > What do you think about that?
> 
> Sure, if this gets merged.

I've put it into main_menu. Question is, if we want another order? e.g. startup_configuration is above server_configuration?

And also: I've put your description of the startup_configuration into the startup_script since it doesn't make sense, if you run this script later, imo. Hope your agree that it makes more sense this way?

szaimen:
> I've put your description of the startup_configuration into the startup_script since it doesn't make sense, if you run this script later, imo. Hope your agree that it makes more sense this way?

I mean, we could alternatively add a check there if the startup-script is available and only show it then, so that we are able to leave it in the menu. It could look like this:
```
if [ -f "$SCRIPTS/nextcloud-startup-script.sh" ]
then
msg_box "Running a server, it's important that certain things are correct. You want the timezone to match your location,
you want the keyboard setup to match your keyboard, and you want mirrors to be as close to you as possible.

In the following menu you will be asked to setup the most basic stuff of your server. 

The script is smart, and have already pre-selected the values that you'd want to change based on the current settings."
fi
```

szaimen:
> and you want mirrors to be as close to you as possible.

BTW: if we say this, shouldn't we preselect the change-mirror option, too?

menu/startup_configuration.sh
```diff
@@ -64,7 +64,7 @@ case "$choice" in
                 run_script ADDONS locales
             fi
             input_box "Please try out all buttons (e.g: @ # \$ : y n) to find out if the keyboard settings were correctly applied.\nIf the keyboard is still wrong, you will be offered to reboot the server in the next step.\n\nPlease continue by hitting [ENTER]" >/dev/null
-            if ! yesno_box_no "Did the keyboard work as expected??\n\nIf you choose 'No' the server will be rebooted. After the reboot, please login as usual and run this script again."
+            if ! yesno_box_yes "Did the keyboard work as expected??\n\nIf you choose 'No' the server will be rebooted. After the reboot, please login as usual and run this script again."
```
enoch85, line 60: Did you forget about the `!` maybe?

szaimen: no, this is correct

enoch85:
> BTW: if we say this, shouldn't we preselect the change-mirror option, too?

No, as we use archive.ubuntu.com as default, that's already the best option in general if you ask me.

>  It could look like this:

Sure, let's do that.

menu/startup_configuration.sh
szaimen: `yesno_box_yes` means, that it is by default on the `yes` button.
and the `!` means, that it will trigger the reboot, if you select `no`.

szaimen:
> No, as we use archive.ubuntu.com as default, that's already the best option in general if you ask me.

okay, but shouldn't we then change the text? I think otherwise many people will select the change_mirror option, no?

enoch85:
Seems like you removed the text? What happened to my changes? :thinking:

szaimen:
> Seems like you removed the text? What happened to my changes? 🤔

I've moved it to the startup script.
But currently move it back to the menu

szaimen:
> > Seems like you removed the text? What happened to my changes? 🤔
> 
> I've moved it to the startup script.
> But currently move it back to the menu

done

szaimen:
> > > > We could then also think about putting this as an additional option into main_menu.
> > > 
> > > 
> > 
> > 
> > Sure, if this gets merged.
> 
> I've put it into main_menu. Question is, if we want another order? e.g. startup_configuration is above server_configuration?

Do we want to change the order?

PR #1424, opened 2020-09-07
remove TLS_INSTALL from the lib

szaimen:
as the title says

### DIFF
--- a/lets-encrypt/activate-tls.sh
+++ b/lets-encrypt/activate-tls.sh
@@ -63,19 +63,17 @@ fi
 
 while :
 do
-print_text_in_color "$ICyan" "Updating lib.sh..."
-# shellcheck disable=2034,2059
-true
-# shellcheck source=lib.sh
-TLS_INSTALL=1 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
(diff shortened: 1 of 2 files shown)


enoch85:
Hmm, I don't remember why I added it from the beginning...

szaimen:
> Hmm, I don't remember why I added it from the beginning...

Do you mean why you added TLS_INSTALL to the lib?

I would say most likely because of this line:
https://github.com/nextcloud/vm/blob/4a6402ebfd7c5e642091643094988db216c40fa2/lib.sh#L128
(The TLSDOMAIN variable is used only there in the lib)

enoch85:
> (The TLSDOMAIN variable is used only there in the lib)

OK, so why isn't that in this PR? :)

And no, that's just a product of the real reason. I think it was due to that I removed or replaced some `WHATEVER=1 . curl....` I wanted to fetch the latest lib, but I don't remember why.

enoch85:
OK, just found it: https://github.com/nextcloud/vm/pull/1025/files

This PR won't affect that as I probably just added the TLS_INSTALL due to the same reason as for SUBDOMAIN in other scripts - it should just be called when needed.

I'm fine with this. :+1:

PR #1425, opened 2020-09-08
move WHATEVER=1 to functions

szaimen:
It does what the title says.

Disclaimer: I didn't test the complete PR but I am pretty confident that it will work reliable this time since we already removed all problematic parts from the library.


szaimen:
I think this is ready for review.

lib.sh
```diff
@@ -1055,11 +1072,9 @@ any_key() {
 }
 
 lowest_compatible_nc() {
-if [ -z "$NC_UPDATE" ]
+if [ -z "$CURRENTVERSION" ]
```
enoch85, line 1075: We should check for `NCVERSION`

lib.sh
```diff
@@ -1095,11 +1110,9 @@ fi
 
 # Check new version
 # shellcheck source=lib.sh
-if [ -z "$NC_UPDATE" ]
+if [ -z "$CURRENTVERSION" ]
```
enoch85, line 1113: We should check for `NCVERSION`

lib.sh
```diff
@@ -1239,11 +1252,9 @@ printf "%b%s%b\n" "$1" "$2" "$Color_Off"
 # 2 = repository
 # Nextcloud version
 git_apply_patch() {
-if [ -z "$NC_UPDATE" ]
+if [ -z "$CURRENTVERSION" ]
```
enoch85, line 1255: We should check for `NCVERSION`

lib.sh
szaimen: done

lib.sh
szaimen: done

lib.sh
szaimen: done

PR #1427, opened 2020-09-09
create a STARTUP_SWITCH

szaimen:
as the title says

@enoch85 Do we have other defaults that would make sense on any instance?
I tried to set all options to ON during the startup script, if they make sense on any instance (at least if I think they make sense on any instance).

### DIFF
--- a/menu/additional_apps.sh
+++ b/menu/additional_apps.sh
@@ -38,7 +38,7 @@ choice=$(whiptail --title "$TITLE" --checklist "Which apps do you want to instal
 "PreviewGenerator" "(Pre-generate previews)" "$STARTUP_SWITCH" \
 "LDAP" "(Windows Active directory)" OFF \
 "Talk" "(Nextcloud Video calls and chat)" OFF \
-"Webmin" "(Server GUI)" OFF \
+"Webmin" "(Server GUI)" "$STARTUP_SWITCH" \
 "SMB-mount" "(Connect to SMB-shares from your local network)" OFF 3>&1 1>&2 2>&3)
 
 case "$choice" in
--- a/menu/additional_apps.sh
+++ b/menu/additional_apps.sh
@@ -17,17 +17,25 @@ debug_mode
 # Must be root
 root_check
 
+# Set the startup switch
+if [ -f "$SCRIPTS/nextcloud-startup-script.sh" ]
+then
+    STARTUP_SWITCH="ON"
+else
+    STARTUP_SWITCH="OFF"
+fi
+
 # Install Apps
 choice=$(whiptail --title "$TITLE" --checklist "Which apps do you want to install?\n\nAutomatically configure and install selected apps\n$CHECKLIST_GUIDE\n$MENU_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Documentserver" "(OnlyOffice or Collabora - Docker or Integrated)" OFF \
 "Bitwarden" "(External password manager)" OFF \
-"Fail2ban  " "(Extra Bruteforce protection)" OFF \
+"Fail2ban  " "(Extra Bruteforce protection)" "$STARTUP_SWITCH" \
 "Fail2ban Statuscheck" "(Check status of banned IPs in iptables and Fail2ban)" OFF \
 "Adminer" "(PostgreSQL GUI)" OFF \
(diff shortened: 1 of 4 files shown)


menu/additional_apps.sh
```diff
@@ -17,17 +17,25 @@ debug_mode
 # Must be root
 root_check
 
+# Set the startup switch
+if [ -f "$SCRIPTS/nextcloud-startup-script.sh" ]
+then
+    STARTUP_SWITCH="ON"
+else
+    STARTUP_SWITCH="OFF"
+fi
```
szaimen, line 26: I mean we could theoratically put that into the library...

enoch85: No, since we can't rely on that it will get fetched each time during the startup script. This implementation is better.

menu/server_configuration.sh
```diff
@@ -25,14 +25,22 @@ else
...
+    STARTUP_SWITCH="ON"
+else
+    STARTUP_SWITCH="OFF"
+fi
+
 # Server configurations
 choice=$(whiptail --title "$TITLE" --checklist "Choose what you want to configure\n$CHECKLIST_GUIDE\n$MENU_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
-"Security" "(Add extra security based on this http://goo.gl/gEJHi7)" OFF \
+"Security" "(Add extra security based on this http://goo.gl/gEJHi7)" "$STARTUP_SWITCH" \
```
enoch85, line 39: This should be OFF

menu/server_configuration.sh
```diff
@@ -25,14 +25,22 @@ else
...
 "DDclient Configuration" "(Use ddclient for automatic DDNS updates)" OFF \
 "Activate TLS" "(Enable HTTPS with Let's Encrypt)" "$ACTIVATE_TLS_SWITCH" \
-"Automatic updates" "(Automatically update your server every week on Sundays)" OFF \
-"Disk Check" "(Check for S.M.A.R.T errors on your disks every week on Mondays)" OFF 3>&1 1>&2 2>&3)
+"Automatic updates" "(Automatically update your server every week on Sundays)" "$STARTUP_SWITCH" \
+"Disk Check" "(Check for S.M.A.R.T errors on your disks every week on Mondays)" "$STARTUP_SWITCH" 3>&1 1>&2 2>&3)
```
enoch85, line 43: Both these should be OFF.

Automatic updates are not stable
Disk check doesn't work on VM disks

menu/additional_apps.sh
```diff
@@ -17,17 +17,25 @@ debug_mode
...
 "Fail2ban Statuscheck" "(Check status of banned IPs in iptables and Fail2ban)" OFF \
 "Adminer" "(PostgreSQL GUI)" OFF \
 "Netdata" "(Real-time server monitoring in Web GUI)" OFF \
 "BPYTOP" "(Real-time server monitoring in CLI)" OFF \
 "FullTextSearch" "(Elasticsearch for Nextcloud [2GB RAM])" OFF \
-"PreviewGenerator" "(Pre-generate previews)" OFF \
+"PreviewGenerator" "(Pre-generate previews)" "$STARTUP_SWITCH" \
 "LDAP" "(Windows Active directory)" OFF \
 "Talk" "(Nextcloud Video calls and chat)" OFF \
 "Webmin" "(Server GUI)" OFF \
```
enoch85, line 41: ON

menu/server_configuration.sh
szaimen: done

menu/server_configuration.sh
szaimen: done

enoch85:
You could add the startup script switch to all menus. Even if it doesn't do anything, we might use it later.

After that I think this is good to merge.

szaimen:
@enoch85 did you remove your comment on webmin? I've seen that you wanted it to be On during the startup script, but I can no longer find the comment, which is a bit strange.

szaimen:
> You could add the startup script switch to all menus. Even if it doesn't do anything, we might use it later.
> 
> After that I think this is good to merge.

okay

szaimen:
> > You could add the startup script switch to all menus. Even if it doesn't do anything, we might use it later.
> > After that I think this is good to merge.
> 
> okay

done

PR #1429, opened 2020-09-09
add a SUBTITLE to sensible places and other small fixes

szaimen:


addons/locate_mirror.sh
```diff
@@ -50,5 +51,6 @@ else
     then
         sudo mv sources.list /etc/apt/
     fi
+    msg_box "The apt-mirror was successfully changed."
```
enoch85, line 54: Move this within the if argument please.

enoch85:
Side note, 

Not when we made such big changes in the scripts. Please test 19.0.2 and choose to install everything. Just to make sure we didn't break something. Please don't forget to run without internet  as well to test the static_ip script properly as well.

I would if I had time, but I don't right now.

enoch85:
> I would if I had time, but I don't right now.

I will take care of it. :+1:

enoch85:
All is good! :D

enoch85:
@szaimen Please fix travis.

Other than that I think it looks OK to merge.

szaimen:
> @szaimen Please fix travis.

done

PR #1432, opened 2020-09-09
always run the locales script directly after changing the keyboard configuration

### LINKED ISSUE #1428: some questions after #1423
some questions after #1423

1. 
> > > > We could then also think about putting this as an additional option into main_menu.
> > > 
> > > 
> > 
> > 
> > Sure, if this gets merged.
> 
> 
> I've put it into main_menu. Question is, if we want another order? e.g. startup_configuration is above server_configuration?

Do we want to change the order?

2. We could remove that check here: https://github.com/nextcloud/vm/pull/1423/files#diff-8212cd239f6504419be211baf0a1a292R64 and always run the locales script here.
Additionally I would then remove this line: https://github.com/nextcloud/vm/pull/1423/files#diff-2912af18ccc7a96fbdf15892788006f2R313
But we would risk, that the locales script doesn't run at all during the startup-script, which should be no problem, since it already runs once during the install-script and has the correct locales already setup, if not changing the keyboard-layout.
What do you think?

szaimen:
@enoch85 Do we want to do this?

fix #1428

### DIFF
--- a/menu/main_menu.sh
+++ b/menu/main_menu.sh
@@ -21,8 +21,8 @@ root_check
 choice=$(whiptail --title "$TITLE" --menu "Choose what you want to do.\n$MENU_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Additional Apps" "(Choose which apps to install)" \
 "Nextcloud Configuration" "(Choose between available Nextcloud configurations)" \
-"Server Configuration" "(Choose between available server configurations)" \
 "Startup Configuration" "(Choose between available startup configurations)" \
+"Server Configuration" "(Choose between available server configurations)" \
 "Update Nextcloud" "(Update Nextcloud to the latest release)" 3>&1 1>&2 2>&3)
 
 case "$choice" in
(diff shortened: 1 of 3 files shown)


enoch85:
Sure, makes sense.

szaimen:
> Sure, makes sense.

And do we also want to move startup_configuration in main_menu above server_configuration?

enoch85:
Scratch that, I now see that this is only run if chosen.

Not a good idea. I plan to extend the locales script further to automatically set ALL locales, so this should always be run.

enoch85:
> And do we also want to move startup_configuration in main_menu above server_configuration?

Yes, that would be nice.

szaimen:
> Scratch that, I now see that this is only run if chosen.
> 
> Not a good idea. I plan to extend the locales script further to automatically set ALL locales, so this should always be run.

But it runs already in the install-script. So is this really needed to run definitely during the startup-script?

szaimen:
> > And do we also want to move startup_configuration in main_menu above server_configuration?
> 
> Yes, that would be nice.

done 👍

enoch85:
> But it runs already in the install-script

Well, the user might not want that we install it with, which is US. 

My plan is to extend the locales.sh to ALL keymaps, basically just continue with the current `elif` statement. If US is used for Keyboard, nothing is done with locales anyway, but if it's not default (US), then we change to what's set.

So, yes, I think it's needed to be run, always. :)

szaimen:
> Well, the user might not want that we install it with, which is US.
> 
> My plan is to extend the locales.sh to ALL keymaps, basically just continue with the current `elif` statement. If US is used for Keyboard, nothing is done with locales anyway, but if it's not default (US), then we change to what's set.
> 
> So, yes, I think it's needed to be run, always. :)

Yes, but if you don't choose to change the keyboard layout (and leave it on us) you most likely don't need other locales than the default us-locales, no?

enoch85:
Hmm, OK. Let's do it like this.

PR #1434, opened 2020-09-09
readd some subtitles that were mistakenly removed

szaimen:
Does the word readd exist? or is re-add correct? I am not sure...


enoch85:
I would say add back, but I'm not sure either. 🤷‍♂️

PR #1435, opened 2020-09-09
reimagine the structure of app-scripts, exemplarily on adminer.sh

szaimen:
POC how we could also design the app-scripts and other scripts. What do ou think?
It is much shorter and more shared code through the lib.
It works already in my testing.


enoch85:
Let's test this only for Adminer for a while.

PR #1438, opened 2020-09-10
some things to discuss after chain libraries

szaimen:
some things to discuss after chain libraries #1414
5 points:
- [x] 1. removing the whitespace around fetch_lib parts in scripts
- [x] 2. move border below complete header
- [x] 3. edit the source commands
- [x] 4. I think a msg_box/whiptail is a bit overkill for fetch_lib. And does it work? (TITLE, SUBTITLE, WT_HEIGHT, WT_WIDTH should be empty...
- [ ] 5. ~~do we want to cache the file or try to download a new one each time?~~
- [ ] 6. ~~move debug_mode and is_root check to fetch_lib?~~

### DIFF
--- a/addons/locales.sh
+++ b/addons/locales.sh
@@ -6,7 +6,7 @@
 true
 SCRIPT_NAME="Locales"
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
 
 ### TODO Remove this after some releases
 # Download fetch_lib.sh to be able to use it
--- a/addons/locales.sh
+++ b/addons/locales.sh
@@ -15,8 +15,16 @@ then
     download_script STATIC fetch_lib
 fi
 
+# Check for errors + debug code and abort if something isn't right
+# 1 = ON
+# 0 = OFF
+DEBUG=0
+debug_mode
+
 # Must be root
 root_check
+
+##################################################################
 
 # Set locales
 print_text_in_color "$ICyan" "Setting locales..."
--- a/addons/locales.sh
+++ b/addons/locales.sh
@@ -6,7 +6,7 @@
 true
 SCRIPT_NAME="Locales"
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
 ### TODO Remove this after some releases
 # Download fetch_lib.sh to be able to use it
--- a/addons/locales.sh
+++ b/addons/locales.sh
@@ -5,11 +5,8 @@
 # shellcheck disable=2034,2059,1091
 true
 SCRIPT_NAME="Locales"
-
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
-
-###################################
+source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
 
 ### TODO Remove this after some releases
 # Download fetch_lib.sh to be able to use it
--- a/menu/main_menu.sh
+++ b/menu/main_menu.sh
@@ -6,7 +6,7 @@
 true
 SCRIPT_NAME="Main Menu"
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
 ### TODO Remove this after some releases
 # Download fetch_lib.sh to be able to use it
--- a/menu/main_menu.sh
+++ b/menu/main_menu.sh
@@ -24,6 +24,8 @@ debug_mode
 # Must be root
 root_check
 
+##################################################################
+
 # Main menu
 choice=$(whiptail --title "$TITLE" --menu "Choose what you want to do.\n$MENU_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Additional Apps" "(Choose which apps to install)" \
--- a/menu/main_menu.sh
+++ b/menu/main_menu.sh
@@ -6,7 +6,7 @@
 true
 SCRIPT_NAME="Main Menu"
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
 
 ### TODO Remove this after some releases
 # Download fetch_lib.sh to be able to use it
--- a/menu/main_menu.sh
+++ b/menu/main_menu.sh
@@ -5,11 +5,8 @@
 # shellcheck disable=2034,2059,1091
 true
 SCRIPT_NAME="Main Menu"
-
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
-
-###################################
+source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
 
 ### TODO Remove this after some releases
 # Download fetch_lib.sh to be able to use it
--- a/menu/menu.sh
+++ b/menu/menu.sh
@@ -6,7 +6,7 @@
 true
 SCRIPT_NAME="Main Menu"
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
--- a/menu/menu.sh
+++ b/menu/menu.sh
@@ -5,11 +5,8 @@
 # shellcheck disable=2034,2059,1091
 true
 SCRIPT_NAME="Main Menu"
-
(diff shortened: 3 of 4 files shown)


addons/locales.sh
```diff
@@ -5,11 +5,8 @@
 # shellcheck disable=2034,2059,1091
 true
 SCRIPT_NAME="Locales"
-
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
-
```
szaimen, line 11: 1. I think the whitespace around here isn't needed

addons/locales.sh
```diff
@@ -5,11 +5,8 @@
...
-###################################
```
szaimen, line 12: 2. I don't like the border here (but I understand what you wanted to do, I only think it is not necessary)

addons/locales.sh
```diff
@@ -5,11 +5,8 @@
...
+source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
```
szaimen, line 9: 3.  I want to see if `source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null` works, too (travis)

menu/menu.sh
```diff
@@ -5,11 +5,8 @@
 # shellcheck disable=2034,2059,1091
 true
 SCRIPT_NAME="Main Menu"
-
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
-
-###################################
+source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
```
szaimen, line 9: here the same like above

menu/main_menu.sh
```diff
@@ -5,11 +5,8 @@
 # shellcheck disable=2034,2059,1091
 true
 SCRIPT_NAME="Main Menu"
-
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
-
-###################################
+source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
```
szaimen, line 9: here the same like above

static/fetch_lib.sh
```diff
@@ -3,30 +3,23 @@
 true
 # see https://github.com/koalaman/shellcheck/wiki/Directive
 
-# Functions
-msg_box() {
-    [ -n "$2" ] && local SUBTITLE=" - $2"
-    whiptail --title "$TITLE$SUBTITLE" --msgbox "$1" "$WT_HEIGHT" "$WT_WIDTH"
-}
-
-
```
szaimen, line 12: 4. I think a msg_box/whiptail is a bit overkill for fetch_lib. And does it work? (TITLE, SUBTITLE, WT_HEIGHT, WT_WIDTH should be empty...

static/fetch_lib.sh
```diff
@@ -3,30 +3,23 @@
...
 IRed='\e[0;91m'         # Red
 IGreen='\e[0;92m'       # Green
 ICyan='\e[0;96m'        # Cyan
 Color_Off='\e[0m'       # Text Reset
 print_text_in_color() {
-        printf "%b%s%b\n" "$1" "$2" "$Color_Off"
+    printf "%b%s%b\n" "$1" "$2" "$Color_Off"
```
szaimen, line 11: just a bit reformatting

static/fetch_lib.sh
```diff
@@ -3,30 +3,23 @@
...
 }
 
 mkdir -p /var/scripts
 if ! [ -f /var/scripts/lib.sh ]
 then
     if ! curl -sfL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh -o /var/scripts/lib.sh
     then
-        msg_box "You don't seem to have an internet connection and the local lib isn't available. Hence you cannot run this script."
+        print_text_in_color "$IRed" "You don't seem to have an internet connection and the local lib isn't available. Hence you cannot run this script."
```
szaimen, line 19: this also counts to point 4

static/fetch_lib.sh
```diff
@@ -3,30 +3,23 @@
...
         exit 1
     fi
-else
+elif test "$(find /var/scripts/lib.sh -mmin +30)"
```
szaimen, line 22: 5. we want to cache the file or try to download a new one each time?

addons/locales.sh
szaimen: It seems to work, too! 🚀

szaimen: We need to change a´something else, too:
it must be: `source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null`
else, we will not get any print_text_in_color from fetch_lib.sh (with the disadvantage, that we will see an error message, if the file isn't there. But this is only cosmetic since we want to switch to it later on everywhere.)

addons/locales.sh
enoch85: It marks where the header ends. Everything above that line is static. Easier to find.

addons/locales.sh
enoch85: Yeah whatever. :)

addons/locales.sh
enoch85: Whut, so some functions doesn't work if you put it in /dev/null? Then don't do that at all. Harder to debug as well.

menu/main_menu.sh
enoch85: remove /dev/null altogether

static/fetch_lib.sh
enoch85: Okay, let's do it like that.

static/fetch_lib.sh
enoch85: We managed fine without for the past 6 years, so what are the benefits?

addons/locales.sh
szaimen: No, only things that get printed out during the loading of the file. Functions work just fine and aren't affected.

szaimen: If we just load the lib, this doesn't have any disadvantage

menu/main_menu.sh
szaimen: I've remeoved it for fetch_lib where it is important

static/fetch_lib.sh
szaimen: Okay

menu/menu.sh
```diff
@@ -6,7 +6,7 @@
 true
 SCRIPT_NAME="Main Menu"
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh &>/dev/null || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh) &>/dev/null
```
enoch85: Remove in both places please, the curl as well.

szaimen: As I wrote this is only an issue with fetch_lib.sh and not with lib.sh

szaimen: It was already fixed for fetch_libs.sh

enoch85: It's harder to debug without an output. Nothing is shown anyway if not run with `-x`.

static/fetch_lib.sh
```diff
@@ -3,33 +3,27 @@
...
+then
     print_text_in_color "$ICyan" "Updating lib..."
     curl -sfL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh -o /var/scripts/lib.sh
 fi
 
 # shellcheck source=lib.sh
-source /var/scripts/lib.sh &>/dev/null
+source /var/scripts/lib.sh
```
szaimen, line 29: this is really not necessary!
But if you do it here, I would suggest that you also remove it from anywhere else

addons/locales.sh
szaimen: Hm, I think they are still very easy to find also without, because they sit at the top.
And what about those in other scripts: aren't they headers, too?
```
# Check for errors + debug code and abort if something isn't right
# 1 = ON
# 0 = OFF
DEBUG=0
debug_mode

# Must be root
root_check
```

static/fetch_lib.sh
szaimen: one advantage would be, that we don't print out "Updating lib..." every time we run any script.
(Could be a bit annoying when you run many scripts)
Also, we maybe speed up the process a bit, if we don't download a new lib each time we run a script. (could be an advantage on a slow network)

menu/menu.sh
szaimen: @enoch85 Yeas, this is why I added the /dev/null in the first place, because I think the information that is printed out by the lib, if running in -x mode is not worth the flooting of the console (I find it much harder to debug if the console gets floated)

addons/locales.sh
szaimen: So IMO we move the border below `root_check` or just remove the border completely.

static/fetch_lib.sh
enoch85: That's the plan :)

menu/menu.sh
szaimen: @enoch85 What is your opinion on my statement above?

addons/locales.sh
enoch85: Sure, let's move if to after `root_check`.

szaimen:
@enoch85 What is your opinion on that statement?
https://github.com/nextcloud/vm/pull/1438#discussion_r486258081

addons/locales.sh
szaimen: done

addons/locales.sh
```diff
@@ -18,9 +15,17 @@ then
...
 fi
 
+# Check for errors + debug code and abort if something isn't right
+# 1 = ON
+# 0 = OFF
+DEBUG=0
+debug_mode
+
 # Must be root
 root_check
```
szaimen, line 25: 6. We could move that header to the fetch_lib file, if we want to shorten the header
And btw: is debug_mode really needed, if you can just run `bash -x /path/to/script` ?

static/fetch_lib.sh
enoch85: Printing "Updating lib..." is a good thing. On slow networks the cursor just stands there an blinks otherwise and the user don't know what's going on.

I feels like untested territory for me, and I know what worked in the past, and I don't know what side effects caching might have on all the scripts in this repo. Sometimes the lib file needs to be 1 second old for stuff to work. I can't tell you exactly where right now, but I know I've coded it such in some places.

menu/menu.sh
enoch85: /devnull is now removed, so this is solved.

addons/locales.sh
enoch85: Honestly I never used debug mode, but I don't know if some of the several thousands of users actually use it. It doesn't hurt to leave it as is imo.

static/fetch_lib.sh
enoch85: Generally, I think caching just cause more issues than damage. As late as yesterday I had a customer that said stuff didn't work one day, and the next day all was good. Caching, no thanks.

szaimen: Okay, we might leave it then as it is.
(I'll remove that change)

szaimen: Done

addons/locales.sh
szaimen: Okay

szaimen:
From my side this is ready

PR #1439, opened 2020-09-10
revert the change to the sed command in webmin

szaimen:
@enoch85 Unfortunately you cannot simply substitute the string since it doesn't exist in the file after installing webmin

### DIFF
--- a/apps/webmin.sh
+++ b/apps/webmin.sh
@@ -67,6 +67,7 @@ print_text_in_color "$ICyan" "Configuring Webmin..."
 check_command systemctl stop webmin
 ## TODO: This will cause Webmin to redirect to the hostname 'nextcloud' and not the IP address if on LAN
 check_command sed -i '/^ssl=.*/a ssl_redirect=1' /etc/webmin/miniserv.conf
+check_command sed -i "/^port=.*/a hostname=$ADDRESS" /etc/webmin/miniserv.conf
 check_command systemctl start webmin
 
 msg_box "Webmin is now installed and can be accessed from this address:
--- a/apps/webmin.sh
+++ b/apps/webmin.sh
@@ -66,7 +66,7 @@ print_text_in_color "$ICyan" "Configuring Webmin..."
 # redirect access on http to https
 check_command systemctl stop webmin
 ## TODO: This will cause Webmin to redirect to the hostname 'nextcloud' and not the IP address if on LAN
-check_command sed -i "s|ssl_redirect=.*|ssl_redirect=1|g" /etc/webmin/miniserv.conf
+check_command sed -i '/^ssl=.*/a ssl_redirect=1' /etc/webmin/miniserv.conf
 check_command systemctl start webmin
 
 msg_box "Webmin is now installed and can be accessed from this address:
--- a/apps/webmin.sh
+++ b/apps/webmin.sh
@@ -67,7 +67,7 @@ print_text_in_color "$ICyan" "Configuring Webmin..."
 check_command systemctl stop webmin
 ## TODO: This will cause Webmin to redirect to the hostname 'nextcloud' and not the IP address if on LAN
 check_command sed -i '/^ssl=.*/a ssl_redirect=1' /etc/webmin/miniserv.conf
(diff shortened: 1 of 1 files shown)


apps/webmin.sh
```diff
@@ -66,7 +66,7 @@ print_text_in_color "$ICyan" "Configuring Webmin..."
 # redirect access on http to https
 check_command systemctl stop webmin
 ## TODO: This will cause Webmin to redirect to the hostname 'nextcloud' and not the IP address if on LAN
-check_command sed -i "s|ssl_redirect=.*|ssl_redirect=1|g" /etc/webmin/miniserv.conf
```
szaimen, line 69: BTW: what does the g at the end of the sed command do?

enoch85: https://unix.stackexchange.com/questions/484659/what-does-this-command-do-sed-s-g

I've always used it and it works, never digged down in the details. :)

szaimen: I've just asking because I've seen this command also without the g...

enoch85:
Would it be possible to force an IP address? Most people only use Webmin locally and the hostname won't resolve.

szaimen:
> Would it be possible to force an IP address? Most people only use Webmin locally and the hostname won't resolve.

I am not sure. Does this also redirect the access on https://ip-address:10000 to the hostname?

Because in my testing it didn't. And imo it is fine this way (That it redirects access on http://ip-address:10000 to the hostname) because before it just sayd no access is possible here

szaimen:
Maybe the hostname in the file could be a starting point: if we add/change this in the file:
`hostname="$ADDRESS"`

enoch85:
> Does this also redirect the access on

Yes it does. :(

> hostname="$ADDRESS"

Yes please!

szaimen:
@enoch85 could you please try if that works for you? For me the rewrite to https://ip-address:10000 has worked before adding the hostname somehow so I am not sure how to reproduce that issue.

enoch85:
It doesn't rewrite to HTTPS now, but it lets me access Webmin directly, so it's better anyway.

I didn't test it thoroughly though.

szaimen:
> It doesn't rewrite to HTTPS now, but it lets me access Webmin directly, so it's better anyway.
> 
> I didn't test it thoroughly though.

Does that mean that you cannot access webmin on http://ip-address:10000 again? If that is the case I guess we can remove all the sed commands since we haven't changed anything in comparison to how it was in the beginning.

enoch85:
I could access Webmin on http even after both seds were done.

But, not sure about the whole thing as it was just a quick test.

enoch85:
OK, so I did a proper test now and the end result is still the same. http://192.168.2.87:10000 --->

Screenshot from 2020-09-12 18:51:43

enoch85:
Actually, I think it's better without any change at all.

At least the users is instructed to go to https, with a simple click.

szaimen:
The problem is, that after activating tls it doesn't proposes you to access webmin via https://ip-address:10000 but shows https://your-nextcloud-domain:10000 which will logically not work if you don't port-foward that port...

And the redirect to the internal hostname works for me.
Btw: accessing webmin on https://ipaddress:10000 doesn't redirect to the hostbame?

enoch85:
> port-foward that port...

Yeah, and we don't want that. 

> And the redirect to the internal hostname works for me.

Good for you, but not here. :D

> Btw: accessing webmin on https://ipaddress:10000 doesn't redirect to the hostbame?

Yes, and that's the problem. Maybe you have internal DNS, but I don't and I guess most users don't either. :/

enoch85:
I've sent you a video via email.

szaimen:
@enoch85 I think I fixed it (the redirect to https works for me now). Could you please try again?

enoch85:
Tested and works!

szaimen:
@enoch85 we forgot to remove this comment:
https://github.com/nextcloud/vm/blob/0f2d96d752545661feef8233ba07f4db27eddc6d/apps/webmin.sh#L68

enoch85:
OK, will fix :+1:

PR #1442, opened 2020-09-11
fix issue with non-root-users

### LINKED ISSUE #1441: scripts must be run as root if lib.sh doesn't exist already
To be able to run any script using fetch_lib, the user must run as root - otherwise it won't be able to get the lib.sh to /var/scripts and the the scipt bugs out with that lib doesn't exists and that internet doesn't work.

This needs to be fixed. We should probably check if sudo and print the correct error message. Right now the error is toally wrong.

We should probably make scripts writable by everyone as well, or maybe not, I don't know right now.

cc @szaimen

szaimen:
shoud fix #1441
I didn't test this but I am pretty confident that it works ;)

### DIFF
--- a/addons/locales.sh
+++ b/addons/locales.sh
@@ -11,7 +11,6 @@ source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercont
 ### TODO Remove this after some releases
 # Download fetch_lib.sh to be able to use it
 download_script STATIC fetch_lib
-chmod 777 "$SCRIPTS"/fetch_lib.sh
 
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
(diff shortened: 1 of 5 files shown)


enoch85:
Please remove the 777 permissions, adding the root check is enough. :)

enoch85:
```
ncadmin@cloud:~$ bash -x /var/scripts/menu.sh 
+ true
+ SCRIPT_NAME='Main Menu'
+ source /var/scripts/fetch_lib.sh
++ true
++ IRed='\e[0;91m'
++ IGreen='\e[0;92m'
++ ICyan='\e[0;96m'
++ Color_Off='\e[0m'
++ mkdir -p /var/scripts
++ '[' -f /var/scripts/lib.sh ']'
++ curl -sfL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh -o /var/scripts/lib.sh
++ print_text_in_color '\e[0;91m' 'You don'\''t seem to have an internet connection and the local lib isn'\''t available. Hence you cannot run this script.'
++ printf '%b%s%b\n' '\e[0;91m' 'You don'\''t seem to have an internet connection and the local lib isn'\''t available. Hence you cannot run this script.' '\e[0m'
You don't seem to have an internet connection and the local lib isn't available. Hence you cannot run this script.
++ exit 1
```

PR #1445, opened 2020-09-12
updatenotification ist stable

szaimen:


enoch85:
We should leave it im the uofstes script so that everyone can get it before the new release, then make a new PR to remove it once the new release is out.

szaimen:
Ah okay, yes you are right!

szaimen:
> We should leave it in the updates script so that everyone can get it before the new release, then make a new PR to remove it once the new release is out.

done 👍

enoch85:
:drum:

PR #1449, opened 2020-09-13
create input_box_flow

szaimen:
standardizes the inputbox-flow with asking if it is correct, etc.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -246,7 +246,7 @@ ask_yes_or_no() {
 
 msg_box() {
     [ -n "$2" ] && local SUBTITLE=" - $2"
-    whiptail --title "$TITLE$SUBTITLE" --msgbox "$1" "$WT_HEIGHT" "$WT_WIDTH"
+    whiptail --title "$TITLE$SUBTITLE" --msgbox "$1" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3
 }
 
 # TODO: delete in a few releases (e.g. NC20) since not needed anymore
@@ -262,7 +262,7 @@ yesno_box() {
 
 yesno_box_yes() {
     [ -n "$2" ] && local SUBTITLE=" - $2"
-    if (whiptail --title "$TITLE$SUBTITLE" --yesno "$1" "$WT_HEIGHT" "$WT_WIDTH")
+    if (whiptail --title "$TITLE$SUBTITLE" --yesno "$1" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3)
     then
         return 0
     else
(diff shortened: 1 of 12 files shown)


lib.sh
```diff
@@ -246,7 +246,7 @@ ask_yes_or_no() {
 
 msg_box() {
     [ -n "$2" ] && local SUBTITLE=" - $2"
-    whiptail --title "$TITLE$SUBTITLE" --msgbox "$1" "$WT_HEIGHT" "$WT_WIDTH"
+    whiptail --title "$TITLE$SUBTITLE" --msgbox "$1" "$WT_HEIGHT" "$WT_WIDTH" 3>&1 1>&2 2>&3
```
enoch85, line 249: Why is the last part removed?

network/ddclient-configuration.sh
```diff
@@ -56,55 +56,13 @@ then
...
 
 # Enter your password
-while :
-do
-    PASSWORD=$(input_box "Please enter the password or api-key that you've got for DynDNS from your DDNS provider.\nIf you are not sure, please refer to the documentation of your DDNS provider.")
-    if ! yesno_box_yes "Is this correct? $PASSWORD"
-    then
-        msg_box "OK, please try again."
-    else
-        if [ -z "$PASSWORD" ]
```
enoch85, line 100: `input_box_flow` doesn't take this into account?

lib.sh
szaimen: you mean added? 
Without this the new function won't work, I cannot really explain why.
I can just say that it doesn't introduce unexpected side effects, so I added it to make the new function work

enoch85: Sorry, I'm obviously blind. :D

network/ddclient-configuration.sh
szaimen: because it is not really necessary. (we show the result box later on and there you should make yourself sure that all values are correct). If not it is the users mistake and not ours.

enoch85: I think we should handle that case anyway. 

Just added a fix.

enoch85:
This is great btw! :D

lib.sh
```diff
@@ -291,7 +291,10 @@ input_box_flow() {
     while :
     do
         RESULT=$(input_box "$1" "$2")
-        if ! yesno_box_yes "Is this correct? $RESULT" "$2"
+        if [ -z "$RESULT" ]
+        then
+            msg_box "Input is empty, please try again."
+        elif ! yesno_box_yes "Is this correct? $RESULT" "$2"
```
szaimen, line 297: I am against this change because we won't be able to do something like this:
https://github.com/nextcloud/vm/pull/1449/files#diff-cb545bead2637c389f51b1ed9215d91eL61

enoch85: I think we should use the default `input_box` in that case, or don't ask the user to leave it empty.

enoch85: Regarding the bitwarden_mailconfig, maybe it would be better to grep that value and use it like we do in static IP?

https://github.com/nextcloud/vm/blob/master/network/static_ip.sh#L129

szaimen: I guess we have to rewrite the bitwarden-mailconfig script then in a follow-up.

szaimen: I opened #1453 for this

enoch85:
So ready to merge?

szaimen:
yes 👍

enoch85:
:partying_face:

PR #1450, opened 2020-09-13
fix adminer bad request and change port

enoch85:
@szaimen Does this work?

### DIFF
--- a/apps/adminer.sh
+++ b/apps/adminer.sh
@@ -55,9 +55,14 @@ curl_to_dir "https://raw.githubusercontent.com/Niyko/Hydra-Dark-Theme-for-Admine
 ln -s "$ADMINERDIR"/latest.php "$ADMINERDIR"/adminer.php
 
 cat << ADMINER_CREATE > "$ADMINER_CONF"
-Listen 8443
+ <VirtualHost *:80>
+     RewriteEngine On
+     RewriteRule ^(.*)$ https://%{HTTP_HOST}$1:9443 [R=301,L]
+ </VirtualHost>
 
-<VirtualHost *:8443>
+Listen 9443
+
+<VirtualHost *:9443>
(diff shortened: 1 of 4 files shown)


apps/adminer.sh
```diff
@@ -55,9 +55,14 @@ curl_to_dir "https://raw.githubusercontent.com/Niyko/Hydra-Dark-Theme-for-Admine
 ln -s "$ADMINERDIR"/latest.php "$ADMINERDIR"/adminer.php
 
 cat << ADMINER_CREATE > "$ADMINER_CONF"
-Listen 8443
+ <VirtualHost *:80>
+     RewriteEngine On
+     RewriteRule ^(.*)$ https://%{HTTP_HOST}$1:9443 [R=301,L]
+ </VirtualHost>
```
szaimen, line 61: This would only work, if I access the server on http://ipaddress
and not on http://ipaddress:9443
So it unfortunately doesn't solve the issue but I guess there is just no solution to this.
BTW: I would remove that part again because you might want to  access your nextcloud on the default ports and not adminer

apps/adminer.sh
```diff
@@ -99,7 +104,7 @@ The script will exit."
     exit 1
 else
 msg_box "Adminer was sucessfully installed and can be reached here:
-https://$ADDRESS:8443
+https://$ADDRESS:9443
```
szaimen, line 107: I would additionally add the port to NONO_PORTS

enoch85: Yes, just wanted to make sure it worked first.

8443 wasn't in nono ports so that's why I used it in the first place. ;)

apps/adminer.sh
enoch85: Hmm don't have time to fix that now. Tried some different stuff and would like to avoid using `.htaccess` Maybe we can figure it out later.

PR #1452, opened 2020-09-13
adminer - redirect access to https

### LINKED ISSUE #1430: Issue with accesing adminer after enabling TLS for nextcloud
So I've just found an issue:

The same happens, after I activate only tls for the nextcloud domain ->
I can no longer acces adminer on the local `ip/adminer.php` but get redirected to the nextcloud domain.
Any idea how to fix this?

szaimen:
This solution has worked for me
fix #1430


enoch85:
Aah, nice catch!

PR #1457, opened 2020-09-13
webmin - use the new reinstall-functions and smaller fixes

szaimen:


szaimen:
Should the `/etc/apt/sources.list.d/webmin.list` file also get removed upon reinstalling/uninstalling?
https://github.com/nextcloud/vm/pull/1457/files#diff-470c274da614a3e50289211998c445f1R52

apps/webmin.sh
```diff
@@ -5,6 +5,10 @@
 # shellcheck disable=2034,2059
 true
 SCRIPT_NAME="Webmin"
+SCRIPT_EXPLAINER="Webmin is a web-based interface for system administration for Unix.
+Using any modern web browser, you can setup user accounts, Apache, DNS, file sharing and much more.
+Webmin removes the need to manually edit Unix configuration files like /etc/passwd, and lets you manage a system from the console or remotely.
+See the following page with standard modules for a list of all the functions built into Webmin: https://webmin.com/standard.html"
```
szaimen, line 11: This description is from the official webmin website. Do you think that we can use that?
https://webmin.com/ (section What is Webmin?)

enoch85: LGTM

lib.sh
```diff
@@ -313,15 +313,19 @@ explainer_popup() {
 }
 
 reinstall_remove_menu() {
-    choice=$(whiptail --title "$TITLE" --menu "It seems like '$SCRIPT_NAME' is already installed.\nChoose what you want to do." "$WT_HEIGHT" "$WT_WIDTH" 4 \
+    choice=$(whiptail --title "$TITLE" --menu "It seems like $SCRIPT_NAME is already installed.\nChoose what you want to do." "$WT_HEIGHT" "$WT_WIDTH" 4 \
     "Reinstall $SCRIPT_NAME" "" \
     "Uninstall $SCRIPT_NAME" "" 3>&1 1>&2 2>&3)
+    if [ -z "$choice" ]
```
enoch85, line 319: We should add this everywhere. Isn't this exiting when you hit cancel? Currently that's not working.

apps/adminer.sh
```diff
@@ -32,7 +32,7 @@ then
     a2disconf adminer.conf
     rm -f $ADMINER_CONF
     rm -rf $ADMINERDIR
-    check_command apt-get purge adminer -y
+    check_command apt --purge autoremove -y adminer
```
enoch85, line 35: No, we use apt-get for purging.

Please revert. Thanks!

enoch85:
> Should the /etc/apt/sources.list.d/webmin.list file also get removed upon reinstalling/uninstalling?

Yes please :+1:

lib.sh
szaimen: done

apps/adminer.sh
szaimen: done

szaimen:
> > Should the /etc/apt/sources.list.d/webmin.list file also get removed upon reinstalling/uninstalling?
> 
> Yes please 👍

done

szaimen:
What about this line: how to remove this?
`curl -fsSL http://www.webmin.com/jcameron-key.asc | sudo apt-key add -`

enoch85:
> curl -fsSL http://www.webmin.com/jcameron-key.asc | sudo apt-key add -

You can list all keys with `apt-key list`,  with that you could probably `grep` for 'cameron' or similar and then do a `apt-key del`. This is from the top of my head, would need confirmation.

enoch85:
> curl -fsSL http://www.webmin.com/jcameron-key.asc | sudo apt-key add -

I will release 19.0.3 today, so let's figure that out in a separate PR.

PR #1460, opened 2020-09-15
use fetch_lib in Nextcloud startup script as well

enoch85:


nextcloud-startup-script.sh
```diff
@@ -135,8 +119,11 @@ Please also post this issue on: https://github.com/nextcloud/vm/issues"
     exit 1
 fi
 
+# shellcheck disable=2034,2059,1091
+true
```
szaimen, line 123: this isn't needed, afaik

enoch85: I think it's there for  reason, otherwise we wouldn't use it :)

This is a quick fix before release. I have major internet connectivity issues, and I think this is better.

PR #1461, opened 2020-09-15
startup-script - use input_box_flow for the password and username change

### LINKED ISSUE #1454: Do something about the passwords promt during first setup
Now when we have the fancy `input_box_flow` merged, maybe we could do something about the prompts for password during the first setup? 

1. ncadmin password
2. New Nextcloud user

That way we can also make sure that the password is correctly written as we could write the output directly on screen - thus minimizing "I lost my password becasue the keystrokes were incorrect in console / putty."

szaimen:
POC
fix #1454

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -358,7 +358,7 @@ do
     elif [ "$NEWUSER" = "$NCADMIN" ]
     then
         msg_box "This username is already in use. Please choose a different one."
-    elif [[ "$NEWUSER" =~ [^a-zA-Z0-9_.@-] ]]
+    elif [[ $NEWUSER == *[!abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_.-@] ]]
     then
         msg_box "Allowed characters are: a-z', 'A-Z', '0-9', and '_.@-'\nPlease try again."
     else
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -345,19 +345,49 @@ bash $SCRIPTS/additional_apps.sh
 clear
 # Change passwords
 # CLI USER
-print_text_in_color "$ICyan" "For better security, change the system user password for [$(getent group sudo | cut -d: -f4 | cut -d, -f1)]"
-any_key "Press any key to change password for system user..."
+msg_box "For better security, we will now change the password for the unix-user"
 while :
 do
-    sudo passwd "$(getent group sudo | cut -d: -f4 | cut -d, -f1)" && break
+    UNIX_PASSWORD=$(input_box_flow "Please type in the new password for the unix-user [$(getent group sudo | cut -d: -f4 | cut -d, -f1)]")
+    if [[ "$UNIX_PASSWORD" == *" "* ]]
+    then
+        msg_box "Please don't use spaces"
+    else
+        break
+    fi
 done
-echo
+check_command echo -e "$UNIX_PASSWORD\n$UNIX_PASSWORD" | (passwd --stdin "$(getent group sudo | cut -d: -f4 | cut -d, -f1)")
+unset UNIX_PASSWORD
 clear
 # NEXTCLOUD USER
 NCADMIN=$(occ_command user:list | awk '{print $3}')
-print_text_in_color "$ICyan" "The current admin user in Nextcloud GUI is [$NCADMIN]"
-print_text_in_color "$ICyan" "We will now replace this user with your own."
-any_key "Press any key to replace the current (local) admin user for Nextcloud..."
+msg_box "We will now change the username and password for the nextcloud-admin user"
+while :
+do
+    NEWUSER=$(input_box_flow "Please type in the name of the nextcloud-admin user. It must differ from [$NCADMIN].\nThe only allowed character are: 'a-z', 'A-Z', '0-9', and '_.@-'")
+    if [[ "$NEWUSER" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    elif [ "$NEWUSER" = "$NCADMIN" ]
+    then
+        msg_box "This username is already in use. Please choose a different one."
+    else
+        break
+    fi
+done
+while :
+do
+    OC_PASS=$(input_box_flow "The current admin user in Nextcloud GUI is [$NCADMIN]\nPlease type in the new password for the nextcloud-admin user")
+    if [[ "$NEWUSER" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    else
+        break
+    fi
+done
+export OC_PASS
+occ_command user:add "$NEWUSER" --password-from-env -g admin
+unset OC_PASS
 # Create new user
 while :
 do
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -373,9 +373,9 @@ do
     then
         msg_box "Please don't use spaces."
     fi
-# Create new user
-export OC_PASS
-    if su -s /bin/sh www-data -c "php /var/www/nextcloud/occ user:add $NEWUSER --password-from-env"
+    # Create new user
+    export OC_PASS
+    if su -s /bin/sh www-data -c "php $NCPATH/occ user:add $NEWUSER --password-from-env -g admin"
     then
         unset OC_PASS
         break
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -364,24 +364,18 @@ do
 done
 while :
 do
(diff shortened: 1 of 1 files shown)


nextcloud-startup-script.sh
```diff
@@ -345,19 +345,49 @@ bash $SCRIPTS/additional_apps.sh
...
+msg_box "We will now change the username and password for the nextcloud-admin user"
+while :
+do
+    NEWUSER=$(input_box_flow "Please type in the name of the nextcloud-admin user. It must differ from [$NCADMIN].\nThe only allowed character are: 'a-z', 'A-Z', '0-9', and '_.@-'")
+    if [[ "$NEWUSER" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    elif [ "$NEWUSER" = "$NCADMIN" ]
+    then
+        msg_box "This username is already in use. Please choose a different one."
```
szaimen, line 373: We most likely need an additional check here for allowed characters. but don't know how to do this

enoch85: fixed

nextcloud-startup-script.sh
```diff
@@ -330,28 +330,58 @@ bash $SCRIPTS/nextcloud_configuration.sh
...
-    read -r NEWUSER
-    sudo -u www-data $NCPATH/occ user:add "$NEWUSER" -g admin && break
+    NEWUSER=$(input_box_flow "Please type in the name of the Nextcloud admin user. It must differ from [$NCADMIN].\nThe only allowed character are: 'a-z', 'A-Z', '0-9', and '_.@-'")
+    if [[ "$NEWUSER" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    elif [ "$NEWUSER" = "$NCADMIN" ]
+    then
+        msg_box "This username is already in use. Please choose a different one."
+    elif [[ "$NEWUSER" =~ [^a-zA-Z0-9_.@-] ]]
```
szaimen, line 362: Why is the '^' needed?

enoch85: Don't know actually, but it works, so I'm satisfied. :)

Got inspiration from this: https://www.golinuxcloud.com/check-if-string-contains-numbers-shell-script/

enoch85:
@szaimen Please test this properly before merge.

I've done some testing, but it's always good with several eyes.

nextcloud-startup-script.sh
```diff
@@ -330,27 +330,58 @@ bash $SCRIPTS/nextcloud_configuration.sh
...
+while :
+do
+    OC_PASS=$(input_box_flow "Please type in the new password for the new nextcloud-admin user $NEWUSER")
+    if [[ "$OC_PASS" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    fi
+# Create new user
+export OC_PASS
+    if su -s /bin/sh www-data -c "php /var/www/nextcloud/occ user:add $NEWUSER --password-from-env"
```
szaimen, line 378: did you forget the `-g admin`?

nextcloud-startup-script.sh
szaimen: Okay cool, but it has to be outside of the `[]` otherwise it will be an allowed character.

nextcloud-startup-script.sh
```diff
@@ -330,27 +330,58 @@ bash $SCRIPTS/nextcloud_configuration.sh
...
+done
+while :
+do
+    OC_PASS=$(input_box_flow "Please type in the new password for the new nextcloud-admin user $NEWUSER")
+    if [[ "$OC_PASS" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    fi
+# Create new user
+export OC_PASS
```
szaimen, line 377: indentation?

nextcloud-startup-script.sh
szaimen: and can't we use $NCPATH here?

enoch85: You are correct on both. :+1:

nextcloud-startup-script.sh
enoch85: Try for yourself. :)

szaimen: I guess it has to be `elif [[ "$NEWUSER" =~ ^[a-zA-Z0-9_.@-]+$ ]]`

szaimen: better is most likely `elif [[ "$NEWUSER" =~ ^[a-zA-Z][a-zA-Z0-9_.@-]+$ ]]`
(we can only allow a letter at the beginning this way)

szaimen: sry I meant `elif ! [[ "$NEWUSER" =~ ^[a-zA-Z][a-zA-Z0-9_.@-]+$ ]]`

nextcloud-startup-script.sh
```diff
@@ -330,27 +330,58 @@ bash $SCRIPTS/nextcloud_configuration.sh
...
+msg_box "We will now change the username and password for the Nextcloud admin user"
 while :
 do
-    print_text_in_color "$ICyan" "Please enter the username for your new user:"
-    read -r NEWUSER
-    sudo -u www-data $NCPATH/occ user:add "$NEWUSER" -g admin && break
+    NEWUSER=$(input_box_flow "Please type in the name of the Nextcloud admin user. It must differ from [$NCADMIN].\nThe only allowed character are: 'a-z', 'A-Z', '0-9', and '_.@-'")
+    if [[ "$NEWUSER" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
```
szaimen, line 358: We don't have to check spaces explicitely if we check for the pattern

enoch85: I'm not 100% sure about the pattern. What you say is correct, but I haven't tested the pattern enough to be 100% sure.

Busy with the official VM today, need to produce a POC.

szaimen: I thought the official was already working?

enoch85: Yes, but they want changes...

enoch85: 7 hours straight now. Finally a first POC. 

Now I will eat and all that other important stuff. :D

nextcloud-startup-script.sh
enoch85: Can you please test all possible combinations before we commit this?

enoch85: The current implementation worked in my testing.

enoch85: I made a little test function:

```
while :
do
read -r NEWUSER
    #if [[ "$NEWUSER" =~ [^a-zA-Z0-9_.@-] ]]
    if ! [[ "$NEWUSER" =~ ^[a-zA-Z0-9_.@-]+$ ]]
    then
        echo "Allowed characters are: a-z', 'A-Z', '0-9', and '_.@-'\nPlease try again."
    else
        echo "VALID PASSWORD!"
    fi
done
```

Please use that and test different scenarios. We still don't cover ÅÄÖ for example, it shows up as valid, and if that happens, the whole thing never stops looping.

enoch85: Both yourrs and mine produce exactly the same result btw.

szaimen: okay great, then just use the one that you like more :D

szaimen: I trust you if you say it works :)
I am currently working on the new smbserver-script ;)

nextcloud-startup-script.sh
```diff
@@ -330,27 +330,58 @@ bash $SCRIPTS/nextcloud_configuration.sh
...
+        break
+    fi
+done
+while :
+do
+    OC_PASS=$(input_box_flow "Please type in the new password for the new nextcloud-admin user $NEWUSER")
+    if [[ "$OC_PASS" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    fi
```
szaimen, line 375: Here is a mistake: we want to bring up the input-box directly after this was positive and not continue, no?

enoch85: I don't know, you wrote it. :)

I have some time today (the whole day) for testing, and I was planning to work on the Wordpress VM, but I can chime in here as well from time to time.

nextcloud-startup-script.sh
enoch85: We still need to figure out how to not allow only english characters, *not* ÅÄÖ and the german U, double S, and so on.

Regarding SMB-Server, that's another project. But maybe you can convince me, we'll see. :)

nextcloud-startup-script.sh
enoch85: I will check this and merge this today :+1:

enoch85: This prompt is needed as it's for the password, not for the user.

I don't know what yo mean here?

PR #1463, opened 2020-09-15
add a different method for getting the first iface

szaimen:
I am not sure if is more stable but it is much faster and doesn't output creepy things like the default command. 
(This way, I think we can get rid of the first_iface function)


enoch85:
Is this done?

szaimen:
> Is this done?

I would say yes 👍

PR #1464, opened 2020-09-15
create open_port function to automatically open ports over upnp

szaimen:
Of course we should discuss this first and also if we want to support this or not.
It makes port-forwarding seamless, though.

### DIFF
--- a/lets-encrypt/activate-tls.sh
+++ b/lets-encrypt/activate-tls.sh
@@ -64,6 +64,13 @@ fi
 # Nextcloud Main Domain (activate-tls.sh)
 TLSDOMAIN=$(input_box_flow "Please enter the domain name you will use for Nextcloud.\n\nMake sure it looks like this:\nyourdomain.com, or cloud.yourdomain.com")
 
+msg_box "Before you start, please make sure that port 80+443 is directly forwarded to this machine!\nAn alternative is opening those ports via UPNP automatically, which you can do in the next step."
+if yesno_box_no "Do you want to use UPNP to open those ports?"
+then
+    open_port 80 TCP
+    open_port 443 TCP
+fi
+
 # Curl the lib another time to get the correct https_conf
 # shellcheck source=lib.sh
 . <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
--- a/lib.sh
+++ b/lib.sh
@@ -709,6 +709,8 @@ open_port() {
     if ! upnpc -a "$ADDRESS" "$1" "$1" "$2" &>/dev/null
     then
         msg_box "Failed to open port $1 $2 automatically. You have to do this manually."
+        apt-get purge miniupnpc -y
+        apt autoremove -y
         return 1
     fi
 }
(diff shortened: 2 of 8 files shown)


enoch85:
It would only work if upnpc is enabled anyway.

I'm not against it since it shuold be up to the user to choose.

szaimen:
> It would only work if upnpc is enabled anyway.

Do you mean if UPNP is enabled in the router? Yes, of course, but then it just fails to open the port automatically...
Or do you mean that miniupnpc has to be installed?

> I'm not against it since it should be up to the user to choose and ft it's not enabled, then you can always do it manually.

Okay great, then I will look into how we can use this.

lib.sh
```diff
@@ -709,6 +709,8 @@ open_port() {
     if ! upnpc -a "$ADDRESS" "$1" "$1" "$2" &>/dev/null
     then
         msg_box "Failed to open port $1 $2 automatically. You have to do this manually."
+        apt-get purge miniupnpc -y
+        apt autoremove -y
```
szaimen, line 713: I was planning to integrate this into check_open_port. Isn't this a bit overkill to install and uninstall this after every command? e.g. for opening port 443 and 80 we would need to call the function 2 times...

enoch85: We open the port with this function, and then check if it's open with another function.

I'm a bit tired now though, been a long day!

enoch85: I don't want any extra leftovers in the VM, that's what --^ about

enoch85:
> Do you mean if UPNP is enabled in the router?

I never used it myself, but doesn't it have to be open in the router for it to work? What happens if UPNP are disabled, will port forwarding stop working then?

szaimen:
> I never used it myself, but doesn't it have to be open in the router for it to work? What happens if UPNP are disabled, will port forwarding stop working then?

No, port forwarding is going to work also if deactivating UPNP.
UPNP just makes it easier to setup port-forwarding since it automates this process from the client side. 
If deactivating UPNP you just have to do this manually. But port-forwarding will still work.

lib.sh
szaimen: Yes, I can understand this but it will be a performance problem if we install and deinstall it every time the function gets called. (I mean it just gets installed if you call the function, which is on `no `by default)

lets-encrypt/activate-tls.sh
```diff
@@ -64,6 +64,13 @@ fi
 # Nextcloud Main Domain (activate-tls.sh)
 TLSDOMAIN=$(input_box_flow "Please enter the domain name you will use for Nextcloud.\n\nMake sure it looks like this:\nyourdomain.com, or cloud.yourdomain.com")
 
+msg_box "Before you start, please make sure that port 80+443 is directly forwarded to this machine!\nAn alternative is opening those ports via UPNP automatically, which you can do in the next step."
+if yesno_box_no "Do you want to use UPNP to open those ports?"
+then
+    open_port 80 TCP
+    open_port 443 TCP
```
szaimen, line 71: I mean we could uninstall it after both are called... What do you think?

PR #1467, opened 2020-09-16
make fetch_lib a bit slicker

szaimen:
This should work, too


enoch85:
not tested, but makes sense.

PR #1469, opened 2020-09-17
create open_port function to automatically open ports over upnp

enoch85:
New PR for https://github.com/nextcloud/vm/pull/1464

### DIFF
--- a/apps/talk_signaling.sh
+++ b/apps/talk_signaling.sh
@@ -172,8 +172,13 @@ fi
 check_command systemctl restart coturn.service
 
 # Warn user to open port
-msg_box "You have to open $TURN_PORT TCP/UDP in your firewall or your TURN/STUN server won't work!
-After you hit OK the script will check for the firewall and eventually exit on failure.
+msg_box "You have to open $TURN_PORT TCP/UDP in your firewall or your TURN/STUN server won't work!\nAn alternative is opening those ports via UPNP automatically, which you can do in the next step."
+if yesno_box_no "Do you want to use UPNP to open those ports?"
+then
+    open_port "$TURN_PORT" TCP
+    open_port "$TURN_PORT" UDP
+fi
+msg_box "After you hit OK the script will check for the firewall and eventually exit on failure.
 To run again the setup, after fixing your firewall:
 sudo -sLO $APP/talk_signaling.sh
 sudo bash talk_signaling.sh"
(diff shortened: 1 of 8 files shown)


szaimen:
Why do we need this? it seems like https://github.com/nextcloud/vm/pull/1464 was already merged into master?

enoch85:
> Why do we need this? it seems like #1464 was already merged into master?

Screenshot from 2020-09-17 12:08:33

szaimen (commented):
Hm... I cant delete this review...

lib.sh
```diff
@@ -698,6 +698,26 @@ fi
...
+    fi
+}
+
+cleanup_open_port() {
+    if [ -n "$FAIL" ]
+    then
+        apt-get purge miniupnpc -y
+        apt autoremove -y
+    fi
+}
```
szaimen, line 719: I mean the solution with cleanup_open_port and unset FAIL?

apps/talk_signaling.sh
```diff
@@ -303,14 +320,6 @@ check_command systemctl enable signaling
 # Apache Proxy
 # https://github.com/strukturag/nextcloud-spreed-signaling#apache
 
-# Check if $SUBDOMAIN exists and is reachable
-print_text_in_color "$ICyan" "Checking if $SUBDOMAIN exists and is reachable..."
-domain_check_200 "$SUBDOMAIN"
-
-# Check open ports with NMAP
-check_open_port 80 "$SUBDOMAIN"
```
enoch85, line 311: why are these removed?

enoch85: Sorry, they are moved.

lib.sh
enoch85: Looks good to me. :+1:

enoch85: But I think thet could be in the same funtion. I will fix.

enoch85: Actually, no you're right, this way is better.

enoch85:
> Hm... I cant delete or hide this comment...

Report the bug to Github :laughing:

szaimen:
@enoch85 this looks strange to me:

enoch85:
No, it's correct. The port check happens after the try to use UPNP.

szaimen:
> No, it's correct. The port check happens after the try to use UPNP - which is after you hit OK.

You didn't get what I meant: it is not indide the msg_box?

PR #1470, opened 2020-09-17
SMB-mount - dramatically enhance the mounting options in Nextcloud via files_external

szaimen:
I've tested this for a while and couldn't find any issue with this new part.


enoch85:
Is this the server you were taking about?

szaimen:
> Is this the server you were taking about?

No, this is an enhancement to the smb-mount script ;)

szaimen:
I think this is ready to go

PR #1473, opened 2020-09-18
update the lib also if  internet is working

szaimen:
I think this makes sense since if the internet works we can also update the lib (should reduce compatibility problems)

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -217,6 +217,8 @@ download_script MENU menu
 download_script MENU server_configuration
 download_script MENU nextcloud_configuration
 download_script MENU additional_apps
+# This is needed to update the lib once during the startup script
+download_script GITHUB_REPO lib
 
 if home_sme_server
 then
--- a/static/fetch_lib.sh
+++ b/static/fetch_lib.sh
@@ -25,7 +25,7 @@ then
     then
         print_text_in_color "$IRed" "You don't seem to have an internet connection and the local lib isn't available. Hence you cannot run this script."
         exit 1
-
+    fi
 elif ! [ -f /var/scripts/nextcloud-startup-script.sh ]
 then
     print_text_in_color "$ICyan" "Updating lib..."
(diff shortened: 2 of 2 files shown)


enoch85:
> update the lib also if internet is working 

I don't know what checking if that file has to do with internet?

szaimen:
> > update the lib also if internet is working 
> 
> I don't know what checking if that file has to do with internet?

I can explain it to you:

If you are so far in the startup script that the you cannot run the startup script twice file is created, you have a working internet connection. That is the reason why I think it makes sense to also update the lib if this file exists. Otherwise you will always don't update the lib during the whole startup script which can cause complications if you run scripts directly from the repository which may need a newer lib.

szaimen:
We could alternatively just download a new library once while downloading all other scripts which might be a better approach...

enoch85:
Actually, in the new official VM you don't need an internet connection at all to run the first setup. I think that should be the default in this as well. 

I'm feeling more and more confident that `fetch_lib.sh` will work in all scripts now, especially since we added it to the startup script.

But sure, we can merge this for now.

szaimen:
Please wait I would like to change the implementation a bit

szaimen:
I think this is done now

nextcloud-startup-script.sh
```diff
@@ -217,6 +217,8 @@ download_script MENU menu
 download_script MENU server_configuration
 download_script MENU nextcloud_configuration
 download_script MENU additional_apps
+# This is needed to update the lib once during the startup script
+download_script GITHUB_REPO lib
```
enoch85, line 221: This is one step away from my plan with running the first startup without internet. Please revert.

enoch85: The other solution was better.

szaimen: If your plan is to run the whole startup script without internet, is the current implementation in master the best.
But how did you plan to get rid of run_script in every menu? 

I would suggest cloning the whole repo furing the install_script and changing run_script and download_script to use the local script if existing.

enoch85: > run_script in every menu?

Haven't figured that out yet, but maybe skip the whole menu section if the user doesn't have internet. They can always run it later.

enoch85:
Closing as we decided that the current master is better,

nextcloud-startup-script.sh
szaimen: Of course this is also a solution, yes 👍
It might also be much easier to implement

PR #1479, opened 2020-09-19
update smbmount to allow to select no Nextcloud user and no group while creating the storage

szaimen:
as the title says


szaimen:
I think this is ready

PR #1480, opened 2020-09-19
POC how to make the execution of not ready scripts much more easy

szaimen:
As the title says.
@enoch85 what do you think about this? Would this be okay?


enoch85:
I'm fine with everything except putting the script in the released version: `nextcloud-startup-script.sh`

nextcloud-startup-script.sh
```diff
@@ -217,6 +217,7 @@ download_script MENU menu
 download_script MENU server_configuration
 download_script MENU nextcloud_configuration
 download_script MENU additional_apps
+download_script NOT_SUPPORTED not-supported
```
szaimen, line 220: > I'm fine with everything except putting the script in the released version: `nextcloud-startup-script.sh`

Do you refer to this line?

enoch85: Yes.

szaimen: okay, I've removed it from the startup script

nextcloud_install_production.sh
```diff
@@ -824,6 +824,7 @@ download_script GITHUB_REPO nextcloud-startup-script
 download_script STATIC instruction
 download_script STATIC history
 download_script NETWORK static_ip
+download_script NOT_SUPPORTED not-supported
```
szaimen, line 827: would it be okay in the install_production script?

enoch85: hehe, no, same thing. :)

szaimen: In this case, I guess this PR doesn't make much sense at all

szaimen:
Closing this to keep the not-supported folder clean.

Issue #1459, opened 2020-09-14
WTH is my bitwarden containers?


mamama1:
Hi

I have been using my nextcloud VM for some time now and suddenly Bitwarden stopped working (503 Service Unavailable). So I went for checking the docker containers and boom, nothing there. That happened before but back then, a reboot solved it.
Now there are no bitwarden containers, even after a reboot of the VM.

```
root@files:/# docker ps -a
CONTAINER ID        IMAGE                   COMMAND                  CREATED             STATUS              PORTS                                                NAMES
1ffab21dc24e        containrrr/watchtower   "/watchtower --clean…"   4 weeks ago         Up 2 days           8080/tcp                                             watchtower
d648be10c8fe        ark74/nc_fts            "/tini -- /usr/local…"   4 weeks ago         Up 2 days           127.0.0.1:9200->9200/tcp, 127.0.0.1:9300->9300/tcp   fts_esror
```

not sure, whether the /dev/loop* stuff is supposed to be full according to df -h and whether that has anything to do with bitwarden containers not being there anymore.

```
root@files:/home/ncadmin# df -h
Filesystem                         Size  Used Avail Use% Mounted on
udev                               3.9G     0  3.9G   0% /dev
tmpfs                              797M  2.3M  794M   1% /run
/dev/mapper/ubuntu--vg-ubuntu--lv   35G   16G   18G  47% /
tmpfs                              3.9G  8.0K  3.9G   1% /dev/shm
tmpfs                              5.0M     0  5.0M   0% /run/lock
tmpfs                              3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/sda2                          976M  197M  713M  22% /boot
/dev/loop0                          48M   48M     0 100% /snap/certbot/500
/dev/loop1                          53M   53M     0 100% /snap/certbot/579
/dev/loop3                          55M   55M     0 100% /snap/core18/1880
/dev/loop2                          56M   56M     0 100% /snap/core18/1885
/dev/loop4                          61M   61M     0 100% /snap/core20/634
/dev/loop5                          71M   71M     0 100% /snap/lxd/16889
/dev/loop6                          30M   30M     0 100% /snap/snapd/8542
/dev/loop7                          30M   30M     0 100% /snap/snapd/8790
/dev/loop8                          71M   71M     0 100% /snap/lxd/16926
ncdata                             985G   17G  968G   2% /mnt/ncdata
[...]

PR #1482, opened 2020-09-20
only update specific docker containers

enoch85:

Fix https://github.com/nextcloud/vm/issues/1459


enoch85:
Tested and works as intended!

PR #1483, opened 2020-09-20
smbmount - allow to use slashes in the external-storage-name

szaimen:
as the title says


szaimen:
I think this is ready. 
Sry for bringing those enhancements part for part and not just all at once but this is how thos ideas come ;)

PR #1489, opened 2020-09-22
add the inotify-option to smbmount

szaimen:
This should fix the problem that the customer had, whose mail you recently forwarded to me.


enoch85:
Thanks @szaimen ❤️

PR #1490, opened 2020-09-22
smbserver: implement the notify option there, too

szaimen:
this is actually the same implementation like #1489 in smbmount, but I had to change some small things
(e.g. use return instead of break)


enoch85:
👍

PR #1492, opened 2020-09-23
allow the system to send mails

szaimen:
I though this can be useful.

- [ ] 1. Question is, if this should be enabled by default during the startup-script (startup_switch)
- [ ] 2. and where exactly we can make use of this. (Maybe calling it directly from notify_admin_gui?)
- [ ] 3. Also it could be in startup_configuration? Or does it fit good in here in server_configuration?
- [ ] 4. We could also move it to the top in the server_configuration or so...

Otherwise I successfully tested this with a mailaccount.

### DIFF
--- a/addons/smtp-mail.sh
+++ b/addons/smtp-mail.sh
@@ -90,7 +90,7 @@ RECIPIENT=$(input_box_flow "Please enter the recipient mail-address that shall r
 msg_box "These are the settings that will be used. Please check that everything seems correct.
 
 Mailserver URL=$MAIL_SERVER
-Protocol=$PROTOCOL
+Encryption=$PROTOCOL
 SMTP-port=$SMTP_PORT
 SMTP-username=$MAIL_USERNAME
 SMTP-password=$MAIL_PASSWORD
--- a/addons/smtp-mail.sh
+++ b/addons/smtp-mail.sh
@@ -51,26 +51,31 @@ install_if_not mailutils
 MAIL_SERVER=$(input_box_flow "Please enter the email server URL that you want to use.\nE.g. smtp.mail.com")
 
 # Enter if you want to use ssl
-while :
-do
-    PROTOCOL=$(input_box_flow "Please type in the encryption protocol for your mailserver.
-The available options are 'SSL', 'STARTTLS' or 'none'.")
-    if [ "$PROTOCOL" = "SSL" ]
-    then
+PROTOCOL=$(whiptail --title "$TITLE" --nocancel --menu \
+"Please choose the encryption protocol for your mailserver.
+$MENU_GUIDE\n\n$RUN_LATER_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+"SSL" "" \
+"STARTTLS" "" \
+"none" "" 3>&1 1>&2 2>&3)
+
+if [ -z "$PROTOCOL" ]
+then
+    exit 1
+fi
+
+case "$PROTOCOL" in
+    "SSL")
         DEFAULT_PORT=465
-        break
-    elif [ "$PROTOCOL" = "none" ]
-    then
-        DEFAULT_PORT=25
-        break
-    elif [ "$PROTOCOL" = "STARTTLS" ]
-    then
+    ;;
+    "STARTTLS")
         DEFAULT_PORT=587
-        break
-    else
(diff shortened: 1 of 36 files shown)


szaimen:
Btw: this was my guide:
https://decatec.de/linux/linux-einfach-e-mails-versenden-mit-msmtp/
Should we somewhere mention him?

lib.sh
```diff
@@ -1409,6 +1409,22 @@ do
...
+send_mail() {
+    local RECIPIENT
+    if [ -f /etc/msmtprc ]
+    then
+        RECIPIENT=$(grep "recipient=" /etc/msmtprc)
+        RECIPIENT="${RECIPIENT##*recipient=}"
+        if [ -n "$RECIPIENT" ]
+        then
+            print_text_in_color "$ICyan" "Sending '$1' to $RECIPIENT"
+            echo -e "$2" | mail --subject "$1" "$RECIPIENT"
```
szaimen, line 1423: We could maybe also prepend `NcVM - ` to the subject

szaimen: this was added

lib.sh
```diff
@@ -1392,6 +1392,7 @@ esac
 #
 # occ_command_no_check notification:generate -l "$2" "$admin" "$1"
 notify_admin_gui() {
+send_mail "$1" "$2"
```
szaimen, line 1395: I've put it into notify_admin_gui

szaimen: Alternatively we could add it manually to all sensible places.

szaimen:
@enoch85 do you have some thoughts on this?

enoch85:
> @enoch85 do you have some thoughts on this?

Not yet.Requires some testing, which requires time, which I don't have right now. :/

addons/smtp-mail.sh
```diff
@@ -0,0 +1,183 @@
...
+install_if_not msmtp-mta
+install_if_not mailutils
+
+# Enter mailserver
+MAIL_SERVER=$(input_box_flow "Please enter the email server URL that you want to use.\nE.g. smtp.mail.com")
+
+# Enter if you want to use ssl
+while :
+do
+    PROTOCOL=$(input_box_flow "Please type in the encryption protocol for your mailserver.
```
enoch85, line 56: Please make this a whiptail with options instead. It's more user friendly.

szaimen: done

szaimen:
What about my questions above? (especially the 4 points to discuss in the PR-explainer?)

enoch85:
If you want to get this merged quickly (and so that users could test) we could move it to not supported for now.

It would make it easier for me to test as well.

szaimen:
> If you want to get this merged quickly (and so that users could test) we could move it to not supported for now.
> 
> It would make it easier for me to test as well.

I've put it into the not-supported menu. 👍

lib.sh
```diff
@@ -1425,6 +1426,22 @@ do
...
+        RECIPIENT=$(grep "recipient=" /etc/msmtprc)
+        RECIPIENT="${RECIPIENT##*recipient=}"
+        if [ -n "$RECIPIENT" ]
+        then
+            print_text_in_color "$ICyan" "Sending '$1' to $RECIPIENT"
+            echo -e "$2" | mail --subject "NcVM - $1" "$RECIPIENT"
+        fi
+    fi
+}
+
```
szaimen, line 1444: BTW: I am voting to leave this in here, because you can then properly test if this works.

It will only send a mail if smtp mail was setup.

not-supported/not-supported_menu.sh
```diff
@@ -17,6 +17,14 @@ debug_mode
...
 root_check
 
+# Set the startup switch
+if [ -f "$SCRIPTS/nextcloud-startup-script.sh" ]
+then
+    STARTUP_SWITCH="ON"
+else
+    STARTUP_SWITCH="OFF"
+fi
+
```
szaimen, line 27: we only need this as long as smtp mail is in here. I will remove it once it is in server_configuration, again.

lib.sh
enoch85: Sure.

enoch85:
Ready to merge?

not-supported/not-supported_menu.sh
enoch85: Sure. :+1:

szaimen:
> Ready to merge?

I would say yes 👍

PR #1496, opened 2020-09-26
add an option to the not-supported folder

szaimen:
I wonder if this new option is okay inside the not-supported folder?
I really think this makes sense and is a valuable addition especially inside this folder ;)


enoch85:
This is turning out to be something completely different from Nextcloud.

But sure, since it's not supported,that could be anything.

PR #1498, opened 2020-09-28
Add ClamAV as AV-protection

szaimen:
As the title says


apps/clamav.sh
```diff
@@ -0,0 +1,142 @@
...
+# stop freshclam and update the database
+check_command systemctl stop clamav-freshclam
+check_command freshclam
+check_command systemctl start clamav-freshclam
+
+# Edit ClamAV settings to fit the installation
+sed -i "s|^MaxDirectoryRecursion.*|MaxDirectoryRecursion 30|" /etc/clamav/clamd.conf
+sed -i "s|^MaxFileSize.*|MaxFileSize 100M|" /etc/clamav/clamd.conf
+sed -i "s|^PCREMaxFileSize.*|PCREMaxFileSize 100M|" /etc/clamav/clamd.conf
+sed -i "s|^StreamMaxLength.*|StreamMaxLength 100M|" /etc/clamav/clamd.conf
```
szaimen, line 67: Does 100 MB make sense?

apps/clamav.sh
```diff
@@ -0,0 +1,142 @@
...
+check_command systemctl restart clamav-daemon
+
+# Install Nextcloud app
+install_and_enable_app files_antivirus
+
+# Configure Nextcloud app
+occ_command config:app:set files_antivirus av_mode --value="socket"
+occ_command config:app:set files_antivirus av_socket --value="/var/run/clamav/clamd.ctl"
+occ_command config:app:set files_antivirus av_stream_max_length --value="104857600"
+occ_command config:app:set files_antivirus av_max_file_size --value="-1"
```
szaimen, line 87: Does 100 MB but unrestricted av_max_file_size make sense?

apps/clamav.sh
```diff
@@ -0,0 +1,142 @@
...
+occ_command config:app:set files_antivirus av_infected_action --value="only_log"
```
szaimen, line 88: should we better remove the file?

apps/clamav.sh
```diff
@@ -0,0 +1,142 @@
...
+debug_mode
+
+# Check if root
+root_check
+
+# Show explainer
+explainer_popup
+
+# Needs 1 GB alone
+ram_check 3 "ClamAV"
```
enoch85, line 27: Why do we need 1 extra GB here?

apps/clamav.sh
```diff
@@ -0,0 +1,142 @@
...
+. <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
+AV_REPORT="$(clamscan \
+--recursive \
+--stdout \
+--infected \
+--cross-fs \
+--log=$VMLOGS/clamav-fullscan.log \
+--max-scantime=43200000 \
+--max-filesize=100M \
```
szaimen, line 118: Does 12h max-scantime makes sense? (the value is in ms)
Also, does 100 MB make sense?

apps/clamav.sh
szaimen: Because the ClamAV-deamon alone needs 1GB and since php-fpm is recommended to run with at least 2GB of available RAM, I though this would make sense.

apps/clamav.sh
szaimen: We could also add the option to move infected files to a different directory...

apps/clamav.sh
enoch85: What's the default? 

I mean 100 MB would only scan small files. Make it 50 GB maybe?

apps/clamav.sh
```diff
@@ -0,0 +1,142 @@
...
+--max-dir-recursion=30 \
+/ )"
```
szaimen, line 120: We could also add the option to move infected files to a different directory...

enoch85: Sure, but it should be up to the user to decide.

apps/clamav.sh
enoch85: OK :+1:

apps/clamav.sh
szaimen: The default is 25 MB

apps/clamav.sh
szaimen: Okay, So shall I add a whiptail where the use can choose what shall happen? (remove it/move it/copy it/just logg it)

apps/clamav.sh
szaimen: Problem is this is written for all those size options:
`Warning: disabling this limit or setting it too high may result in severe damage to the system.`

apps/clamav.sh
szaimen: Problem is this is written for max filesize:
`Warning: disabling this limit or setting it too high may result in severe damage to the system.`

szaimen: https://manpages.debian.org/testing/clamav-daemon/clamd.conf.5.en.html

apps/clamav.sh
szaimen: https://manpages.debian.org/testing/clamav-daemon/clamd.conf.5.en.html

enoch85:
Please ask the maintainer of the `files_antivirus` app if the above is sane defaults, I have no clue. :)

szaimen:
Here is the question: https://github.com/nextcloud/files_antivirus/issues/172

szaimen:
This is what I have used as guide:
https://markus-blog.de/index.php/2018/05/23/how-to-install-antivirus-app-in-nextcloud-with-clamav/

enoch85:
> The recommendation there was 100M

OK, let's go with 100M then.

apps/clamav.sh
enoch85: Nope

apps/clamav.sh
enoch85: Yes please.

szaimen: done

apps/clamav.sh
```diff
@@ -0,0 +1,185 @@
...
+    "")
+        exit 1
+    ;;
+    *)
+    ;;
+esac
+
+if [ "$choice" = "Copy to a folder" ] || [ "$choice" = "Move to a folder" ]
+then
+    AV_PATH="/root/.clamav/clamav-fullscan.jail"
```
szaimen, line 135: Is this location good?

enoch85: Doesn't clamav have it's on folder?

Is this for logging?

szaimen: I am not sure if clamav has its own folder.
no, this is for storing the infected files.

szaimen:
> @szaimen Isn't this better?

yes, I guess so...

apps/clamav.sh
```diff
@@ -0,0 +1,186 @@
...
+# Edit ClamAV settings to fit the installation
+sed -i "s|^MaxDirectoryRecursion.*|MaxDirectoryRecursion 30|" /etc/clamav/clamd.conf
+sed -i "s|^MaxFileSize.*|MaxFileSize 100M|" /etc/clamav/clamd.conf
+sed -i "s|^PCREMaxFileSize.*|PCREMaxFileSize 100M|" /etc/clamav/clamd.conf
+sed -i "s|^StreamMaxLength.*|StreamMaxLength 100M|" /etc/clamav/clamd.conf
+
+# Start ClamAV
+check_command systemctl restart clamav-freshclam
+check_command systemctl restart clamav-daemon
+
```
enoch85, line 73: We should check for the existence of `/var/run/clamav/clamd.ctl` to be sure the daemon works properly - and then continue so that we don't clutter the system before we know everything is good.

apps/clamav.sh
```diff
@@ -0,0 +1,186 @@
...
+notify_admin_gui \
+"Your weekly full-scan ClamAV report" \
+"$AV_REPORT"
+CLAMAV_REPORT
+
+# Make the script executable
+chmod +x "$SCRIPTS"/clamav-fullscan.sh
+
+# Create the cronjob
+crontab -u root -l | { cat; echo "0 10 * * 7 $SCRIPTS/clamav-fullscan.sh 2>&1"; } | crontab -u root -
```
enoch85, line 175: Have you been taking the automatic updates script in consideration here? I'm thinking of the reboots....

szaimen: I thought updates happen on saturday evening?

szaimen: https://github.com/nextcloud/vm/blob/476ea4e96a65dd37a0b0da9a9335c793d6fd33ee/addons/automatic_updates.sh#L39

apps/clamav.sh
szaimen: you mean something like this?
```
print_text_in_color "$ICyan" "Waiting for clamd to start up..."
while ! [ -f /var/run/clamav/clamd.ctl ]
do
    sleep 5
done
```

I mean we would need to have a check so that it doesnt run endless

szaimen:
BTW: I wonder if the full-scan can really scan all files if it runs as clamav-user or is the clamav-user member of the root group? 
based on this website: https://www.howtoforge.com/tutorial/clamav-ubuntu/
I think it should work:
image

apps/clamav.sh
enoch85: OK great!

enoch85:
> I think it should work:

We're only interested in `/mnt/ncdata` and other Nextcloud related paths. For now at least.

apps/clamav.sh
enoch85: Maybe an until loop is better? 

Until like 10 retries or something.. and after than tell the user that something failed and report to $ISSUES.

szaimen: done

szaimen:
What about this one?

szaimen:
Also add it to not-supported first?

enoch85:
> Also add it to not-supported first?

Yes please.

szaimen:
> Yes please.

done

Issue #1497, opened 2020-09-27
Docker error on update.sh script


silenx:
Hello,
i’m using NC 19.0.3 on ubuntu server 18.04

This is the output of docker ps -a
28245ad4d08c collabora/code:latest “/bin/sh -c 'bash st…” 8 weeks ago Up 34 minutes 192.168.99.5:9980->9980/tcp co
de
17baf414376b ark74/nc_fts:latest “/tini – /usr/local…” 2 months ago Up 34 minutes 0.0.0.0:9200->9200/tcp, 0.0.0.0:9300->9300/tcp ft
s_esror
e2d62754b836 containrrr/watchtower:latest “/watchtower --clean…” 3 months ago Up 34 minutes 8080/tcp wa
tchtower

-->
## Steps to reproduce
1. Run update.sh script
2.
3.

## Actual behaviour
docker: Error response from daemon: Conflict. The container name “/watchtower” is already in use by container “e2d62754b83669e714bbb0b3d448adc978cb6a9687c75579dcc0513e
dea2287b”. You have to remove (or rename) that container to be able to reuse that name.
See ‘docker run --help’.
Collabora docker image just got updated!
Posting notification to users that are admins, this might take a while…
Posting ‘Docker image just got updated!’ to: B792C65F-960E-4F12-A752-DA5768194273
Posting ‘Docker image just got updated!’ to: ncadmin2
docker: Error response from daemon: Conflict. The container name “/watchtower” is already in use by container “e2d62754b83669e714bbb0b3d448adc978cb6a9687c75579dcc0513e
dea2287b”. You have to remove (or rename) that container to be able to reuse that name.
See ‘docker run --help’.
Full Text Search docker image just got updated!

Script finish his cycle without error.

## Server configuration

**Nextcloud server version: 19.0.3

**Server OS** Ubuntu Server 10.04.5 LTS

**How did you install the VM?** (Scripted install from master **OR** Released version)
Script from master

enoch85:
Thanks for reporting this!

Do you have Bitwarden installed on your system?

silenx:
> Thanks for reporting this!
> 
> Do you have Bitwarden installed on your system?

No, server is dedicated to nextcloud only

Danteba:
Hello, I'm receiving the same error...

[...]

PR #1500, opened 2020-09-29
change the name of the watchtower container

szaimen:
should fix https://github.com/nextcloud/vm/issues/1497

We could alternatively remove watchtower on every instance and not only on instances that have bitwarden installed.
Then the docker run watchtower command should work, too. (Problem is if the watchtower container (with this name) is already present on an instance -> docker cannot create a second one with the same name.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -1307,7 +1307,7 @@ fi
 docker_update_specific() {
 if does_this_docker_exist "$1" "$2"
 then
-    docker run --rm --name watchtower -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once "$1"
+    docker run --rm --name temporary-watchtower -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once "$1"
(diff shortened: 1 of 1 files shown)


lib.sh
```diff
@@ -1307,7 +1307,7 @@ fi
 docker_update_specific() {
 if does_this_docker_exist "$1" "$2"
 then
-    docker run --rm --name watchtower -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once "$1"
+    docker run --rm --name temporary-watchtower -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once "$1"
```
szaimen, line 1310: I am not sure if hyphens are allowed for docker container names?
We can also use a different one then this. it just has to be something else than `watchtower`

enoch85:
Great solution, never even thought of this. 

Is this ready to merge?

enoch85:
Seems OK to merge, so I'll go ahead and do it.

PR #1503, opened 2020-09-30
rename not-supported to home-server

szaimen:
This is my vision for the not-supported folder. The applications in the not-supported folder are clearly meant for Home servers.
Because of that I made it possible to separate between Home servers and other servers.

Home servers will see a new option inside the main_menu and during the startup-script where they can choose to install those additional applications. I am willed to support those people here on Github, if bugs should appear.

By limiting the amount of people that can see this home server menu, I think it will be a managable amount of additional support.

Also offering those additional options to home servers will be a very valuable addition for those users (also in terms of marketing)

I hope you understand my motivation and aggree that it makes sense.

Thanks for reading!


enoch85:
I understand.

Though this is not a "Home Server" repo, this is "Nextcloud". Sure, Nextcloud can be run on a "home server", but that's not what this repo is about. This PR just takes this further away from the original idea - to bring Nextcloud (and only Nextcloud) to users as easily as possible.

We can start another repo outside of the Nextcloud Github, but it's not suitable here.

I'm not against adding whatever to the "not-supported" folder, but:
1. It should not at all be included in the released version
2. It should be very clear that it has nothing to do with Nextcloud
3. It will probably never end up in the released version, if not requested by (many) users.

This repo in general, and especially the released version, should be enterprise ready, not some hobby project IMHO.

I like the change from "script" to "Menu" though.

szaimen:
Thank you for your honest answer! I really appreciate it. 👍

In my opinion is the Nextcloud VM a really good starting point for Home servers, too because it already brings the whole Nextcloud App Ecosystem with it and already has some good integrations and services. Also it is also already widely etablished, which is great to reach many users. To make it really really great as a home server, there are still some things (applications) missing, though which I will try to add in the following weeks/months. 

The biggest disadvantage with running the Nextcloud VM on a home-server without having all services integrated, that I currently see is, that most users only have one public ipaddress so you can use only one server that has to proxy all requests, which is not easily doable without making it really complex (with more devices and such) and is much slicker when everything runs from one device, imo.

As long as you say it is okay to add more scripts to the not-supported folder, I think it will be okay to me as it is today (with the not-supported folder and so on).  The only disadvantage is, that interested users need to download one file initially and then have the menu separated. So this should probably be okay. I would in this case go ahead and open PRs with additional scripts to the not-supported folders.

> * It should not at all be included in the released version
> * It should be very clear that it has nothing to do with Nextcloud
> * It will probably never end up in the released version, if not requested by (many) users.

As I see it, the current version (master) respects all three points?

Are you okay if I close this PR? I would then open a PR with the change from "script" to "Menu".

szaimen:
BTW: what do you think in general about making those options visible to home users by trying to find out if it is a home server or not? Because I think this would solve all concerns if we could somehow find out if it is a home server or an organization server.

In other words: if you have the problem with the following line, I could try to find a better solution.
`if lshw -c system -quiet | grep "product:" | grep -q " NUC"`

What about the idea to ask during the initial setup if it shall be a home or business installation? We could also explain what the difference would be (showing additional apps). Then the admin could decide during the startup script what it shall be.
Also we could only show this question on NUC systems as it was planned in the beginning, btw ;)

szaimen:
I think this makes more sense to be discussed by mail. I've send you one.

szaimen:
If we should ever decide that it makes sense, I can still reopen this PR and restore the branch.

Issue #1481, opened 2020-09-20
vm os suggestion: use thin lv instead of regular lv


mailinglists35:
with manual partitioning in terminal before launching the installer from the live iso, ubuntu 20.04 can be installed on thin lv, which gives the advantage that issuing fstrim would release the storage blocks all the way down to the last device that supports discards.

for example I could fill the vm space then I could release it after mass deleting files so the backing qcow2 files would free it's blocks when guest is configured with `discard='unmap'`

don't forget to enable less than 100 thin_pool_autoextend_threshold value in lvm.conf and thin_pool_autoextend_percent

random example setup steps https://www.theurbanpenguin.com/thin-provisioning-lvm2/

enoch85:
@mailinglists35 Thanks for your suggestion.

enoch85:
Wouldn't this allow overcommit of the storage? Not a good thing to add in a default setup. 

The user would then have to have a really close look at the storage at all times, which I don't think should be a default setting.

enoch85:
@mailinglists35 Hey, still no answer?

szaimen:
If I understand this issue correctly, this should be fixed with https://github.com/nextcloud/vm/pull/1518
@mailinglists35 please reopen if this isn't the case.

PR #1505, opened 2020-10-02
allow to use the LVM configuration as it was setup by the installer

szaimen:
To be able to use features like lvm snapshots, it is incredibly useful to have the choice to not run the resize2fs comand. 

I hope this way it is okay (you just hit enter to continue as usual)

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -115,10 +115,9 @@ then
     # Resize LVM (live installer is &%¤%/!
     # VM
     print_text_in_color "$ICyan" "Extending LVM, this may take a long time..."
-    lvextend -l 100%FREE --resizefs /dev/ubuntu-vg/ubuntu-lv
+    lvextend -l +100%FREE /dev/ubuntu-vg/ubuntu-lv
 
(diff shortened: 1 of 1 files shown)


enoch85:
While digging into LVM, maybe you could have a look at this? https://github.com/nextcloud/vm/issues/1481

Are you saying that you can't do snapshots if you do `rezie2fs`?

szaimen:
> While digging into LVM, maybe you could have a look at this? #1481

Maybe after this, yes. The thing is, I don't completely understand what the guy requested in this issue and the link doesn't fit, either?

> Are you saying that you can't do snapshots if you do `rezie2fs`?

Yes. I cannot create LVM-snapshots if I do resize2fs. ZFS snapshots still work, though.

enoch85:
You know that this will give you a full sized PV, but it won't give you the full sized LV, right?

szaimen:
> You know that this will give you a full sized PV, but it won't give you the full sized LV, right?

Yes of course 👍

enoch85:
Sure, why not.

PR #1506, opened 2020-10-02
Disable hibernation

szaimen:
Imo, hibernation doesn't make sense, does it?

We could alternatively ask the user...

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -845,6 +845,9 @@ then
     check_command sed -i "s|VM|Home/SME Server|g" $SCRIPTS/nextcloud-startup-script.sh
 fi
 
+# Disable hibernation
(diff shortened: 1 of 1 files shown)


enoch85:
Makes sense.

Thanks!

enoch85:
Question is, does this apply to servers? 

Do you have a source?

szaimen:
> Question is, does this apply to servers?
> 
> Do you have a source?

I have already tested this successfully on a test machine.

enoch85:
> I have already tested this successfully on a test server

That's great! But, if I chose to leave my server on 24/7 it would never hibernate anyway, that's why I'm asking if this is really needed?

enoch85:
Hmm, seems like it could happen in some rare cases: https://www.unixtutorial.org/disable-sleep-on-ubuntu-server/

PR #1509, opened 2020-10-03
some small fixes to smtp-mail

szaimen:
as the title says

### DIFF
--- a/addons/smtp-mail.sh
+++ b/addons/smtp-mail.sh
@@ -104,14 +104,14 @@ esac
 # Enter your SMTP username
 if yesno_box_yes "Does $MAIL_SERVER require any credenitals, like username and password?"
 then
-    MAIL_USERNAME=$(input_box_flow "Please enter the SMTP username to your email provider.\nE.g. you@$MAIL_SERVER")
+    MAIL_USERNAME=$(input_box_flow "Please enter the SMTP username to your email provider.\nE.g. you@mail.com")
 
     # Enter your mailuser password
     MAIL_PASSWORD=$(input_box_flow "Please enter the SMTP password to your email provider.")
 fi
 
 # Enter the recipient
-RECIPIENT=$(input_box_flow "Please enter the recipient email address that shall receive all mails.\nE.g. recipient@$MAIL_SERVER")
+RECIPIENT=$(input_box_flow "Please enter the recipient email address that shall receive all mails.\nE.g. recipient@mail.com")
 
 # Present what we gathered, if everything okay, write to files
 msg_box "These are the settings that will be used. Please check that everything seems correct.
@@ -132,16 +132,16 @@ fi
 # Add the encryption settings to the file as well
 if [ "$PROTOCOL" = "SSL" ]
 then
-    export MSMTP_ENCRYPTION1="tls             on"
-    export MSMTP_ENCRYPTION2="tls_starttls    off"
+    MSMTP_ENCRYPTION1="tls             on"
+    MSMTP_ENCRYPTION2="tls_starttls    off"
 elif [ "$PROTOCOL" = "STARTTLS" ]
 then
-    export MSMTP_ENCRYPTION1="tls             on"
-    export MSMTP_ENCRYPTION2="tls_starttls    on"
+    MSMTP_ENCRYPTION1="tls             on"
+    MSMTP_ENCRYPTION2="tls_starttls    on"
 elif [ "$PROTOCOL" = "NO-ENCRYPTION" ]
 then
-    export MSMTP_ENCRYPTION1="tls             off"
-    export MSMTP_ENCRYPTION2="tls_starttls    off"
+    MSMTP_ENCRYPTION1="tls             off"
+    MSMTP_ENCRYPTION2="tls_starttls    off"
 fi
 
 # Check if auth should be set or not
@@ -189,7 +189,7 @@ tls_trust_file  /etc/ssl/certs/ca-certificates.crt
 account         $MAIL_SERVER
 host            $MAIL_SERVER
 port            $SMTP_PORT
-from            $MAIL_USERNAME@$MAIL_SERVER
+from            $MAIL_USERNAME
 user            $MAIL_USERNAME
 password        $MAIL_PASSWORD
 
@@ -213,9 +213,9 @@ sudo chmod 0644 $VMLOGS/mail_msmtp.log
 
 # Create aliases
 cat << ALIASES_CONF > /etc/aliases
-root: $RECIPIENT
-default:$RECIPIENT
-cron: $RECIPIENT
+root: $MAIL_USERNAME
+default:$MAIL_USERNAME
+cron: $MAIL_USERNAME
 ALIASES_CONF
 
 # Define the mail-program
--- a/addons/smtp-mail.sh
+++ b/addons/smtp-mail.sh
@@ -232,11 +232,11 @@ To change the settings please check /etc/msmtprc on your server, or simply just 
 
 YOUR CURRENT SETTINGS:
 -------------------------------------------
-$(cat /etc/msmtprc)
+$(cat /etc/msmtprc | grep -v password)
 -------------------------------------------
 
 Best regards
-The Nextcloud VM team
+The NcVM team
 https://nextcloudvm.com" \
 | mail -s "Test email from your NcVM" "$RECIPIENT" &>/dev/null
 then
--- a/addons/smtp-mail.sh
+++ b/addons/smtp-mail.sh
@@ -232,7 +232,7 @@ To change the settings please check /etc/msmtprc on your server, or simply just 
 
 YOUR CURRENT SETTINGS:
 -------------------------------------------
-$(cat /etc/msmtprc | grep -v password)
+$(grep -v password /etc/msmtprc)
 -------------------------------------------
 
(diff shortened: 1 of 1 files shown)


addons/smtp-mail.sh
```diff
@@ -132,16 +132,16 @@ fi
...
-    export MSMTP_ENCRYPTION1="tls             on"
-    export MSMTP_ENCRYPTION2="tls_starttls    on"
+    MSMTP_ENCRYPTION1="tls             on"
+    MSMTP_ENCRYPTION2="tls_starttls    on"
 elif [ "$PROTOCOL" = "NO-ENCRYPTION" ]
 then
-    export MSMTP_ENCRYPTION1="tls             off"
-    export MSMTP_ENCRYPTION2="tls_starttls    off"
+    MSMTP_ENCRYPTION1="tls             off"
+    MSMTP_ENCRYPTION2="tls_starttls    off"
```
szaimen, line 144: we don't need to export the values

addons/smtp-mail.sh
```diff
@@ -189,7 +189,7 @@ tls_trust_file  /etc/ssl/certs/ca-certificates.crt
 account         $MAIL_SERVER
 host            $MAIL_SERVER
 port            $SMTP_PORT
-from            $MAIL_USERNAME@$MAIL_SERVER
+from            $MAIL_USERNAME
```
szaimen, line 192: `$MAIL_USERNAME@$MAIL_SERVER` won't work in all situations.

addons/smtp-mail.sh
```diff
@@ -213,9 +213,9 @@ sudo chmod 0644 $VMLOGS/mail_msmtp.log
 
 # Create aliases
 cat << ALIASES_CONF > /etc/aliases
-root: $RECIPIENT
-default:$RECIPIENT
-cron: $RECIPIENT
+root: $MAIL_USERNAME
+default:$MAIL_USERNAME
+cron: $MAIL_USERNAME
```
szaimen, line 218: the aliases file is for sending mails and not the recipient, afaik.

addons/smtp-mail.sh
```diff
@@ -224,7 +224,9 @@ set sendmail="/usr/bin/msmtp -t"
 DEFINE_MAIL
 
 # Test sending of mails
-if ! echo -e "Congratulations! Since this email reached you, it seems like everything is working properly. :) To change the settings please check /etc/msmtp on your server, or simply just run the setup script again." | mail -s "Test email from your NcVM" "$RECIPIENT" &>/dev/null
+if ! echo -e "Congratulations! Since this email reached you, it seems like everything is working properly. :) 
+To change the settings please check /etc/msmtp on your server, or simply just run the setup script again." \
+| mail -s "Test email from your NcVM" "$RECIPIENT" &>/dev/null
```
szaimen, line 229: this is just to make it more readable

addons/smtp-mail.sh
enoch85: Sorry, forgot while testing.

addons/smtp-mail.sh
enoch85: aliases [file]
              Replace local recipients with addresses in the aliases file.  The aliases file is a
              plain  text  file  containing mappings between a local address and a list of domain
              addresses.  A local address is defined as one without an `@' character and a domain
              address is one with an `@' character.  The mappings are of the form:
                  local: someone@example.com, person@domain.example
              Multiple  domain  addresses are separated with commas.  Comments start with `#' and
              continue to the end of the line.
              The local address default has special significance and  is  matched  if  the  local
              address  is  not found in the aliases file.  If no default alias is found, then the
              local address is left as is.
              An empty argument  to  the  aliases  command  disables  the  replacement  of  local
              addresses.  This is the default.

http://manpages.ubuntu.com/manpages/precise/en/man1/msmtp.1.html

addons/smtp-mail.sh
enoch85: It didn't work in my testing.

addons/smtp-mail.sh
```diff
@@ -104,14 +104,14 @@ esac
...
-    MAIL_USERNAME=$(input_box_flow "Please enter the SMTP username to your email provider.\nE.g. you@$MAIL_SERVER")
+    MAIL_USERNAME=$(input_box_flow "Please enter the SMTP username to your email provider.\nE.g. you@mail.com")
 
     # Enter your mailuser password
     MAIL_PASSWORD=$(input_box_flow "Please enter the SMTP password to your email provider.")
 fi
 
 # Enter the recipient
-RECIPIENT=$(input_box_flow "Please enter the recipient email address that shall receive all mails.\nE.g. recipient@$MAIL_SERVER")
+RECIPIENT=$(input_box_flow "Please enter the recipient email address that shall receive all mails.\nE.g. recipient@mail.com")
```
enoch85, line 114: Why not use the variable?

szaimen: Because
https://github.com/nextcloud/vm/pull/1509#discussion_r499128287

addons/smtp-mail.sh
enoch85: OK tested again, and it works.

addons/smtp-mail.sh
enoch85: Bit that has nothing to do with the example. 

IMHO it's nicer to show the server. 

But ok. :+1:

addons/smtp-mail.sh
enoch85: I've tested with LKDCNoini09.com (example) and it works, it just says .invalid after.

addons/smtp-mail.sh
enoch85: In other words, please revert.

addons/smtp-mail.sh
```diff
@@ -224,7 +224,21 @@ set sendmail="/usr/bin/msmtp -t"
...
+
+To change the settings please check /etc/msmtprc on your server, or simply just run the setup script again.
+
+YOUR CURRENT SETTINGS:
+-------------------------------------------
+$(cat /etc/msmtprc)
+-------------------------------------------
+
+Best regards
+The Nextcloud VM team
```
enoch85, line 239: Maybe change to NcVM..?

addons/smtp-mail.sh
szaimen: Please read this:
https://mytechiethoughts.com/linux/simple-way-to-get-email-notifications-from-a-linux-server/#the-alias-file

szaimen: photo5460666311588818772

szaimen: if I understand it correctly, we don't need the aliases file at all

addons/smtp-mail.sh
szaimen: what I mean is, that I have a mailaddress:
`test@mail.de` and the mailserver address is `smtp.mail.de`
so `$MAIL_USERNAME@$MAIL_SERVER` would produce `test@smtp.mail.de` which is wrong.

addons/smtp-mail.sh
```diff
@@ -224,8 +224,20 @@ set sendmail="/usr/bin/msmtp -t"
...
-To change the settings please check /etc/msmtp on your server, or simply just run the setup script again." \
+if ! echo -e "Congratulations!
+
+Since this email reached you, it seems like everything is working properly. :)
+
+To change the settings please check /etc/msmtprc on your server, or simply just run the setup script again.
+
+YOUR CURRENT SETTINGS:
+-------------------------------------------
+$(cat /etc/msmtprc)
```
szaimen, line 235: I don't think that we should send the content of this file via mail since the unencrypted password is stored in there.

addons/smtp-mail.sh
szaimen: That is your decision. Both are okay for me.

addons/smtp-mail.sh
szaimen: we need to separate them

addons/smtp-mail.sh
enoch85: Good point!

enoch85:
Let's merge this now, I will start working on the NC 20 release today.

addons/smtp-mail.sh
```diff
@@ -224,7 +224,21 @@ set sendmail="/usr/bin/msmtp -t"
...
-if ! echo -e "Congratulations! Since this email reached you, it seems like everything is working properly. :) To change the settings please check /etc/msmtp on your server, or simply just run the setup script again." | mail -s "Test email from your NcVM" "$RECIPIENT" &>/dev/null
+if ! echo -e "Congratulations!
+
+Since this email reached you, it seems like everything is working properly. :)
+
+To change the settings please check /etc/msmtprc on your server, or simply just run the setup script again.
+
+YOUR CURRENT SETTINGS:
+-------------------------------------------
+$(grep -v password /etc/msmtprc)
```
szaimen, line 235: does this work?

addons/smtp-mail.sh
szaimen: sry, I was wrong

addons/smtp-mail.sh
enoch85: Yup, just tested.

PR #1510, opened 2020-10-03
20.0.0

enoch85:
@szaimen Please help me fill this out with all your changes.

https://github.com/nextcloud/vm/compare/19.0.3...master


szaimen:
Im currently not at home. Can help later.
I would suggest not deleting the 19.03 changelog and just adding the 20.0 changelog at the top.

enoch85:
> I would suggest not deleting the 19.03 changelog and just adding the 20.0 changelog at the top.

We want to keep history. :)

PR #1511, opened 2020-10-03
set archive.ubuntu.com as default repo

enoch85:


enoch85:
fyi @szaimen archive.ubuntu.com is now default.

PR #1514, opened 2020-10-03
suggest to change the mirror if it is the default one

szaimen:
as the title says. I think it was wrong before.


enoch85:
I was thinking the same, but the trade off is actually worse as we can't trust the mirrors that are set in the locate_mirror script.

I will think about this a bit more.

szaimen:
> I was thinking the same, but the trade off is actually worse as we can't trust the mirrors that are set in the locate_mirror script.

Then just set it to OFF always?

enoch85:
Sure, but please leave the startup switch in there in case I  change my mind. :+1:

szaimen:
> Sure, but please leave the startup switch in there in case I change my mind. 👍

done

PR #1515, opened 2020-10-03
insert smtp-mail into server_configuration

szaimen:
I wonder if we want to merge that before the release?


enoch85:
Yup! Nice catch. :)

Issue #1513, opened 2020-10-03
Wrong minimum php version definition for NC20


rakekniven:
## Steps to reproduce
1. call '/var/scripts/update.sh'
2. Message appears pointing out that php version is not ok
<img width="931" alt="Bildschirmfoto 2020-10-03 um 16 28 20" src="https://user-images.githubusercontent.com/2069590/94994075-9d2a6780-0595-11eb-99c3-d41701066887.png">

## Expected behaviour
Update should run fine.

## Actual behaviour
Message appears

## Server configuration

**Nextcloud server version:** (see your admin page)
19.0.3

**Server OS** (Ubuntu server is default)
Ubuntu 18.04

**How did you install the VM?** (Scripted install from master **OR** Released version)
Bought device in hansson it shop.

## Network
**Do you use DHCP?**
no.

**Is port 80 and/or 443 open?**
yes

<img width="438" alt="Bildschirmfoto 2020-10-03 um 16 28 42" src="https://user-images.githubusercontent.com/2069590/94994077-a61b3900-0595-11eb-889c-8940500f87b4.png">

rakekniven:
Forgot to mention that NC devs confirmed that minimum php version for NC20 is 7.2

szaimen:
yes, I saw that too. But NC21 will only work with PHP 7.3 and above.

rakekniven:
You are right. But for now I am unable to upgrade caused by wrong version definitions :-)

rakekniven:
Can you fix this? Thank you.

enoch85:
Hmm, at least the check works. :D

I'll have a look soon, if not @szaimen fixed it already.

rakekniven:
The fix needs to get merged

rakekniven:
Update works pretty well.
Only issue for me is that new dashboard cannot be loaded.
Nearly whitescreen is shown :-(
Will dig into it.

PR #1516, opened 2020-10-03
fix the update to NC20 issue

szaimen:
fix https://github.com/nextcloud/vm/issues/1513


rakekniven (approved):
For NC20 minimum php version is 7.2
For NC21 it will be 7.3

enoch85:
Looks good @rakekniven?

rakekniven:
Dialogs are still the same.
Do I need to do something else other than running '/var/scripts/update.sh' again to get the change?

enoch85:
> Do I need to do something else other than running '/var/scripts/update.sh' again to get the change?

Nope, once this is merged, wait 10 minutes for Github cache and then it should work automatically. :+1:

PR #1523, opened 2020-10-04
make the install script vagrant compatible

szaimen:


enoch85:
One more menu in the installer... :sleepy: We want as few menus as possible in part one.

I will think about this.

szaimen:
okay

szaimen:
@enoch85 if you want as less questions as possible, why not bringing ask_yes_or_no back only for the install production script and just run it always this way:
`yes | bash nextcloud_install_production.sh`

enoch85:
> @enoch85 if you want as less questions as possible, why not bringing ask_yes_or_no back only for the install production script and just run it always this way:
yes | bash nextcloud_install_production.sh

Yes, beacuse with menus, vagrant installation breaks.

nextcloud_install_production.sh
```diff
@@ -17,9 +17,30 @@ fi
...
 true
 SCRIPT_NAME="Nextcloud Install Script"
+SCRIPT_EXPLAINER="Welcome to the Nextcloud VM installer.
+It simplifies the installation of Nextcloud and its integrations by a lot and has other great features.
+You can find the sourcecode and other advices here: https://github.com/nextcloud/vm"
 # shellcheck source=lib.sh
 source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
+# Show explainer
+any_key "$SCRIPT_EXPLAINER\nPress any key to continue."
```
szaimen, line 27: does any_key continue if running
`yes | bash nextcloud_install_production.sh`
?

nextcloud_install_production.sh
```diff
@@ -17,9 +17,30 @@ fi
...
+
+if [ "no" = $(ask_yes_or_no "Do you want to continue with this script?") ]
+then
+    exit 1
+fi
+
+# Let the user run an initial packet update
+if [ "no" = $(ask_yes_or_no "Do you want to run an initial packet update?") ]
```
szaimen, line 35: What would be a better question to not make running an initial packet update the default?

szaimen:
@enoch85 how does vagrant manage whiptail menu's? I am not sure how to skip this...

szaimen:
> Yes, beacuse with menus, vagrant installation breaks.

Done

enoch85:
It would be super nice if you could test that Vagrant headless install actually works. I've only got it confirmed by one user, and I haven't tested myself.

szaimen:
This feature was implemented with #1535

Issue #1522, opened 2020-10-03
Use lvm snapshots as restore system for the update script?


szaimen:
Hey, using lvm snapshots for the update script would make the restoring much easier by simply running
```
lvconvert --merge /dev/ubuntu-vg/SnapshotName
reboot
```
Together with ZFS restore we could simplify and automate the process by a lot if an update fails. (obviously we couldn't get rid of backing up the data to the backup directory for old instances)

The only thing is that this would obviously only work for new instances and we would need to retain free space which, to reach the normal 40GB for the system drive, would mean, that we had to make the system drive e.g. double the size like 80GB and then have the normal capacity of 40GB for data. (this should be the default configuration that is made by the ubuntu server installer)

Here is a good source:
https://www.tecmint.com/take-snapshot-of-logical-volume-and-restore-in-lvm/

for later:
```
check_free_space() {
if vgs &>/dev/null
then
    FREE_SPACE=$(vgs | grep ubuntu-vg | awk '{print $7}' | grep g | grep -oP "^[0-9]+\.[0-9]" | sed 's|\.||')
fi
if [ -z "$FREE_SPACE" ]
then
    FREE_SPACE=0
fi
}
if [ "$FREE_SPACE" -ge 25 ]
then
    # TODO: check if a snapshot with the same name already exists. Exit if yes and suggest manually deleting or merging it.
    # TODO: create the snapshot
    # TODO: check if it gets created successfully
elif [ "$FREE_SPACE" -eq 0 ]
then
    # TODO: do the normal backup
else
   # TODO: report a fail
fi
```

enoch85:
In that case we would only use 50% of the capacity of the disk, as that's the new default in Ubuntu.

That would mean 20 GB as default space for:
- DB
- Logs
- And eeeverything else.

When it comes to expanding LVM, should we script that and put it in server configuration instead? And if we do, then snapsots would stop functioning (as you say).

enoch85:
I'm not sure we should increase storage to 80 GB.Have to think about that.

szaimen:
If I think about it, 10GB of free space should already be plenty enough for snapshots. If we want to make it as small as possible, 1GB would most likely enough, too.

szaimen:
I'll go with 2.5GB. It sounds like a good size, imo.

PR #1525, opened 2020-10-04
allow lvm snapshots

szaimen:
I made it possible to choose if you want to use it during the install script. It is not the default so we could test this from master, if we wanted because it will not effect people who have not setup lvm snapshots.

fix https://github.com/nextcloud/vm/issues/1522

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -1427,7 +1427,7 @@ fi
 
 # Check for free space on the ubuntu-vg
 check_free_space() {
-    if ! vgs &>/dev/null
+    if vgs &>/dev/null
     then
         FREE_SPACE=$(vgs | grep ubuntu-vg | awk '{print $7}' | grep g | grep -oP "[0-9]+\.[0-9]" | sed 's|\.||')
     fi
@@ -1442,10 +1442,8 @@ does_snapshot_exist() {
     local SNAPSHOTS
     local snapshot
     if lvs &>/dev/null
-    then
-        return 1
+        SNAPSHOTS="$(lvs | grep ubuntu-vg | awk '{print $1}' | grep -v ubuntu-lv)"
     fi
-    SNAPSHOTS="$(lvs | grep ubuntu-vg | awk '{print $1}' | grep -v ubuntu-lv)"
     if [ -z "$SNAPSHOTS" ]
     then
         return 1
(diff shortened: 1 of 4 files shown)


szaimen:
We should discuss whether 2.5GB will be enough. 

I made a test:
I run the install script. it takes around 1.3GB of space from the snapshot.
Afterwards I ran the startup script and installed fail2ban, Adminer, Netdata, BPytop, Fulltextsearch, Previewgenerator and Webmin. I ran it to the end incl. the update script and it took 2.7GB of space from the snapshot. So to make sure that you are able to use the snapshot as restore point during the startup script, we should most likely make it bigger than 2.5GB. I guess 5GB should be more than sufficient.

What do you think about this?

szaimen:
btw: after running the ubuntu server installer, it has this config:
image

szaimen:
> We should discuss whether 2.5GB will be enough.
> 
> I made a test:
> I run the install script. it takes around 1.3GB of space from the snapshot.
> Afterwards I ran the startup script and installed fail2ban, Adminer, Netdata, BPytop, Fulltextsearch, Previewgenerator and Webmin. I ran it to the end incl. the update script and it took 2.7GB of space from the snapshot. So to make sure that you are able to use the snapshot as restore point during the startup script, we should most likely make it bigger than 2.5GB. I guess 5GB should be more than sufficient.
> 
> What do you think about this?

I've changed it to 5GB. It seems to make more sense.

szaimen:
@enoch85 since this PR is relatively short: what are you missing to merge this? I've tested this already successfully. It will not effect instances that don't use snapshots (no instances that haven't run through this new option during the install script)

enoch85:
> since this PR is relatively short: what are you missing to merge this? I've tested this already successfully. It will not effect instances that don't use snapshots (no instances that haven't run through this new option during the install script)

I'm not sure we want this.

enoch85:
It also touches all the most important scripts...

szaimen:
Am I right that I cannot convince you that this PR makes sense by saying that I've made sure that this won't effect any existing install? 
It would btw also don't effect any install starting after this PR that hasn't consciously opted in during the install script...

enoch85:
It should say, I'm not sure since I didn't test.

Now I've tested and as it's not default, and not executed if not chosen during the install script, we can merge this.

PR #1526, opened 2020-10-04
warn that apps could be incompatible before upgrading to a major version

szaimen:
as it was suggested here:
https://help.nextcloud.com/t/nc-20-too-many-incompatible-apps-what-to-do/93869/7

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -431,6 +431,19 @@ msg_box "$NCREPO seems to be down, or temporarily not reachable. Please try agai
     exit 1
 fi
 
+if [[ "${CURRENTVERSION%%.*}" -lt "${NCVERSION%%.*}" ]]
+then
+    msg_box "By continuing, you will upgrade to the next major Nextcloud version ${NCVERSION%%.*}.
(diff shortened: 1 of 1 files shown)


enoch85:
No, better notify the user of how to re-enable untested apps, at their own risk with `notify_admin_gui` or `send_mail`since updates can be automated and that menu won't show up.

Also, we can't use that function in the update script as it would break it if running silent.

Right now we have like 1000 popups (`msg_box`) and I'm starting to think it's too much already.

szaimen:
> No, better notify the user of how to re-enable untested apps, at their own risk with `notify_admin_gui` or `send_mail`since updates can be automated and that menu won't show up.
> 
> Also, we can't use that function in the update script as it would break it if running silent.

But major updates are not automated.
At least not by us?

enoch85:
> But major updates are not automated.
At least not by us?

Yes, we didn't add the minor switch until after a while, so there may still be full upgrades taking place.

Also, the one in the issue was experiencing that from what I understand.

szaimen:
Okay, didn't know that the minor switch was just added after.

No, as far as I understand, he updated manually via cli.

PR #1533, opened 2020-10-06
Add two options to the not-supported folder

szaimen:
If I remember correclty, you said that it is okay if I add options to the not-supported folder :)


szaimen:
hm... It seems like Travis doesn't want to approve :D

PR #1535, opened 2020-10-07
Allow the complete automatic provisioning of default options during the install script

szaimen:
As the title says

This PR fixes #1523

### DIFF
--- a/vagrant/install.sh
+++ b/vagrant/install.sh
@@ -4,10 +4,10 @@
 true
 SCRIPT_NAME="Install NcVM with Vagrant"
 # shellcheck source=lib.sh
-source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
-install_if_not wget
-check_command wget https://raw.githubusercontent.com/nextcloud/vm/master/nextcloud_install_production.sh
+check_command git clone https://github.com/nextcloud/vm.git
 
-check_command sudo bash nextcloud_install_production.sh --provisioning
+cd vm || exit && print_text_in_color "$IRed" "Could not cd into the 'vm' folder."
 
+sudo bash nextcloud_install_production.sh --provisioning
--- a/vagrant/install.sh
+++ b/vagrant/install.sh
@@ -4,11 +4,10 @@
 true
 SCRIPT_NAME="Install NcVM with Vagrant"
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
-check_command git clone https://github.com/nextcloud/vm.git
+install_if_not wget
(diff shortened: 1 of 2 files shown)


szaimen:
What looks better to you:
`PROVISIONING=1 bash nextcloud_install_production.sh`
or
`bash nextcloud_install_production.sh --provisioning` ?
Both is possible...

szaimen:
We should probably change that command here too: https://github.com/nextcloud/vm/blob/a4eb6a6ac44059c5a800d0b6f103ac15d1c6c1f2/vagrant/install.sh#L13

enoch85:
So this removes all questions if `--provisioning` is set?

enoch85:
> bash nextcloud_install_production.sh --provisioning ?
Both is possible...

:+1:

szaimen:
> So this removes all questions if `--provisioning` is set?

yes 👍

szaimen:
I've added it to the vagrant file. Also I added a short `-p` flag to do the same.

btw: I've already tested this successfuly :D

The only remaining question is this:
> BTW: I think the cloning is unnecessary (we only need the install_script, no?) :
> Or would wget and curl not work on vagrant?
> https://github.com/nextcloud/vm/blob/a4eb6a6ac44059c5a800d0b6f103ac15d1c6c1f2/vagrant/install.sh#L9

enoch85:
Is this done?

szaimen:
Yes, I only need an answer for this

> BTW: I think the cloning is unnecessary (we only need the install_script, no?) :
> Or would wget and curl not work on vagrant?
> [vm/vagrant/install.sh](https://github.com/nextcloud/vm/blob/a4eb6a6ac44059c5a800d0b6f103ac15d1c6c1f2/vagrant/install.sh#L9)
> 
> Line 9 in [a4eb6a6](/nextcloud/vm/commit/a4eb6a6ac44059c5a800d0b6f103ac15d1c6c1f2)

szaimen:
@enoch85 I think this is ready now.

vagrant/install.sh
```diff
@@ -4,11 +4,10 @@
...
-source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
-check_command git clone https://github.com/nextcloud/vm.git
+install_if_not wget
+check_command wget https://raw.githubusercontent.com/nextcloud/vm/master/nextcloud_install_production.sh
 
-cd vm || exit && print_text_in_color "$IRed" "Could not cd into the 'vm' folder."
-
-yes no | sudo bash nextcloud_install_production.sh
```
enoch85, line 13: The vm folder needs to be local. Hence the code for that.

szaimen: I doubt that. We have nothing implemented to use the local scripts.
I am pretty sure that it just runs the script like everywhere else. (Loads the scripts during nextcloud_iinstall_production.sh directly from github)

enoch85: Well, have you actually tested it? 

I know for a fact the the current implementation works, so of you don't want to test the new one, please revert. :)

szaimen: Okay, I'll do 👍

szaimen: done

szaimen:
I think this is ready

PR #1536, opened 2020-10-07
smtp-mail: don't remove the logfile if something fails

szaimen:
as suggested in https://help.nextcloud.com/t/nexcloud-vm-smtp-setup-does-not-work/94306/2

### DIFF
--- a/addons/smtp-mail.sh
+++ b/addons/smtp-mail.sh
@@ -207,7 +207,9 @@ chmod 600 /etc/msmtprc
 # Create logs
 rm -f $VMLOGS/mail_msmtp.log
 sudo touch $VMLOGS/mail_msmtp.log
-sudo chmod 666 $VMLOGS/mail_msmtp.log # Even this doesn't work, strange
+chown msmtp:msmtp $VMLOGS/mail_msmtp.log
+sudo chmod 644 $VMLOGS/mail_msmtp.log
+sed -i "s| /var/log/msmtp | $VMLOGS/mail_msmtp.log |" /etc/apparmor.d/usr.bin.msmtp
 
 # Create aliases
 cat << ALIASES_CONF > /etc/aliases
@@ -236,7 +238,7 @@ $(grep -v password /etc/msmtprc)
 Best regards
 The NcVM team
 https://nextcloudvm.com" \
-| mail -s "Test email from your NcVM" "$RECIPIENT" > $VMLOGS/mail_msmtp.log 2>&1 # But this works
+| mail -s "Test email from your NcVM" "$RECIPIENT"
 then
     # Fail message
     msg_box "It seems like something has failed.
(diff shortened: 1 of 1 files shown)


addons/smtp-mail.sh
```diff
@@ -205,7 +205,6 @@ fi
 chmod 600 /etc/msmtprc
 
 # Create logs
-# TODO: not working due to permissions error
```
szaimen, line 208: Is this still true?

enoch85: When you tested it and it works as expected then it's true. :)

szaimen: still permission error, but I rewrite it manually to the logfile. So this is kind of resolved.

addons/smtp-mail.sh
```diff
@@ -238,19 +236,19 @@ $(grep -v password /etc/msmtprc)
 Best regards
 The NcVM team
 https://nextcloudvm.com" \
-| mail -s "Test email from your NcVM" "$RECIPIENT" &>/dev/null
+| mail -s "Test email from your NcVM" "$RECIPIENT" > $VMLOGS/mail_msmtp.log 2>&1 # But this works
```
szaimen, line 239: this way we rewrite any output to the log file

enoch85: Aah, forgot about that one. Nice catch!

Still would be nice to get msmtp to log everything to that log.

szaimen: The strange thing is that even chmod 666 doesn't work so I really don't know how to fix this.

szaimen: if you don't know either, I would suggest merging this because debugging will be easier for most users (at least for the setup) and solving the permission issue later.

szaimen: This is most likely the problem:
https://askubuntu.com/a/1191200

enoch85: I saw that before, but never tried it. Hence the code in the script:

```
groups msmtp
sudo touch /var/log/msmtp
sudo chown msmtp:msmtp /var/log/msmtp
sudo chmod 660 /var/log/msmtp
```

Please try and see if it works.

Thanks!

szaimen: This works now

szaimen:
@enoch85 this logs to the file now.

addons/smtp-mail.sh
```diff
@@ -205,11 +205,9 @@ fi
...
 
 # Create logs
-# TODO: not working due to permissions error
-rm -f $VMLOGS/mail_msmtp.log
-sudo touch $VMLOGS/mail_msmtp.log
-sudo chown msmtp:msmtp $VMLOGS/mail_msmtp.log
-sudo chmod 0644 $VMLOGS/mail_msmtp.log
+rm -f /var/log/msmtp
+touch /var/log/msmtp
+chmod 666 /var/log/msmtp
```
szaimen, line 210: I couldn't find any other solution
```
chown msmtp:msmtp logfile
chmod 600 logfile
```
Doesnt work

enoch85:
I saw the change, is this working now?

szaimen:
Yes, the log fileworks now but as I sadi I had to chmod 666

PR #1539, opened 2020-10-08
Add Midnight Commander as option to additional_apps

szaimen:
as the title says


szaimen:
We could think about adding this theme:
https://gist.github.com/holms/5005629

szaimen:
I think this is ready

PR #1545, opened 2020-10-08
Add deSEC to ddclient options

### LINKED ISSUE #1541: deSEC ddclient
Please add support deSEC. It's free DNS hosting service. 

https://desec.io/

szaimen:
fix #1541

### DIFF
--- a/network/ddclient-configuration.sh
+++ b/network/ddclient-configuration.sh
@@ -23,17 +23,25 @@ You have to setup an account before you can start.\n\nPlease choose your DynDNS-
 $MENU_GUIDE\n\n$RUN_LATER_GUIDE
 If your DDNS provider isn't already supported, please open a new issue here:\n$ISSUES" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Cloudflare" "(cloudflare.com)" \
+"deSEC" "(desec.io)" \
 "Strato" "(strato.de)" 3>&1 1>&2 2>&3)
 
 case "$choice" in
     "Cloudflare")
         PROVIDER="Cloudflare"
-        INSTRUCTIONS="Register an email address for your domain and get an Cloudflare API-key"
+        INSTRUCTIONS="register an email address for your domain and get an Cloudflare API-key"
         GUIDE="https://www.techandme.se/setup-multiple-accounts-with-ddclient-and-cloudflare/"
         PROTOCOL="cloudflare"
         SERVER="www.cloudflare.com"
         USE_SSL="yes"
     ;;
+    "deSEC")
+        PROVIDER="deSEC"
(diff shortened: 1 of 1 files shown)


szaimen:
@XSR1 could you please test if it works?
I don't have an account there and don't want to register one.
Please tell if you don't know how to run this script.

BTW: do you maybe have an englisch guide how to register a ddns account and get all needed credentials for desec?
I only found this: https://www.os-privacy.de/dyndns-desec-3181/

XSR1:
I have tested and its working great. It successfully updates A record. I use AAAA records and I my configs are posted:

https://github.com/nextcloud/vm/issues/1541

Not a biggie. Nice Work :)

szaimen:
@XSR1 Thanks for testing!
Since you seem to use both IPv4 and IPv6, is it posible to add both configs to the same ddclient config file?

I found this: 
https://github.com/ddclient/ddclient/issues/75
So it seems like this isn't supported?

XSR1:
As far as I know, is not currently possible to use both configs at same time. 
It deletes other record, example if i update A record using ddlient it deletes AAAA record and vice versa.

By the way this deSEC if much faster (Low-latency) than any other DNS hosting service, like duckdns, no-ip..
Also it works perfectly Let's Encrypt certs and I use this DNS service to point my ip chances with Cloudflare. 
So I have two domains with ssl one for family (deSEC) and other (floudlare) for other users.

szaimen:
@XSR1 deSEC sounds great 👍

If this limitation with ddclient exists, are you fine if we merge this PR?
Because as you said, it works at least for ipv4...

XSR1:
Yes this is fine. Thank you

szaimen:
@enoch85 please merge if this looks fine to you

network/ddclient-configuration.sh
```diff
@@ -23,17 +23,26 @@ You have to setup an account before you can start.\n\nPlease choose your DynDNS-
...
+        INSTRUCTIONS="register an email address for your domain and get an Cloudflare API-key"
         GUIDE="https://www.techandme.se/setup-multiple-accounts-with-ddclient-and-cloudflare/"
         PROTOCOL="cloudflare"
         SERVER="www.cloudflare.com"
         USE_SSL="yes"
     ;;
+    "deSEC")
+        PROVIDER="deSEC"
+        INSTRUCTIONS="get a DDNS account with password"
+        GUIDE="https://www.os-privacy.de/dyndns-desec-3181/"
```
enoch85, line 41: Please ad this instead: https://desec.io/#!

enoch85: Or maybe we should actually create the account? https://desec.readthedocs.io/en/latest/auth/account.html#register-account

PR #1549, opened 2020-10-10
lvm snapshots - make snapshot failures a bit more informative

szaimen:
It could only happen rarely that this appears and also only if you have opted in during the install script as I said before.
It just makes it a bit more informative


szaimen:
This is also done

nextcloud_update.sh
```diff
@@ -53,21 +57,37 @@ then
     then
         if ! lvremove /dev/ubuntu-vg/NcVM-snapshot -y
         then
+            nextcloud_occ maintenance:mode --off
+            if [ -n "$DOCKER" ]
+            then
+                check_command systemctl start docker
```
enoch85, line 63: Btw, we have `start_if_stopped docker`, just remembered now....

Please look for all start occurrences and change accordingly.

Thanks!

PR #1550, opened 2020-10-10
notify users about started Nextcloud update

szaimen:
As the title says.

I also removed the admin notification about updated docker containers since that is a bit too much imo.
(we just log it now)


nextcloud_update.sh
```diff
@@ -583,6 +583,9 @@ then
     print_text_in_color "$ICyan" "$BACKUP/apps/ exists"
     echo 
     print_text_in_color "$IGreen" "All files are backed up."
+    send_mail \
```
enoch85, line 586: We should use the notify function here. If en email server is setup in the GUI, users will get the notification on ALL devices, not only email.

Also, we can't be sure that email is setup on all hosts.

szaimen: I thought this too but if you get a nextcloud notification shortly before the update starts and then you are not able to read it because the update already started, I imagine that this could make some users try to restart the vm.

Btw: send_mail only works if you have setup mail and for all other it just doesn't send any. So it doesn't change anything for anybody who hasn't setup the smtp relay which is okay imo.

PR #1555, opened 2020-10-12
Don't clear the cli history

szaimen:
I think this makes sense to keep the history for better debugging.
If I remember correclty, one guy wanted something like this.
E.g. during the startup script you will see anything which the person has done if we don't clear the history.
@enoch85 What do you think about this?


enoch85:
Sure, let's merge it.

I was using it before to get a cleaner experience, but now when we use `msg_box` so much, I don't think clear is needed.

Issue #1556, opened 2020-10-14
After the installation of f2b initiates.


EasyCompZLD:
![Schermafbeelding 2020-10-14 om 05 04 52](https://user-images.githubusercontent.com/67462665/95938824-f89cf680-0dda-11eb-8e30-42aef1f2f763.png)

Got nothing more to say

![loadcat](https://user-images.githubusercontent.com/67462665/95938839-0783a900-0ddb-11eb-8660-f92a5ef8bdb2.gif)

szaimen:
@EasyCompZLD Please try again. This time it should work.

EasyCompZLD:
It's updated?

szaimen:
> It's updated?

Yes it is.

EasyCompZLD:
Oké I'll try it later. How fast can you respond? ;-p time now 13:01 UTC +1 ams

PR #1557, opened 2020-10-14
Update fetch_lib to update the library during the startup script, too

szaimen:
This is the best way to solve #1556

But since fetch_lib gets downloaded during the install script, we have no other choice but either don't use fetch_lib for all scripts (only use the online version) that use the new nextcloud_occ commands or revert the change to nextcloud_occ for this release.


szaimen:
I am wrong. There is a good solution to this

PR #1559, opened 2020-10-14
fix two small issues

szaimen:
This fixes two minor details.
1. it makes geoblock executable
2. it inserts a space


enoch85:
Note to self, don't forget to test the actual script before merge.

szaimen:
> Note to self, don't forget to test the actual script before merge.

Yes that makes always sense but I think those are not big problems and fast to fix as you can see ;-)

Issue #1573, opened 2020-10-15
Onlyoffice URL error


ezraholm50:
## Steps to reproduce
1. Use menu.sh
2. Install integrated onlyoffice
3. view url stated in CLI/onlyoffice admin settings

## Expected behaviour
should display domain.tld/index.php/apps/documentserver_community

## Actual behaviour
Displays domain.tld//index.php/apps/documentserver_community

## Server configuration

**Nextcloud server version:**
20.0.0

**Server OS** 
Ubuntu server 20.04

**How did you install the VM?**
Scripted

## Network
**Do you use DHCP?**
No 

**Is port 80 and/or 443 open?**
Yes 

## Logs / Screenshots
![DeepinScreenshot_select-area_20201015213056](https://user-images.githubusercontent.com/13510720/96177386-1b8bef80-0f2e-11eb-8808-4949823b5b29.png)

#### Log file (/var/log/nextcloud/nextcloud.log)
```
[...]

PR #1574, opened 2020-10-15
Update onlyoffice_integrated.sh

ezraholm50:
fix https://github.com/nextcloud/vm/issues/1573


ezraholm50:
You might turn it around and change the `echo cli overwrite url` to exclude an end `/`

enoch85:
Sorry, saw this now. I committed to master instead. 

Thanks for the fix though!

PR #1579, opened 2020-10-16
smtpmail: Allow to not reset all configs and packets at the end of the script

szaimen:
As the title says


addons/smtp-mail.sh
```diff
@@ -218,7 +218,7 @@ ALIASES_CONF
 
 # Define the mail-program
 cat << DEFINE_MAIL > /etc/mail.rc
-set sendmail="/usr/bin/msmtp"
```
szaimen, line 221: This doesn't work for me so I restored the old config.

szaimen: I have an idea how to test both.

szaimen: done

szaimen:
Now i've made it possible to test both variants before failing. 
It was already successfully tested by myself.
I hope this is okay now :)

enoch85:
OK, much better!

PR #1581, opened 2020-10-17
improve remotedesktop

szaimen:
As the tilte says.

Those changes were already successfully tested on a new test-vm.
It works now as intended.


szaimen:
It is ready now :)

PR #1582, opened 2020-10-17
make bpytop a script

enoch85:


enoch85:
@szaimen Please check this.I'm not that used to the new removal_popup function. Does this look correct?

apps/bpytop.sh
```diff
@@ -0,0 +1,73 @@
...
+    # Removal
+    sed -i "s|.*bpytop'||g" /home/"$UNIXUSER"/.bash_aliases
+    sed -i "s|.*bpytop'||g" /root/.bash_aliases
+    snap remove bpytop
+    # Show successful uninstall if applicable
+    removal_popup "$SCRIPT_NAME"
+fi
+
+# Install it!
+if yesno_box_yes "Do you want to install BPYTOP?" "$SUBTITLE"
```
enoch85, line 41: maybe this is not needed?

szaimen: Yes, this is not needed. It is already handled by the install_popup

apps/bpytop.sh
```diff
@@ -0,0 +1,73 @@
...
+# Check for errors + debug code and abort if something isn't right
+# 1 = ON
+# 0 = OFF
+DEBUG=0
+debug_mode
+
+# Must be root
+root_check
+
+# Needed for snaps to run
```
szaimen, line 21: Her is missing a `msg_box "$SCRIPT_EXPLAINER"`
otherwise except the other issue it looks fine 👍

szaimen:
You can remove the SUBTITLE everywhere since this is its own script now.

apps/bpytop.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
+
+# shellcheck disable=2034,2059
+true
+SCRIPT_NAME="BPYTOP"
```
enoch85, line 7: @szaimen Isn't this enough already?

szaimen: Yes it is. Because of this you can remove the SUBTITLE variable now from the whole script.

apps/bpytop.sh
enoch85: See comment above.

apps/bpytop.sh
```diff
@@ -0,0 +1,70 @@
...
+SCRIPT_EXPLAINER="BPYTOP is an amazing alternative to ressource-monitor software like top or htop."
```
enoch85, line 8: I mean this, SCRIPT_EXPLAINER is already there.

szaimen: Yes, but I changed the implementation recently so that an additional `msg_box "$SCRIPT_EXPLAINER"` is needed at the beginning of any script that has one.

enoch85: That seems unnecessary, wouldn't it be possible to only provide the variable at the top?

enoch85: Why did you change it? :thinking:

szaimen: Because it was bound to the explainer_popup which I removed. If you want to reintroduce it, we could bind it to install_popup and reinstall_remove_menu instead.

szaimen: > If you want to reintroduce it, we could bind it to install_popup and reinstall_remove_menu instead.

But there are some problems with that. Because of that I just did it this way.

szaimen: We could also reintroduce the explainer_popup function which could only show a msg_box with description if SCRIPT_EXPLAINER is set.
But I'd rather add a msg_box "$SCRIPT_EXPLAINER"

szaimen: Here is an example from the Remotedesktop script:
https://github.com/nextcloud/vm/blob/f0f21aabcd4ee7049febdd82ddcc074c2245b329/not-supported/remotedesktop.sh#L23-L24

enoch85: Well, it seems like double code.

First you set `SCRIPT_EXPLAINER` then you execute `SCRIPT_EXPLAINER` with `msg_box` - instead of just using the content in `SCRIPT_EXPLAINER``like plain text in the `msg_box` like:

Correct:
msg_box "This is the text of the `SCRIPT_EXPLAINER`"

Seems unessesary:
SCRIPT_EXPLAINER="This is the text of the `SCRIPT_EXPLAINER`"
msg_box "`$SCRIPT_EXPLAINER`"

szaimen: I don't think so because it makes the function of the script directly visible to the user that reads the script because it is on the top and labeled with script_explainer.
This wouldn't be as clear if we'd just use msg_box "text" anywhere down there.

apps/bpytop.sh
```diff
@@ -0,0 +1,72 @@
...
+    msg_box "BPYTOP is now installed! Check out the amazing stats by runnning 'bpytop' from your CLI.
+You can check out their Gihub repo here: https://github.com/aristocratos/bpytop/blob/master/README.md"
+    # Ask for aliases
+    if [ -f /root/.bash_aliases ]
+    then
+        if yesno_box_yes "You you want to add an alias for bpytop to replace both htop and top?"
+        then
+            echo "alias top='bpytop'" >> /root/.bash_aliases
+            echo "alias htop='bpytop'" >> /root/.bash_aliases
+            if [ -d /home/"$UNIXUSER" ]
```
szaimen, line 61: What if UNIXUSER is empty?

enoch85: Then it's not set.

szaimen: @enoch85 Yes but this check would return true because the `/home/ ` directory exists, no?

szaimen:
@enoch85 I've tested this and the alias only works after a reboot...

enoch85:
> I've tested this and the alias only works after a reboot...

Or if you login our out.

Issue #1563, opened 2020-10-14
SMTP Relay with msmtp does not work


baditaflorin:
## Steps to reproduce
1. Try to use google documentation https://support.google.com/a/answer/176600?hl=en for 
```
SMTP Relay URL=smtp-relay.gmail.com    
   │ Encryption=NO-ENCRYPTION       
         │ SMTP Port=25           
   │ SMTP Username=user@domain.com    
  │ SMTP Password=****************                                                            
Recipient=myemailthatwillgettheemail@gmail.com   
                 ```                                                             
3.

## Expected behaviour
Tell us what should happen
Allow you to set SMTP and receive alerts, emails. 

## Actual behaviour
Tell us what happens instead
```
root@nextcloud:~# cat /var/log/msmtp
mail: /usr/bin/msmtp -t: No such file or directory
Can't send mail: sendmail process failed with error code 1
```
## Server configuration

**Nextcloud server version:** (see your admin page)
latest
**Server OS** (Ubuntu server is default)
latest
**How did you install the VM?** (Scripted install from master **OR** Released version)
https://shop.hanssonit.se/product/nextcloud-home-sme-server-4tb-2/
## Network
**Do you use DHCP?**

**Is port 80 and/or 443 open?**
Cannot open port 80 or 443, ISP blocks all ports under 1000
 
#### Log file (/var/log/nextcloud/nextcloud.log)
```
empty file```

#### Installation log (command output)
```
```

#### Screenshoots
<!--
Please use http://imgur.com/ for screenshots. Thanks!
-->

szaimen:
@baditaflorin have you used an app password for that or just your regular password?
https://support.google.com/accounts/answer/185833?hl=en

baditaflorin:
I tried simple user/pass with an account that I created in google suite for our domain.

szaimen:
Okay then please try again with an app password for that account.

Also please test if the `Use the Gmail SMTP server` method works for you

szaimen:
This is what I would suggest you should try:
![Screenshot_20201015-030230_Brave](https://user-images.githubusercontent.com/42591237/96061346-4412db80-0e93-11eb-9a79-e28d4330aa13.jpg)

szaimen:
Please report if https://github.com/nextcloud/vm/issues/1563#issuecomment-708809167 works for you, thanks!

baditaflorin:
Same, getting error code 1 and exits. And the log does not provide anything useful. 
[...]

PR #1590, opened 2020-10-18
fix some issues with smtp-mail

szaimen:
While debugging #1563 I discovered this issue.

As I said this was my guide:
https://decatec.de/linux/linux-einfach-e-mails-versenden-mit-msmtp/
And there you can see that those values (account and account default) definitely should be the mail_username.
I've tested this successfully with my account.

### DIFF
--- a/addons/smtp-mail.sh
+++ b/addons/smtp-mail.sh
@@ -162,12 +162,12 @@ tls_trust_file  /etc/ssl/certs/ca-certificates.crt
 # logfile         /var/log/msmtp
 
 # Account to send emails
-account         $MAIL_SERVER
+account         $MAIL_USERNAME
 host            $MAIL_SERVER
 port            $SMTP_PORT
 from            $MAIL_USERNAME
 
-account default : $MAIL_SERVER
+account default : $MAIL_USERNAME
 
 ### DO NOT REMOVE THIS LINE (it's used in one of the functions in on the Nextcloud Server)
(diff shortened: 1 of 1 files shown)


szaimen:
oh, wrong label. This is ready for review :)

enoch85:
What issue does this fix? My german isn't perfect. :)

enoch85:
With passwordless auth there is no account...

szaimen:
I am refering ro those lines (they are english)

enoch85:
Yes, I saw that. But what does it actually fix? Is it an issue to have the server instead?

Also: 

> With passwordless auth there is no account...

szaimen:
> Yes, I saw that. But what does it actually fix? Is it an issue to have the server instead?

I had some issues with this while testing (it said something like account not defined or so)
And it is recommended as you can see in the screenshot.

> With passwordless auth there is no account...

Do you mean that "$MAIL_USERNAME" is empty in this case?
I don't think so:
https://github.com/nextcloud/vm/blob/1d92bb6f1b075ccb83e53961f3fe0304dbb54224/addons/smtp-mail.sh#L147-L150

szaimen:
What about changing it to mail_username only for configurations where auth is on?
What do you think?

enoch85:
Nvm, let's merge this.

enoch85:
Once you start merging you can stop.

external-content duckduckgo com

Issue #1589, opened 2020-10-17
Move previewgenerator and fail2ban-menu over to nextcloud_configuration menu?


szaimen:
Hey, I would like to discuss whether it makes sense to move the previewgenerator and fail2ban menu over to the nextcloud_configuration menu since this is directly Nextcloud related and fits from the topic perfectly into this menu, imo.
All other Apps in the additional_apps menu are not that directly related to Nextcloud.
What do you think?

szaimen:
Maybe we could rename nextcloud_configuration to nextcloud_menu?

enoch85:
Naah, I think they're good at the current location.

Previewgenerator is an app, and fail2bansort of as well..

Maybe move Fail2ban to ADDONS, but not needed.

szaimen:
okay

szaimen:
Just one proposal

PR #1591, opened 2020-10-18
add Nextcloud to description in additional_apps

szaimen:
@enoch85 what do you think?
fix https://github.com/nextcloud/vm/issues/1589

### DIFF
--- a/apps/midnight-commander.sh
+++ b/apps/midnight-commander.sh
@@ -71,5 +71,5 @@ MC_INI
 done
 
 # Inform the user
-msg_box "The dark theme for Midnight Commander was successfully applied.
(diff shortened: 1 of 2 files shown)


enoch85:
Same thing again, only half menu is shown. Please have a look in this PR as well.

Thanks!

enoch85:
I don't think we need to extend the text, it will make the menu wider, and maybe don't fit all screens

szaimen:
I am not sure what you are experiencing here. This is how it looks like for me:
image

szaimen:
BTW: it fits in the CLI of VMplayer without problem. So I guess this should fit anywhere.
image

PR #1593, opened 2020-10-18
Update previewgenerator with new function for reinstalling

szaimen:
As the title says


enoch85:
Maybe something we should consider...? https://github.com/nextcloud/server/issues/5717

Possible to remove, but we need the instanceid.

PR #1595, opened 2020-10-18
update fulltextsearch with new function for reinstalling

szaimen:
As the title says

### DIFF
--- a/apps/fulltextsearch.sh
+++ b/apps/fulltextsearch.sh
@@ -45,7 +45,7 @@ else
     # Reset Full Text Search to be able to index again, and also remove the app to be able to install it again
     nextcloud_occ_no_check fulltextsearch:reset
     APPS=(fulltextsearch fulltextsearch_elasticsearch files_fulltextsearch)
-    for app in "${APSS[@]}"
+    for app in "${APPS[@]}"
     do
         if is_app_installed "$app"
         then
--- a/apps/fulltextsearch.sh
+++ b/apps/fulltextsearch.sh
@@ -31,7 +31,7 @@ msg_box "$SCRIPT_EXPLAINER"
 lowest_compatible_nc 18
 
 # Check if Full Text Search is already installed
-if ! does_this_docker_exist "$nc_fts"
+if ! does_this_docker_exist "$nc_fts" || ! is_app_installed fulltextsearch
 then
     # Ask for installing
     install_popup "$SCRIPT_NAME"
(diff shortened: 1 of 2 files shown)


lib.sh
```diff
@@ -282,6 +282,7 @@ input_box_flow() {
 }
 
 install_popup() {
+    msg_box "$SCRIPT_EXPLAINER"
```
szaimen, line 285: I am against this change. I think it is better to show it always at the beginning of the script no matter if you have it installed or not

apps/fulltextsearch.sh
```diff
@@ -6,6 +6,7 @@
 # shellcheck disable=2034,2059
 true
 SCRIPT_NAME="Full Text Search"
+export SCRIPT_EXPLAINER="Full Text Search provides Elasticsearch for Nextcloud, which makes it possible to search for text inside files."
```
szaimen, line 9: You don't need to export it

lib.sh
enoch85: if you already installed it, you probably know what it is...

apps/fulltextsearch.sh
enoch85: let's do it like this, then we can test.

Less code, better function. :partying_face:

szaimen, line 9: Really, you don't need to export it to use it in a function

lib.sh
```diff
@@ -299,6 +300,8 @@ install_popup() {
             exit 1
         fi
     fi
+    # SCRIPT_EXPLAINER is exported in each script, this unsets it.
+    unset SCRIPT_EXPLAINER
```
szaimen, line 304: unsetting the variable is not needed when you don't export it which is not needed either to use it in a function

apps/fulltextsearch.sh
enoch85: Since it's used in another script, best practice is to export it. 

We source the lib, but the app script itself isn't sourced, which means exporting a variable inside the app script is needed to be able to run it in the lib. 

Please prove me wrong.

szaimen: I am not sure how to prove you wrong by words. But I have tested that it works without export and unset.

szaimen: exporting and unsetting fells like a big hack which is not necessary!

enoch85:
Please remove all the msg_boxes in the other scripts, I need to go to bed now. 

Thanks! :rocket:

PR #1596, opened 2020-10-18
update fail2ban with new function for reinstalling

szaimen:
As the title says


enoch85:
OK, last one for today

PR #1597, opened 2020-10-18
update talk with new function for reinstalling

szaimen:
As the title says


enoch85:
I think we should remove this script and the talk menu. 

1. Rename talk_signaling.sh to talk.sh
2. Remove the menu for Talk

Reason? You can choose if you want signaling or not when installing Talk.

szaimen:
> I think we should remove this script and the talk menu.

Ok for me 👍

PR #1604, opened 2020-10-18
Remove msg_box "$SCRIPT_EXPLAINER" where install_popup exists

szaimen:
As the title says


szaimen:
@enoch85 I think this should be the next PR that gets reviewed because for some scripts there are now 2 msg_boxes with an explainer which is not really great ;)

enoch85:
Thanks!

PR #1610, opened 2020-10-19
fix some small bugs with talk.sh

szaimen:
As the title says

### DIFF
--- a/apps/talk.sh
+++ b/apps/talk.sh
@@ -53,8 +53,7 @@ else
         $VMLOGS/talk_apache_access.log \
         $VMLOGS/turnserver.log \
         /var/www/html/error
-            apt-get purge coturn -y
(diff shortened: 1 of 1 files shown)


apps/talk.sh
```diff
@@ -34,14 +34,10 @@ else
...
     # Removal
-    APPS=(spreed signaling_servers turn_servers stun_servers)
-    for app in "${APPS[@]}"
-    do
-        if is_app_installed "$app"
-        then
-            nextcloud_occ app:remove "$app"
-        fi
-    done
+    nextcloud_occ_no_check config:app:delete spreed stun_servers
```
enoch85, line 37: Why did you revert this?

enoch85: Aaah, it's config!

PR #1611, opened 2020-10-19
only check if the fulltextsearch docker container exists

szaimen:
I think this check is enough

### DIFF
--- a/apps/fulltextsearch.sh
+++ b/apps/fulltextsearch.sh
@@ -28,7 +28,7 @@ root_check
 lowest_compatible_nc 18
 
 # Check if Full Text Search is already installed
-if ! does_this_docker_exist "$nc_fts" || ! is_app_installed fulltextsearch
+if ! does_this_docker_exist "$nc_fts"
 then
     # Ask for installing
     install_popup "$SCRIPT_NAME"


enoch85:
No it's not enough, hence I added it.

szaimen:
@enoch85 why is it not enough?

szaimen:
BTW:
I think the check must be:
`if ! does_this_docker_exist "$nc_fts" && ! is_app_installed fulltextsearch`
no?

enoch85:
Nope. to be sure it's never been installed before we check if either is installed, not both.

szaimen:
Yes, problem is that this will also show the install popup if the fulltextsearch docker container is installed but the fulltextsearch app is not installed which will lead to an error during reinstallation because it cannot reinstall fulltextsearch if a container with this name already exists...
(This will not show the reinstall_remove_menu if the container exists but the fulltexsearch app is not installed what it should imo. Instead it will show the install_popup in this case which is wrong imo.)

enoch85:
> Yes, problem is that this will also show the install popup if the fulltextsearch docker container is installed but the fulltextsearch app is not installed which will lead to an error during reinstallation because it cannot reinstall fulltextsearch if a container with this name already exists...
(This will not show the reinstall_remove_menu if the container exists but the fulltexsearch app is not installed what it should imo. Instead it will show the install_popup in this case which is wrong imo.)
This is the reason why I wamted ro change rhe check.

Much better explanation! 👍 

But, this will check if either one of those are installed (not together), and do the reinstall if that's the case, and that's what we want, right?

szaimen:
> But, this will check if either one of those are installed (not together), and do the reinstall if that's the case, and that's what we want, right?

yes, we want that but the current check will show the install_popup and no the reinstall_remove_menu if one of those is not installed.
This is because it should be `if ! does_this_docker_exist "$nc_fts" && ! is_app_installed fulltextsearch`

PR #1612, opened 2020-10-19
improve fulltextsearch check

szaimen:
In my opinion the current check is wrong so I improved it.


enoch85:
Please see my previous comment.

PR #1622, opened 2020-10-24
rename the NcVM-snapshot-update name to NcVM-snapshot-pending

szaimen:
As the title says

As you know, this doesn't effect any current OVA.
(only if explicitely chosen during the nextcloud_install_production script and it is also not the default)

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -31,7 +31,7 @@ root_check
 is_process_running apt
 is_process_running dpkg
 
-if does_snapshot_exist "NcVM-snapshot-update"
+if does_snapshot_exist "NcVM-snapshot-pending"
 then
     msg_box "It seems like the last update was not successful.
 Cannot proceed because you would loose the last snapshot."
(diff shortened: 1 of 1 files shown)


enoch85:
This seems unfinished... Are you sure about this?

szaimen:
> Are you sure about this?

Yes, I am sure about this. 👍

szaimen:
> This seems unfinished..

What lets you think this way?

szaimen:
> > This seems unfinished..

I only renamed the snapshot name during the backup.
This doesn't change any functionality. I don't know why this should be unfinished?

enoch85:
What about people that already chose the snapshot option? They will still have the old name.

enoch85:
Please answer my question above :point_up_2: 

Thanks!

szaimen:
> What about people that already chose the snapshot option? They will still have the old name.

They only have the old temporary snapshot name if the update has failed. In this case they will be able to execute the update script again but no new snapshot will be created. Imo this should be fine.

PR #1623, opened 2020-10-24
update send_mail to feature return values

szaimen:
I need the return values for a certain usecase.


szaimen:
I've successfully tested that this doesn't effect notify_admin_gui, hence marking as ready for review.

enoch85:
What's the use case?

szaimen:
> What's the use case?

I want to find out if the smtp-relay is setup by calling the function.

PR #1630, opened 2020-10-27
fix the fail2ban removal for bitwarden_rs

szaimen:
This is how I would do it

### DIFF
--- a/apps/fail2ban.sh
+++ b/apps/fail2ban.sh
@@ -27,7 +27,7 @@ debug_mode
 root_check
 
 # Check if fail2ban is already installed
-if ! [ -f /etc/fail2ban/filter.d/nextcloud.conf ]
+if ! [ -f /etc/fail2ban/filter.d/nextcloud.conf ] || is_this_installed fail2ban
 then
     # Ask for installing
     install_popup "$SCRIPT_NAME"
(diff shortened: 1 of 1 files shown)


enoch85:
@szaimen Please have a look at m changes. Thanks! :+1:

apps/fail2ban.sh
```diff
@@ -27,24 +27,23 @@ debug_mode
 root_check
 
 # Check if fail2ban is already installed
-if ! [ -f /etc/fail2ban/filter.d/nextcloud.conf ]
+if ! [ -f /etc/fail2ban/filter.d/nextcloud.conf ] || is_this_installed fail2ban
```
szaimen, line 30: The correct check should be
`if ! [ -f /etc/fail2ban/filter.d/nextcloud.conf ] && ! is_this_installed fail2ban`

apps/fail2ban.sh
```diff
@@ -27,18 +27,28 @@ debug_mode
...
-    rm -Rf /etc/fail2ban/   
+    if ! does_this_docker_exist bitwarden_rs
+    then
+        print_text_in_color "$ICyan" "Unbanning all currently blocked IPs..."
+        fail2ban-client unban --all
+        apt purge fail2ban -y
+        rm -rf /etc/fail2ban
+    else
+        print_text_in_color "$ICyan" "Unbanning all currently blocked IPs..."
+        fail2ban-client unban --all
```
szaimen, line 46: I guess this would also unban all bitwarden_rs banned ip addresses but should most likely be fine...

szaimen:
> @szaimen Please have a look at m changes. Thanks! 👍

Looks good except the first check as commented

PR #1631, opened 2020-10-27
update previewgenerator

szaimen:
the preview:pre-generatore command doesn't allow usernames
But the preview:generate-all does.

also run the cronjob every 10 minutes which is recommended:
https://github.com/rullzer/previewgenerator#how-to-use-the-app

The crontab has to be added last because the generrate-all command has to run to the end before the pre-generate works.

I also made generate-all more verbose because it can take a very long time.

### DIFF
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -223,6 +223,11 @@ nextcloud_occ config:system:set preview_max_y --value="2048"
 nextcloud_occ config:system:set jpeg_quality --value="60"
 nextcloud_occ config:app:set preview jpeg_quality --value="60"
 
+# Add crontab for www-data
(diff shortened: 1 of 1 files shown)


enoch85:
Nice!

Please add a check to avoid double crontabs like this: https://github.com/nextcloud/vm/blob/master/lib.sh#L1646

szaimen:
> Please add a check to avoid double crontabs like this: https://github.com/nextcloud/vm/blob/master/lib.sh#L1646

Do we really need this?
https://github.com/nextcloud/vm/blob/9ca46708a5a24aee9912d99026ac98c3314b06cb/apps/previewgenerator.sh#L50

szaimen:
I just reverted the verbose option because it already shows the scanned folder during the scan. Showing every scanned file is not that important, I think.

PR #1632, opened 2020-10-27
20.0.1

enoch85:
@szaimen Please add the stuff you changed.

I will have a look soon. 

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ PR's are more than welcome. Happy Nextclouding!
 - Created geblock.sh in order to allow access from configured countries and/or continents
 - Made it more clear that a Nextcloud update started
 - Added DuckDNS to DDclient-configuration
-- Fixed a wrong OnlyOffice-URL
+- Fixed an incorrect OnlyOffice-URL
 - Improved the guidance how to control whiptails
 - Added some popups that explain the Additional Apps Menu and Server Configuration Menu during the startup script
 - Switched to TLS1.3 for new website-configurations on Ubuntu 20.04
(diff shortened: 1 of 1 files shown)


szaimen:
I think those should be all ;)

enoch85:
Perfect! Thanks @szaimen!

szaimen:
Wow, I am amazed!
Travis finally finished after 2 days 😅

szaimen:
@enoch85 When do you want to do the release?

TheNomad11:
Sorry for interrupting you, but I got your 20.0.1 update already last weekend? It did not show up under releases, that was weird, though!

enoch85:
> that was weird, though!

Haha, not that weird since I never announced it.

When I make the final release, then, and only then I publish it here on Github as well, in the docs, and all that. :)

TheNomad11:
> > that was weird, though!
> 
> Haha, not that weird since I never announced it.
> 
> When we make the final release, then, and only then I publish it here on Github as well, in the docs, and all that. :)

So in the automatic updates we get beta versions?

enoch85:
> So in the automatic updates we get beta versions?

No? I don't think I understand...

TheNomad11:
> > So in the automatic updates we get beta versions?
> 
> No? I don't think I understand...

Last weekend I got a NC-notification that 20.0.1 is ready, then I ran your update script and since then I have been using 20.0.1 although you have not released it officially and it does not seem to be stable? Everything seems to work fine, though

enoch85:
@TheNomad11 Naah, the released version here isn't downloaded by the update script. The update script updates the already existing Nextcloud and it's dependencies, it doesn't fetch the release from this repo.

A release in this term, means that we make 20.0.1 available for every new download. There are people that downloaded NC 10, and are now on NC 20 with the update script, but they don't have to get a new release from here every time. That's part of the beauty with the update script. :)

PR #1634, opened 2020-10-28
Fixed shellcheck errors for docker-migration script

JackGlobetrotter:


migrate/docker/migrate.sh
```diff
@@ -67,27 +67,27 @@ if (( $PSQLVERSION < $PSQLVERSION_DOCKER )); then
...
 echo "making new folders"
-mkdir $DIR 
+mkdir "$DIR"
 
 echo "Copying docker-compose file"
-cp docker-compose.yml $DIR
+cp docker-compose.yml "$DIR"
 
-cd $DIR
+cd "$DIR" || exit 0
```
enoch85, line 90: You could make it cleaner by using msg_box here, stating why it exists.

if ! cd "$DIR"
then
    msg_box "$DIR was not found, this script will exit. Please report this to $ISSUES.
    exit 1
fi

In general, try to be more verbose on why things happen so that the user knows what's going on instead of the script just exiting.

Thanks!

migrate/docker/migrate.sh
```diff
@@ -165,7 +168,7 @@ sed -i "$start s/.*/listen_addresses = '*'/" $PG_CFG_PATH
 echo "Patching Postgresql HBA file"
 echo "host all all all md5" >> db/pg_hba.conf
 
-chown -R www-data:docker *
+chown -R www-data:docker ./*
```
enoch85, line 171: Please use the full path. This has potential to fail.

JackGlobetrotter: Thanks for the feedback, will change asap.

enoch85:
Not tested, but I trust that you will maintain this @JackGlobetrotter 

Thanks for your efforts so far!

PR #1637, opened 2020-10-31
Fix an issue with clamav

szaimen:
@enoch85 btw: Currently clamav is in the not-supported menu, is there still the plan to make it accessible from the additionals apps menu?


enoch85:
Not sure, but maybe :)

PR #1643, opened 2020-11-01
Update geoblock.sh to get rid of jq

szaimen:
Also, this now uses the original by maxmind provided country codes which is better.


szaimen:
I've successfully tested ths, hence marking as ready for review

PR #1646, opened 2020-11-01
Add another flag to the install-script

szaimen:
As the title says
This is for what I need #1644 (So #1644 would need to get merged first)

@enoch85 What do you think? I really want this!


enoch85:
No, we will never add the not supported stiff into any of the released scripts.

enoch85:
Thanks for the effort though.

szaimen:
Too bad :/
I thought this was finally a solution you would be okay with ☹️

enoch85:
Sorry

PR #1648, opened 2020-11-01
Add allow-snapshots flag to the install-script

szaimen:
@enoch85 Would at least this be okay?
I would just like to have a different default setup for my needs...

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -811,6 +811,9 @@ restart_webserver
 if [ -n "$PROVISIONING" ]
 then
     choice="Calendar Contacts IssueTemplate PDFViewer Extract Text Mail Deck Group-Folders"
+elif [ -n "$ALLOW_SNAPSHOTS" ] 
+then
+    choice="Calendar Contacts PDFViewer Text Deck"
 else
     choice=$(whiptail --title "$TITLE - Install apps or software" --checklist \
 "Automatically configure and install selected apps or software
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -293,6 +293,9 @@ do
     if [ -n "$PROVISIONING" ]
     then
         choice="Quad9"
+    elif [ -n "$ALLOW_SNAPSHOTS" ] 
+    then
+        choice="Local"
     else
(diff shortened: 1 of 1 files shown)


szaimen:
Maybe a different name for the flag would be better? Do you have an idea?

nextcloud_install_production.sh
```diff
@@ -136,7 +141,7 @@ if [ -z "$PROVISIONING" ]
 then
     msg_box "Your current download repository is $REPO"
 fi
-if [ -n "$PROVISIONING" ] || yesno_box_yes "Do you want use http://archive.ubuntu.com as repository for this server?"
+if [ -n "$PROVISIONING" ] || yesno_box_no "Do you want use http://archive.ubuntu.com as repository for this server?"
```
szaimen, line 144: This change doesn't have really much todo with the new flag, I just think that since this is set by the provisioning variable automatically, it doesn't have to be the default during the setup anymore, since most likely you already have the best mirror available for you when also setting up ubuntu before.

enoch85:
Are you making your own VM of some kind, using this repo?

szaimen:
> Are you making your own VM of some kind, using this repo?

No

szaimen:
@enoch85 If you don't like this, too, just say it and I'll close it.

szaimen:
> > Are you making your own VM of some kind, using this repo?
> 
> No

This is only for my own usage to speed up the install script setup and always choosing the correct options.

szaimen:
> > > Are you making your own VM of some kind, using this repo?
> > 
> > 
> > No
> 
> This PR is only for my own usage to speed up the install script setup and always choosing the for me correct options.

@enoch85 Is anything wrong with that?

enoch85:
More options = more things that could go wrong.

That's why I'm being restrictive on changes in the main scripts. 

My time is limited so I can't catch up on the same phase as you all the time. :)

szaimen:
> More options = more things that could go wrong.
> 
> That's why I'm being restrictive on changes in the main scripts.
> 
> My time is limited so I can't catch up on the same phase as you all the time. :)

Thanks for the explanation! So then, please go ahead and close it 👍

PR #1649, opened 2020-11-02
Add an option to ntfs-3g mounts to speed up performance

szaimen:


szaimen:
Sry, seems like it is deprecated

PR #1650, opened 2020-11-02
Create smart-monitoring.sh to allow continuosly smart checking

szaimen:
I've already tested that the notifications work by manually setting `-W 5,45,55` to something like `-W 1,15,20` and restarting smartmontools, hence marking as ready for review.


disk/smart-monitoring.sh
```diff
@@ -0,0 +1,91 @@
...
+# Write conf to file
+# https://wiki.debianforum.de/Festplattendiagnostik-_und_%C3%9Cberwachung#Beispiel_3
+echo "DEVICESCAN -a -I 194 -W 5,45,55 -r 5 -R 5 -n standby,24 -m <nomailer> -M exec $SCRIPTS/smart-notification.sh \
+-s (O/../.././(06|18)|S/../../6/02|L/../20/./02)" > /etc/smartd.conf
+
+# Create smart notification script
+cat << SMART_NOTIFICATION > "$SCRIPTS/smart-notification.sh"
+#!/bin/bash
+
+# T&M Hansson IT AB © - 2020, https://www.hanssonit.se/
```
enoch85, line 55: Copy paste error?

szaimen: You mean that there is a copyright notice?
I thought this would be good to have...

enoch85:
You know, this actually on works on bare bone servers. Also, it would cause unnecessary tear and wear on the HDD if it's always checking. 

I think it's enough to test once per week as it's not suuper urgent to replace a disk with SMART errors, one week should be more than enough. Please try to convince me that this is better. :)

disk/smart-monitoring.sh
```diff
@@ -0,0 +1,91 @@
...
```
enoch85, line 3: copyright

disk/smart-monitoring.sh
enoch85: copyright

disk/smart-monitoring.sh
```diff
@@ -0,0 +1,91 @@
...
+exit
+SMART_NOTIFICATION
+
+# Make it executable
+chmod +x "$SCRIPTS/smart-notification.sh"
+
+# Restart service
+start_if_stopped smartmontools
+
+msg_box "S.M.A.R.T Monitoring was successfully set up."
```
enoch85, line 89: This will always output success even if it failed.

disk/smart-monitoring.sh
szaimen: Could you please be a bit clearer what you want me to do here?
It is acutally like this in every script e.g.: https://github.com/nextcloud/vm/blob/98f3ac340cfeab813133c6397d90e61e8d874a0c/static/change-ncadmin-profile.sh#L3

disk/smart-monitoring.sh
szaimen: Could you please be a bit clearer what you want me to do here?
It is acutally like this in every script e.g.: https://github.com/nextcloud/vm/blob/98f3ac340cfeab813133c6397d90e61e8d874a0c/static/change-ncadmin-profile.sh#L3

disk/smart-monitoring.sh
szaimen: What do you mean with `even if it failed` ?
If the script reaches this popup, it technically was a success.
What would you suggest to change here?

szaimen:
> You know, this actually on works on bare bone servers. Also, it would cause unnecessary tear and wear on the HDD if it's always checking.

yes, I know, that this only works on bare bone servers, but that is also the case with the smartctl.sh script.
Actually it would not tear and wear on the HDD since it only checks the smart stats when the disk is online:
`-n standby,24` means that it only wakes up a sleeping HDD every 24th time. Based on the standard checking time every half an hour, it would thus only wake up any sleeping HDD every 12 hours, which should be fine, imo.

> I think it's enough to test once per week as it's not suuper urgent to replace a disk with SMART errors, one week should be more than enough. Please try to convince me that this is better. :)

As new script also reports other things like too high temperatures and such, I think, that this makes sense to report as soon as possible. Here is an explanation what it does:

The command `DEVICESCAN -a -I 194 -W 5,45,55 -r 5 -R 5 -n standby,24 -m <nomailer> -M exec $SCRIPTS/smart-notification.sh -s (O/../.././(06|18)|S/../../6/02|L/../20/./02)` does the following things:

`DEVICESCAN` means that it checks all disks that have smart stats
`-a` Observes the health indicator, all attributes in the event of failures and their changes in the normalized values, error and self-test log and whether the current pending sector count and the offline uncorrectable count are greater than 0.
`-I 194` Ignores changes in normalized temperature values.
`-W 5.45.55` Informs when the temperature has changed by 5 ° C, rises above 45 ° C and warns at 55 ° C as a critical value.
`-r 5, -R 5` Also shows the raw value of the Reallocated_Sector_Ct and reports changes.
`-n standby,24` Does not poll the hard drive 24 times when it is on standby. With 30 minutes * 24 = 660 minutes, that's 12 hours that you will not be woken up by a smartd check.
`-m <nomailer>` Sends no mail directly, if there are errors and `-M exec $SCRIPTS/smart-notification.sh` invokes the script which gets executed if there are errors
`-s (O/../.././(06|18)|S/../../6/02|L/../20/./02)` Performs a direct offline test every 6am and 18am; a short selftest every Saturday at 2 a.m. and a long selftest every 20th of the month at 2 a.m.

Is everything clear now? :)

disk/smart-monitoring.sh
enoch85: Do `start_if_stopped smartmontools then`, so that we have some kind of check at least.

enoch85:
> the disk is online

Since it's a cloud server it would be kind of strange if the disk slept for more than 5 minutes. :)

Sure, we could notify for high disk temperatures, but who are we to decide which is high or not. I mean there's a big difference between if the server is located in the northern parts of Sweden or Florida in USA.

I also think that we shouldn't make another script, but improve the already existing one instead.

I still also think that the checks are excessive. On my production servers I check SMART every two weeks... I never had any issues with replacing a disk when it warned. SMART is just warning that this *might* fail *soon*. Not that it has failed, or that it's urgent to replace the disk.

> Is everything clear now? :)

Thanks for the explanation. :)

szaimen:
> > the disk is online
> 
> Since it's a cloud server it would be kind of strange if the disk slept for more than 5 minutes. :)

I would say it depends if you have connected more than one disk. If yes, then you won't want all of them to wake up in this case.

> Sure, we could notify for high disk temperatures, but who are we to decide which is high or not. I mean there's a big difference between if the server is located in the northern parts of Sweden or Florida in USA.

No, since most HDDs are not meant to run at a higher temperature than around 50-60 °C
So, the temperature checking is already quite good, imo.

> I also think that we shouldn't make another script, but improve the already existing one instead.

Hm, not sure how to improve the existing script. It is just a different approach imo.
What about putting this into the not-supported folder, too? If you don't want it then its fine, but I definitely want it ;)

> I still also think that the checks are excessive. On my production servers I check SMART every two weeks... I never had any issues with replacing a disk when it warned. SMART is just warning that this _might_ fail _soon_. Not that it has failed, or that it's urgent to replace the disk.

Hm, I don't think so: e.g. if a drive is reporting any problem I want to be informed as soon as possible and not in the worst case after 7 days. Also concerning high disk temperatures.

> > Is everything clear now? :)
> 
> Thanks for the explanation. :)

👍

disk/smart-monitoring.sh
szaimen: okay, good idea 👍

szaimen: done

disk/smart-monitoring.sh
enoch85: There's copyright in two places, I don't see why?

Is it two different scripts?

enoch85:
Here we have two scripts doing the same thing, this being a little more extensive.

For merging to happen;

1. Merge the already existing script and this into one.
2. Make it optional to get notified directly, or once per week.
3. Remove the offline testing "Performs a direct offline test every 6am and 18am;" or change it to once every second day. Remove it it's not supported: https://www.smartmontools.org/wiki/test_offline (in other words, make a check it it's supported or not).
4. Translate the values into something understandable when notifying, or document it properly.

disk/smart-monitoring.sh
szaimen: yes, I see it as a different script, since we create the script "$SCRIPTS/smart-notification.sh" here.
If you would open it after running this script here, you would see the copyright header, too. I thought this would make sense?

szaimen:
> 1. Merge the already existing script and this into one.

done

> 2. Make it optional to get notified directly, or once per week.

done

> 3. Remove the offline testing "Performs a direct offline test every 6am and 18am;" or change it to once every second day. Remove it it's not supported: https://www.smartmontools.org/wiki/test_offline (in other words, make a check it it's supported or not).

I've changed it to once every second day.

> 4. Translate the values into something understandable when notifying, or document it properly. Like `Reallocated_Event_Count = 98321` Eehm, OK? What does that actually mean?

This is out of scope for this PR, imo. Maybe this would be something for a follow up PR.

szaimen:
I've tested this successfully hence marking as ready for review

PR #1652, opened 2020-11-02
ClamAV - add a mechanism to inform about found files

szaimen:
Todo: 
- [x] fix all todos
- [x] escape all $ inside the cat file

### DIFF
--- a/apps/clamav.sh
+++ b/apps/clamav.sh
@@ -185,7 +185,7 @@ chown root:root "$SCRIPT_PATH"
 chmod 700 "$SCRIPT_PATH"
 
 # Create the cronjob
-crontab -u root -l | { cat; echo "*/30 * * * * $SCRIPT_PATH 2>&1"; } | crontab -u root -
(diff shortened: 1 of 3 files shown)


szaimen:
This was successfully tested with a test-virus-file hence marking as ready for review

apps/clamav.sh
```diff
@@ -36,7 +36,9 @@ else
     rm -f /etc/systemd/system/clamav-daemon.service
     rm -f "$SCRIPTS"/clamav-fullscan.sh
     rm -f "$VMLOGS"/clamav-fullscan.log
+    rm -f "$SCRIPTS/nextcloud-av-notification.sh"
     crontab -u root -l | grep -v 'clamav-fullscan.sh'  | crontab -u root -
```
enoch85, line 40: It should still delete any old entry if this script was run previous to this change imo.

enoch85:
I will merge when you add it to the not-supported again. :+1:

szaimen:
> I will merge when you add it to the not-supported again. 👍

done 👍

PR #1653, opened 2020-11-04
fix a bug in midnight-commander.sh

szaimen:


szaimen:
This was successfully tested, hence marking as ready for review

PR #1659, opened 2020-11-08
fix some more issues with the backup scripts

szaimen:


szaimen:
This was successfully tested :)

szaimen:
Now it is ready :)

enoch85:
@szaimen Could this be used to cluser a device?

I mean a 1:1 copy?

szaimen:
> @szaimen Could this be used to cluser a device?
> 
> I mean a 1:1 copy?

Hm... I am honestly not sure...
Of course the whole root partition is backed up but I don't know how this should be used in a real time cluster...

enoch85:
I haven't even checked the script, but does it do DB backup as well? And are the apache server stopped while the backup runs?

szaimen:
The script stops all services (at least the most important currently), creates a snapshot, starts the services again and makes a backup from the snapshot.
This has the advantage, that the downtime can get reduced to a minimum.

szaimen:
https://github.com/nextcloud/vm/blob/master/not-supported/borgbackup.sh#L195-L213
https://github.com/nextcloud/vm/blob/master/not-supported/borgbackup.sh#L232-L268

enoch85:
Is it posdible to send the backup to another location?

Would be cool if you could build in support for that as well, so that the complete server coukd be started at another location in just seconds.

Also, restoring the ZFS snaphot on the co-location would be nice. 👍🏻

I'm thinking we could use this to have an exact copy of a server somewhere else that could be restored in case of something went wrong with the original one.

szaimen:
> Is it posdible to send the backup to another location?

Borg supports ssh, yes

> Would be cool if you could build in support for that as well, so that the complete server coukd be started at another location in just seconds.

I doubt that it will be seconds since you have to restore the important files from the backup repository first. More likey minutes hours depending on the size and network speed if sent over ssh

> Also, restoring the ZFS snaphot on the co-location would be nice. 👍🏻

This is theoretically possible but has the disadvantage the you cannot simply mount one backup archive and show the files inside.

> I'm thinking we could use this to have an exact copy of a server somewhere else that could be restored in case of something went wrong with the original one.

Theoretically possible but you will need a second server for that (a server that receives the backup) and since you are using a backup repository, you will first need to restore the files from there before you can use it again.

PR #1663, opened 2020-11-12
previewgenerator - make sure that the cronjob gets created

szaimen:
I've changed my mind: since the initial scan can take very long depending on how much pictures and files you have (it can be days), we should make sure that the cronjob is added before the initial scan is done. Else it is very likely that the cronjob will never be added if something fails during the initial scan (e.g. a restart would also kill it)


enoch85:
Please change the subtitle so it's easier to search for old commits.

Thanks!

PR #1677, opened 2020-11-14
fix clamav full-scan

szaimen:
It seems like I somehow missed this...
Now it starts correctly

The creation of the full-scan script was successfully tested.


enoch85:
@szaimen Please sign off your commits so that DCO passes.

Thanks!

PR #1678, opened 2020-11-14
allow to reinstall automatic updates

szaimen:

### DIFF
--- a/addons/automatic_updates.sh
+++ b/addons/automatic_updates.sh
@@ -33,10 +33,8 @@ In the next step you will be able to choose to proceed or exit." "$SUBTITLE"
 
 if yesno_box_yes "Do you want to enable automatic updates?"
 then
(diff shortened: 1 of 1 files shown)


szaimen:
@enoch85 What about beeing able to reinstall automatic updates? (hence beeing able to disable it?)

enoch85:
> @enoch85 What about beeing able to reinstall automatic updates? (hence beeing able to disable it?)

Sure, why not. 

I also just added a new issue regarding the updates here: https://github.com/nextcloud/vm/issues/1680

I would like to refactor this in general actually.

enoch85:
But, I don't get this fix? This wouldn't write anything to the log, right?

szaimen:
> But, I don't get this fix? This wouldn't write anything to the log, right?

No, this will write also error messages to the logfile which it doesn't do currently.

szaimen:
this will be implemented with #1681

PR #1681, opened 2020-11-16
allow to reinstall automatic updates

szaimen:

### DIFF
--- a/addons/automatic_updates.sh
+++ b/addons/automatic_updates.sh
@@ -4,10 +4,13 @@
 
 true
 SCRIPT_NAME="Automatic Updates"
+SCRIPT_EXPLAINER="This option will update your server every week on Saturdays at $AUT_UPDATES_TIME:00.
+The update will run the built in script '$SCRIPTS/update.sh' which will update both the server packages and Nextcloud itself.\n
+You can read more about it here: https://www.techandme.se/nextcloud-update-is-now-fully-automated/
+Please keep in mind that automatic updates might fail hence it's \
+important to have a proper backup in place if you plan to run this option."
 # shellcheck source=lib.sh
 source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
-
-print_text_in_color "$ICyan" "Configuring automatic updates..."
 
 # Check for errors + debug code and abort if something isn't right
 # 1 = ON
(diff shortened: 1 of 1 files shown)


addons/automatic_updates.sh
```diff
@@ -18,29 +21,30 @@ debug_mode
...
+if ! crontab -u root -l | grep -q "$SCRIPTS/update.sh"
+then
+    # Ask for installing
+    install_popup "$SCRIPT_NAME"
+else
+    # Ask for removal or reinstallation
+    reinstall_remove_menu "$SCRIPT_NAME"
+    # Removal
+    crontab -u root -l | grep -v "$SCRIPTS/update.sh"  | crontab -u root -
+    sed -i "s|/sbin/shutdown -r +1||g" "$SCRIPTS"/update.sh
```
enoch85, line 34: I think we use `reboot` in an earlier version as well. Please check history.

addons/automatic_updates.sh
```diff
@@ -18,29 +21,30 @@ debug_mode
...
-    then
-        sed -i "s|exit|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
-        echo "exit" >> "$SCRIPTS"/update.sh
-    fi
-    msg_box "Please remember to keep backups in case something should go wrong, you never know." "$SUBTITLE"
+    sed -i "s|exit|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
+    echo "exit" >> "$SCRIPTS"/update.sh
 fi
+
+msg_box "Please remember to keep backups in case something should go wrong, you never know."
```
enoch85, line 48: Please change from *should* to *would*.

addons/automatic_updates.sh
szaimen: hm... I think I am just deleting this line... done

addons/automatic_updates.sh
szaimen: I've deleted the word since it sounds wrong

addons/automatic_updates.sh
enoch85: No, we don't want to reboot if it's uninstalled... Don't make it easy, make it stable. :D

szaimen:
Is anything missing here?

addons/automatic_updates.sh
szaimen: done

szaimen:
This should be done now

szaimen:
@enoch85 is anything missing here?

enoch85:
> @enoch85 is anything missing here?

No, is just don't have time. Need to test this first. :+1:

enoch85:
`AUT_UPDATES_TIME` does not work.

Please have a look.

enoch85 (changes requested):
See comment

szaimen:
> `AUT_UPDATES_TIME` does not work.

I`ve just tested this and it works for me.

Maybe there is something wrong with your test instance?

enoch85:
It's a production server. ;)

Please check again, the script explainer doesn't pick up the variable.

szaimen:
Ah, this is the problem. Thanks for clarifying!

szaimen:
Should work now 👍

Issue #1694, opened 2020-11-23
grub-probe: error: unknown filesystem.


enoch85:
```
grub-probe: error: unknown filesystem.
/usr/sbin/grub-probe: error: unknown filesystem.
Found Ubuntu 20.04.1 LTS (20.04) on /dev/mapper/ubuntu--vg-NcVM--snapshot
done
Maintenance mode already disabled
Trying to automatically update all Nextcloud apps...
Your apps are already up to date!
Checking for new Nextcloud version...
Latest release is: 20.0.2. Current version is: 19.0.5.
```

cc @szaimen 

Saw this during an upgrade at a customer.

szaimen:
@enoch85 This is an issue of update-grub because all partitions get scanned while updating grub for existing bootable partitions (and snapshots). But this message hasn't caused any issue in my testing. Also update-grub returns `0`  if this message appears (so it works without issue except this message)

If you want to suppress this message, we could add `GRUB_DISABLE_OS_PROBER=true` to `/etc/default/grub` which does this:

![image](https://user-images.githubusercontent.com/42591237/100004920-7edc2d80-2dc8-11eb-9b1c-560fd86ae622.png)

**Translation:** _If this variable is added with true, a renewed check for other operating systems on the PC is suppressed during an update grub. This means that only the kernel images contained in the /boot directory and your own scripts are recorded in the Grub menu according to the numbering._

szaimen:
BTW: this only happens if you have opted in to use snapshots (during the nextcloud_install_script) which is still not the default.

szaimen:
if opened #1695 with a fix

PR #1695, opened 2020-11-23
fix issue with grub-probe

szaimen:
fix https://github.com/nextcloud/vm/issues/1694

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -77,6 +77,11 @@ It should work afterwards again."
     fi
     nextcloud_occ maintenance:mode --off
     start_if_stopped docker
+    # Adjust grub (https://github.com/nextcloud/vm/issues/1694)
+    if ! grep -q "GRUB_DISABLE_OS_PROBER" /etc/default/grub
+    then
+        echo "GRUB_DISABLE_OS_PROBER=true" >> /etc/default/grub
+    fi
 fi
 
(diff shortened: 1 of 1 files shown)


szaimen:
https://github.com/nextcloud/vm/pull/1695/files#diff-67bceb0888b398c45857e1bc41f753da1a696075ff08b9b82f8136bc6af0139aR40-R85

enoch85:
Great fix, but I'm not sure we want to mess with kernel *defaults*.

Which downside does this bring?

szaimen:
> Which downside does this bring?

As I've written here: https://github.com/nextcloud/vm/issues/1694#issuecomment-732373150

> **Translation:** _If this variable is added with true, a renewed check for other operating systems on the PC is suppressed during an update grub. This means that only the kernel images contained in the /boot directory and your own scripts are recorded in the Grub menu according to the numbering._

enoch85:
Hmm sorry missed that.

Let's leave it as is then. We would want the latest stuff right?

szaimen:
> Let's leave it as is then. We would want the latest stuff right?

Hm... since this issue only appears when the snapshot exists and this fix only gets applied if the snapshot exists, I would say that this fix is fine. 

This is the only reliable way to fix #1694 so I guess it is up to you to decide what to do with this PR.

enoch85:
I've investigated. Let's not do this.

Thanks for the PR anyway. :+1:

PR #1702, opened 2020-11-30
Fix PHP error mesage

szaimen:
This fixes the following error message:
```
Notice: Trying to access array offset on value of type bool in PEAR/REST.php on line 187
PHP Notice:  Trying to access array offset on value of type bool in /usr/share/php/PEAR/REST.php on line 187
```

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -197,6 +197,9 @@ then
     fi
 fi
 
+# Fix PHP error mesage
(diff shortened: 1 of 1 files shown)


szaimen:
https://stackoverflow.com/questions/59720692/php-7-4-1-pecl-is-not-working-trying-to-access-array-offset-on-value-of-type for reference

Issue #1577, opened 2020-10-16
Native English speakers are wanted!


enoch85:
Both @szaimen and I are not native English speakers, and the text in the inform boxes might be a bit off in some places.

We would really need someone that looked through all the `msg_box`, `yesno_yes`, `yesno_no` and other informative boxes in general that are presented to the users in the scripts.

Payment can be discussed when the PR is merged.

Thanks!

tyleroconnell:
@enoch85 @szaimen I'd be happy to help! I'm a native English speaker that came across this repo when looking for a simple and easily maintainable solution to manage Nextcloud with my hypervisor. I'm more than willing to contribute free of charge, given all the time I hope your work will save me 😄

To confirm so I can complete this PR in one go, you'd like me to review the text contained within any instance of: `msg_box`, `yesno_box_yes`, `yesno_box_yes`, `input_box`, and `input_box_flow`? Any others?

enoch85:
Wow @tyleroconnell that's great!

Yes, correct, those boxes and basically everything in [lib.sh](https://github.com/nextcloud/vm/blob/master/lib.sh) that contains text presented to the user. Easiest way to find out is to search for `$1` `$2` `$3` and see if you can find any occurrences were text is presented to the user.

It would be great to actually check *everything* but that's much I know. :)

A single big PR would work, yes. :)

szaimen:
@tyleroconnell thanks from my side, too!

As a small remark: for now it should be enough to inspect all scripts outside the [not-supperted folder](https://github.com/nextcloud/vm/tree/master/not-supported) and the [old folder](https://github.com/nextcloud/vm/tree/master/old), I guess.

enoch85:
Here's something to work with:


Filtered Findings (70)

migrate/docker/readme.md|5 col 115| "soley" is a misspelling of "solely"
migrate/docker/readme.md|17 col 26| "administor" is a misspelling of "administer"
nextcloud-startup-script.sh|110 col 22| "recomended" is a misspelling of "recommended"
nextcloud-startup-script.sh|206 col 14| "desinged" is a misspelling of "designed"
nextcloud_update.sh|293 col 58| "compability" is a misspelling of "compatibility"
not-supported/remotedesktop.sh|94 col 22| "successfuly" is a misspelling of "successfully"
not-supported/smbserver.sh|526 col 64| "directoy" is a misspelling of "directory"
not-supported/smbserver.sh|920 col 45| "additonal" is a misspelling of "additional"
[...]

PR #1712, opened 2020-12-09
Issue 1577: Grammatical corrections

tyleroconnell:
Per the discussion in #1577, here are some grammatical fixes. With text frequently repeated across files, I think it would be beneficial to have scripts source their text from one file. This would help from both a maintainability and re-use perspective, though there's a legitimate counter-argument that this would make the scripts harder to read.

I did not review the entire repository but my first hour of review has made clear the great job you all have already done! As a native English speaker, I found everything to be very clear, with these changes more stylistic than anything else. If there are any specific files you would like me to review before merging I'd be happy to do so, but like I said this is in a very good spot already.

Thank you for all of this hard work, it's quite impressive!


enoch85:
Thank you soo much! This is great!

I'll leave this open in case you want to continue on your fork doing the rest.

Please let me know when you're done. :rocket:

tyleroconnell:
@enoch85 given how good of a state the grammar already is, it doesn't seem necessary to me for all files to be reviewed. Is there any specific one not touched that you would like reviewed?

enoch85:
Naah, it's fine. Let's merge this. :+1: 

Thank you very much for both this fix and the feedback!

PR #1718, opened 2020-12-14
smbserver - add a fixed netbios name

szaimen:
This is needed to be able to connect always using `nextcloud` as server name.

### DIFF
--- a/not-supported/smbserver.sh
+++ b/not-supported/smbserver.sh
@@ -96,6 +96,14 @@ then
     fi
 fi
 
+# Set netbios name to a fixed name to reach the server always by using nextcloud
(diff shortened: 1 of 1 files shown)


szaimen:
This is ready :)

enoch85:
Would it be possible to add the hostname instead? `$(hosname)`

szaimen:
> Would it be possible to add the hostname instead? `$(hosname)`

The hostname already works in my testing. So you are with this fix able to access the smb server using the hostname and nextcloud as server name.

This fix is necessary since the hostname is proven to change during the activate tls script and accessing the smb server using the hostname is thus not stable because it can change

szaimen:
Anything missing here?

Issue #1724, opened 2020-12-16
Implement deSEC with domain and DynDNS (automatically)


enoch85:
Had a talk with the deSEC guys and they made it very easy for us to provide both domain and DynDNS to the end users!

Domain Registration:
https://desec.readthedocs.io/en/latest/auth/account.html#zone-creation-during-account-registration

DynDNS:
https://desec.readthedocs.io/en/latest/dyndns/update-api.html

Things to consider:
1. The user must use a VALID email address since the domain will be destroyed after 6-7 months of they don't confirm it
2. DynDNS registration can only happen with the API token so it has to be done after the registration of the domain, with a `msg_box` to fetch the API token and then eneter it in an input field.

enoch85:
cc  @peterthomassen

PR #1725, opened 2020-12-16
deSEC magic

enoch85:
cc @peterthomassen Just notes basically.

@szaimen This is suuuper cool stuff.

Fix https://github.com/nextcloud/vm/issues/1724

> https://github.com/desec-io/desec-stack/pull/511#issuecomment-767882975


enoch85:
@peterthomassen

The current error are:

```
{"captcha":{"id":["CAPTCHA does not exist."]}}
```

Is that expected?

peterthomassen:
> The current error are:
> 
> ```
> {"captcha":{"id":["CAPTCHA does not exist."]}}
> ```
> 
> Is that expected?

That happens if you are using a capture that you've used before. (Captchas are for one-time use and invalidated after first use. In addition, they also expire after 24 hours.)

addons/desec.sh
```diff
@@ -0,0 +1,76 @@
...
+    if [[ "$SUBDEDYN" == *" "* ]]
+    then
+        msg_box "Please don't use spaces."
+    elif [ "${SUBDEDYN//[A-Za-z0-9]}" ]
+    then
+        msg_box "Allowed characters for the username are:\na-z', 'A-Z', and '0-9'\n\nPlease try again."
+    else
+        DEDYNDOMAIN="$SUBDEDYN.dedyn.io"
+        break
+    fi
```
szaimen, line 30: we should already check in this loop if the subdomain is avaliable. Is there an API for that?

addons/desec.sh
```diff
@@ -0,0 +1,76 @@
...
+# Ask user if DynDNS should be added to the domain
+if yesno_box_yes "Do you want to add automatic updates of your WAN IP - IPv4 and/or IPv6?"
+then
+    # Add DynDNS
+    curl --user "$DEDYNDOMAIN":"$AUTHTOKEN" \
+      https://update.dedyn.io/?myipv4="$WANIP4"\&myipv6=fd08::1234
+
+    curl https://update6.dedyn.io/?hostname="$DEDYNDOMAIN"?myipv4="$WANIP4"\&myipv6=fd08::1234 \
+      --header "Authorization: Token $DEDYNAUTHTOKEN"
+fi
```
szaimen, line 76: There is a cronjob missing that updates the ip address if it changes...

addons/desec.sh
peterthomassen: You can query the DNS to see if the subdomain has an SOA record. If yes, it is taken. If not, it generally is available (with the exception of some reserved names, in particular names like `autoconfig` and `autodiscover`, and names beginning with `_`).

addons/desec.sh
peterthomassen: Also, it is sufficient to to the update via `update.dedyn.io`, which is a dual-stack interface. The second `curl` call to `update6.dedyn.io` resolves to IPv6 only.

enoch85: @peterthomassen like that?

peterthomassen: Almost. You have two ways of authentication:

1. Basic auth: `curl --user $DEDYNDOMAIN:$AUTHTOKEN`
2. Authorization header: `--header "Authorization: Token $DEDYNAUTHTOKEN"`
3. Query string auth: `?...&username=$DEDYNDOMAIN&password=$AUTHTOKEN`

Query string authentication is discouraged, as we log request URLs temporarily.

The easiest method is Basic auth. If you go for header authorization, it is advisable to stlil pass `&username=$DEDYNDOMAIN` in the query string so that the domain can be identified if the user has several. (The request would otherwise fail.)

You may want to consider sanitizing the variables to prevent shell command injection (not sure if you're doing that already).

enoch85: Thanks!

> You may want to consider sanitizing the variables

Yeah sure, hence WIP. :)

enoch85: Btw, still waiting for the change at your end so that we can start test and devlop this for real.

peterthomassen: yep, I'm planning to work on this this month.

addons/desec.sh
```diff
@@ -0,0 +1,74 @@
...
+$DEDYNPASS
+
+Please write this down now."
+
+# Register the domain
+curl -X POST https://desec.io/api/v1/auth/ \
+    --header "Content-Type: application/json" --data @- <<EOF
+    {
+      "email": "$VALIDEMAIL",
+      "password": "$DEDYNPASS",
```
peterthomassen, line 51: This is the account password, not the update token. Most users don't need to log into their account, so I recommend disabling password login with `"password": null`.

What you need is an API token for updating the IP address of the domain. That "update token" is displayed after the user has confirmed their email address. You'll have to ask the user to check their mail, confirm their account, and then copy the token from the confirmation page.

enoch85:
Plan to work on this once the code is merged upstream.

peterthomassen:
We merged and deployed https://github.com/desec-io/desec-stack/pull/511, so you should be able to start using machine sign-ups with postponed captcha.

enoch85:
@peterthomassen Super cool, thanks! :rocket: 

Now *I* need to find the time to dig into this again. Right now all I need is :sleeping:

enoch85:
@peterthomassen Hey, would be great if we could book a session where we implement this together, I need to freshen my memory from the last time we spoke.

You know dnsSec inside out, and I know the Nextcloud VM inside out. I think it would be a quick fix if we sat down together and had a look at this.

Would it work for you?

peterthomassen:
@enoch85 Sure, how about in the next three hours, or tomorrow after 15:30 CET? Feel free to send me an invite by email.

addons/desec.sh
```diff
@@ -0,0 +1,75 @@
...
+if host -t SOA $DEDYNDOMAIN
+then
+    msg_box "Sorry, but it seems like $DEDYNDOMAIN is taken. Please try with another domain."
+    exit
+fi
+
+# Email address
+msg_box "You will now be prompted to enter an email address. It's very important that the email address you enter it a 100% valid one.
+
+Every 6 months you will get an email asking you to confrim your domain. If deSeC doesn't get an answer within 3 weeks, you domain will be destroyed!"
```
peterthomassen, line 42:
```suggestion
Every 6 months you will get an email asking you to confrim your domain. If you do not react within a few weeks, your domain will be destroyed!"```

File: addons/desec.sh [hunk 5]
```diff
@@ -0,0 +1,75 @@
...
+
+# Check for SOA record
+if host -t SOA $DEDYNDOMAIN
+then
+    msg_box "Sorry, but it seems like $DEDYNDOMAIN is taken. Please try with another domain."
+    exit
+fi
+
+# Email address
+msg_box "You will now be prompted to enter an email address. It's very important that the email address you enter it a 100% valid one.
```
peterthomassen, line 39:
```suggestion
msg_box "You will now be prompted to enter an email address. It's very important that the email address you enter it a 100% valid one. deSEC will verify your email address by sending you a verification link.
```

addons/desec.sh
```diff
@@ -0,0 +1,74 @@
...
+done
+
+# Email address
+msg_box "You will now be prompted to enter an email address. It's very important that the email address you enter it a 100% valid one.
+
+Every 6 months you will get an email asking you to confrim your domain. If deSeC doesn't get an answer within 3 weeks, you domain will be destroyed!"
+
+VALIDEMAIL=$(input_box_flow "Please enter a valid email address. NOT a fake or a temporary one.")
+
+msg_box "If you want to enter your account, please reset your password here:
```
peterthomassen, line 45:
```suggestion
msg_box "If you later want to log into your account, please reset your password here (you don't need to do this now):
```

addons/desec.sh
```diff
@@ -0,0 +1,74 @@
...
+# Email address
+msg_box "You will now be prompted to enter an email address. It's very important that the email address you enter it a 100% valid one. deSEC will verify your email address by sending you a verification link.
+
+Every 6 months you will get an email asking you to confrim your domain. If deSeC doesn't get an answer within 3 weeks, you domain will be destroyed!"
+
+VALIDEMAIL=$(input_box_flow "Please enter a valid email address. NOT a fake or a temporary one.")
+
+msg_box "If you want to enter your account, please reset your password here:
+
+https://desec.io/reset-password"
```
peterthomassen, line 47:
```suggestion
msg_box "```suggestion
msg_box "If you later want to log into your deSEC account, you need to set a login password here: https://desec.io/reset-password

You don't need to do this now."
```

addons/desec.sh
```diff
@@ -0,0 +1,75 @@
...
+    else
+        DEDYNDOMAIN="$SUBDEDYN.dedyn.io"
+    fi
+    # Check for SOA record
+    if ! host -t SOA $DEDYNDOMAIN
+    then
+        msg_box "Sorry, but it seems like $DEDYNDOMAIN is taken. Please try with another domain."
+    else
+        break
+    fi
```
peterthomassen, line 33:
```suggestion
    else
        DEDYNDOMAIN="$SUBDEDYN.dedyn.io"
        # Check for SOA record
        if ! host -t SOA $DEDYNDOMAIN
        then
            msg_box "Sorry, but it seems like $DEDYNDOMAIN is taken. Please try with another domain."
        else
            break
        fi
    fi
```

addons/desec.sh
```diff
@@ -0,0 +1,88 @@
...
+
+Please copy that and enter it in the next box after you hit OK."
+
+# Check if DEYNAUTH is valid
+while :
+do
+    DEDYNAUTHTOKEN=$(input_box_flow "Please enter your auth token for deSEC, please make sure it's valid!")
+    if [ $(curl -s -o /dev/null -w '%{http_code}' --user "$DEDYNDOMAIN":"$DEDYNAUTHTOKEN" https://update.dedyn.io/?myipv4=\&myipv6=) -eq 401 ]
+    then
+        if ! yesno_box_yes "Sorry, but it seems like the CAPTCHA is incorrect. Do you want to try again?"
```
peterthomassen, line 71:
```suggestion
        if ! yesno_box_yes "Sorry, but it seems like the auth token (update password) is incorrect. Do you want to try again?"
```

addons/desec.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# Ask user to check email and confirm to get the token
+msg_box "If the registration was successful you should have got an email with your auth token.
+
+Please copy that and enter it in the next box after you hit OK."
+
+# Check if DEYNAUTH is valid
+while :
+do
+    DEDYNAUTHTOKEN=$(input_box_flow "Please enter your auth token for deSEC, please make sure it's valid!")
```
peterthomassen, line 68:
```suggestion
    DEDYNAUTHTOKEN=$(input_box_flow "Please enter your auth token (update password) for deSEC, exactly as it was displayed (use correct casing, no extra spaces).")
```

enoch85:
@szaimen This is now done. Could you please review this properly?

I'm thinking we will add the TLS function to this as well as a second step. Right now you "only" get a domain automatically with deSEC (which is super cool).

addons/desec.sh
```diff
@@ -0,0 +1,91 @@
...
+WANIP6=$(curl -s -k -m 5 https://ipv6bot.whatismyipaddress.com)
+
+# Ask user if DynDNS should be added to the domain
+if yesno_box_yes "Do you want to add automatic updates of your WAN IP - IPv4 and/or IPv6?"
+then
+    # Add DynDNS
+    curl --user "$DEDYNDOMAIN":"$DEDYNAUTHTOKEN" \
+      https://update.dedyn.io/?myipv4="$WANIP4"\&myipv6="$WANIP6" >/dev/null 2>&1
+fi
+
```
szaimen, line 89: I think here is still a cronjob missing that updates the ip address regularly (very possible that it changes regularly, at least of home networks). Is there a rate limit concerning this? If not, I would recommend to just run the update every 5 minutes, or what is recommended how to do this?

addons/desec.sh
```diff
@@ -0,0 +1,91 @@
...
+#### TODO
+# Ask if the user wants to add TLS (use the function)
```
szaimen, line 91: I would recommend to allow to run the activate-tls script here directly instead of adding just another place where we add TLS. Doesn't this make more sense? Or is this also menat for other usecases like e.g. for onlyoffice, collabora, bitwarden, etc.?

szaimen:
This looks already pretty good apart frome some places with strange indentation and my comments. 👍
And does it already work? :)

addons/desec.sh
peterthomassen: Our rate limit is 1/min.

The recommended way is to first check whether an IP update is necessary, by determining the current IPv4 and IPv6 addresses, e.g. via https://checkipv4.dedyn.io/ or https://checkipv6.dedyn.io/.

There is tooling which automates this, e.g. ddclient. You can run it as a cron job, and it will only trigger an update request if there is something to update.

This greatly reduces load on our side. With the (exciting!) prospect of many Nextcloud VM instances using this, we would appreciate if you go go this route (not necessarily via ddclient; you may want to store the last IP values in /var/tmp/ or something).

enoch85:
@szaimen It already works! 

Please feel free to add your changes, then test that everything is working by running it. @peterthomassen will delete your email if you email the deSEC support after the test was made.

addons/desec.sh
enoch85: if "A-Record" -ne $WANIP4
then
   run updatededyn
fi

Something like that would work.

addons/desec.sh
enoch85: Yes, that was my first thought as well. But, we have to make it work with the current menu flow. 

I'm too tired to think right now.

addons/desec.sh
szaimen:
```shell
dig "$DEDYNDOMAIN" AAAA +short
dig "$DEDYNDOMAIN" A +short
```

peterthomassen: I think that IPv6 should generally be supported today; not doing so just creates more technical debt. Just my 2ct ...

szaimen:
4cd5b33 should address bove of my comments.
It should already work but I'll test it tommorrow. Feel free to test it as well.

addons/desec.sh
```diff
@@ -0,0 +1,88 @@
...
+               exit
+            fi
+        else
+            break
+        fi
+    fi
+done
+
+# Email address
+msg_box "You will now be prompted to enter an email address. It's very important that the email address you enter it a 100% valid one. deSEC will verify your email address by sending you a verification link.
```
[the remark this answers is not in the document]
szaimen: @enoch85 what about this comment?

szaimen:
I've successfully tested my changes 👍
(I've also made sure that both ddclient-configuration.sh and activate-tls.sh still work as usual when running them separately from the desec.sh script).

szaimen:
I've just took the opportunity to add it to server_configuration.sh menu.
The menu was also already successfully tested,
So this should be finally ready 👍

menu/server_configuration.sh
```diff
@@ -48,6 +48,7 @@ choice=$(whiptail --title "$TITLE" --checklist \
 $CHECKLIST_GUIDE\n\n$RUN_LATER_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
 "Security" "(Add extra security based on this http://goo.gl/gEJHi7)" OFF \
+"deSEC" "(Automatically get a domain from desec.io, configure ddclient and activate TLS)" ON \
```
szaimen, line 51:
```suggestion
"deSEC" "(Automatically get a domain from desec.io, configure ddclient and activate TLS)" "$STARTUP_SWITCH" \
```
I think this makes more sense

addons/desec.sh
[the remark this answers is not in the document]
enoch85: I actually don't know what's wrong here, maybe you can have a look?

Issue #1715, opened 2020-12-13
"Error updating .htaccess file" at the end of the update process


e-caste:
## Steps to reproduce
1. `sudo bash /var/scripts/update.sh`

## Expected behaviour
The update finishes correctly.

## Actual behaviour
At the end of the update I get the following message:
```
Restoring the status of apps. This can take a while...
Setting RewriteBase to "/" in config.php...
System config value htaccess.RewriteBase set to string /
Error updating .htaccess file, not enough permissions or "overwrite.cli.url" set to an invalid URL?
Sorry but something went wrong. Please report this issue to https://github.com/nextcloud/vm/issues and include the output of the error message. Thank you!
sudo -u www-data php /var/www/nextcloud/occ maintenance:update:htaccess failed
Posting notification to users that are admins, this might take a while...
```
After this error is thrown, everything works correctly, so I wouldn't classify it as a critical bug. The only exception is that a new login from the iOS Nextcloud Talk app does not work unless I add the line `'overwriteprotocol' => 'https',` to `/var/www/nextcloud/config/config.php` (only the first time this error presents itself, then it is saved).  
This error has begun appearing after I had manually modified the config file I mentioned above. I have not modified the permissions of the .htaccess file mentioned in the error message.

## Server configuration

**Nextcloud server version:** 20.0.3

**Server OS** Ubuntu Server 18.04

**How did you install the VM?** Scripted install from master

## Network
**Do you use DHCP?**
No, static IP.

**Is port 80 and/or 443 open?**
Only 80 since I am using a reverse proxy on another VM.

## Logs / Screenshots

See error messages above.

#### Log file (/var/log/nextcloud/nextcloud.log)
```
Insert your nextcloud.log file here
```

#### Installation log (command output)
```
Insert the CLI output here
```

szaimen:
@e-caste what is `overwrite.cli.url` set to in config.php?

e-caste:
@szaimen it is set to `'https://subdomain.domain.tld'` (where subdomain, domain, and tld are actually valid values, I'm just censoring it here). I am behind a reverse proxy, so I'm not sure if that value is correct, or even if I should set it. Some online tutorials suggested to set it to this value, so that's why I did it.

Output of `sudo ls -l /var/www/nextcloud/.htaccess && sudo ls -l /mnt/ncdata/.htaccess`:
```bash
-rw-r--r-- 1 www-data www-data 3032 Dec  9 20:30 /var/www/nextcloud/.htaccess
[...]

PR #1730, opened 2020-12-19
remove trailing slash

enoch85:
Possible fix for https://github.com/nextcloud/vm/issues/1715#


szaimen:
@enoch85 this fix is incorrect: the trailing slash is needed for this to work. I guess @e-caste has just forgotten to set the trailing slash in overwrite.cli.url...

enoch85:
Aah, I was reading to quick!

PR #1731, opened 2020-12-19
geoblock: don't fail on nonexisting GeoIP.dat

thorstenb:
Data files /usr/share/GeoIP/GeoIP*.dat tend not to exist on fresh
installs.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -370,7 +370,8 @@ get_newest_dat_files() {
             print_text_in_color "$ICyan" "Downloading new IPv4 dat file..."
             sleep 1
             check_command curl_to_dir "$GEOBLOCKDAT" "$IPV4_NAME" "$SCRIPTS"
-            check_command rm /usr/share/GeoIP/GeoIP.dat
+            mkdir -p /usr/share/GeoIP
+            rm -f /usr/share/GeoIP/GeoIP.dat
             check_command cp "$SCRIPTS/$IPV4_NAME" /usr/share/GeoIP
             check_command mv "/usr/share/GeoIP/$IPV4_NAME" /usr/share/GeoIP/GeoIP.dat
             chown root:root /usr/share/GeoIP/GeoIP.dat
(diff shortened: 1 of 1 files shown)


lib.sh
```diff
@@ -370,7 +370,8 @@ get_newest_dat_files() {
             print_text_in_color "$ICyan" "Downloading new IPv4 dat file..."
             sleep 1
             check_command curl_to_dir "$GEOBLOCKDAT" "$IPV4_NAME" "$SCRIPTS"
-            check_command rm /usr/share/GeoIP/GeoIP.dat
+            mkdir -p /usr/share/GeoIP
+            rm /usr/share/GeoIP/GeoIP.dat
```
enoch85, line 374: Please don't assume that the file is there, it will produce an error if not `-f` is used.

lib.sh
```diff
@@ -395,7 +396,8 @@ get_newest_dat_files() {
             print_text_in_color "$ICyan" "Downloading new IPv6 dat file..."
             sleep 1
             check_command curl_to_dir "$GEOBLOCKDAT" "$IPV6_NAME" "$SCRIPTS"
-            check_command rm /usr/share/GeoIP/GeoIPv6.dat
+            mkdir -p /usr/share/GeoIP
+            rm /usr/share/GeoIP/GeoIPv6.dat
```
enoch85, line 400: Same thing here. :)

enoch85:
Thans for your PR! :rocket: 

@szaimen What's your take on this?

szaimen:
@thorstenb The question is, why the files don't get created on your instance because they should by installing `libapache2-mod-geoip` 
https://github.com/nextcloud/vm/blob/bd560effac2c4711ae3090d33ee4ebf729ad3a13/network/geoblock.sh#L49

But I am fine with this change. 👍

thorstenb:
> @thorstenb The question is why the files don't get created on your instance because they should by installing `libapache2-mod-geoip`
> 

That's not installed, as geoip-database is only a recommended dependency for ibapache2-mod-geoip - could of course be added   explicitely in geoblock.sh, but what's the point if we're updating it anyway here?

szaimen:
@thorstenb okay, got it 👍
@enoch85 please go ahead an merge

PR #1732, opened 2020-12-19
Fix README typo deps->kernel->version

lnksz:
Ubuntu Server 20.04 ships with kernel 5.4.
(as of 2020.12 there isn't yet a 5.14 kernel)


enoch85:
Aaah true, thanks! :rocket:

PR #1740, opened 2020-12-28
fix sending watchtower removal notifications every time

### LINKED ISSUE #1739: Watchtower removed everytime I update
Every week, I get the message that Watchtower has been removed, due to the automatic updates being enabled. It doesn't seem like there is a check to see if it was actually removed or not and just posts the message regardless. 
https://github.com/nextcloud/vm/blob/50d5b0742d477a52b96b76cffb72c5e8787ce10f/nextcloud_update.sh#L323

szaimen:
fix #1739


EmilyLove26:
Thanks!

PR #1747, opened 2020-12-29
move clamav to the additional_apps menu

szaimen:
I am running ClamAV since a while now (a few weeks) on a test server and I haven't experienced any bugs, so IMO this is ready for the additional_apps menu. What do you think?


enoch85:
Sure, we can try it in the apps menu. :+1:

PR #1754, opened 2021-01-04
Move download scripts

enoch85:
This makes older versions of the VM work as well, even if locations are moved and scripts renamed.


szaimen (approved):
I like this 👍

PR #1757, opened 2021-01-07
change info

enoch85:
not needed anymore since we moved the download of scripts to part 1.

@szaimen Your take on this?


szaimen (approved):
Lgtm 👍

PR #1764, opened 2021-01-11
allow to cancel the restart if nextcloud_update.sh fails

szaimen:

### DIFF
--- a/addons/automatic_updates.sh
+++ b/addons/automatic_updates.sh
@@ -42,8 +42,8 @@ touch $VMLOGS/update.log
 crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/update.log 2>&1"; } | crontab -u root -
 if yesno_box_yes "Do you want to reboot your server after every update? *recommended*"
 then
-    sed -i "s|^exit$|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
-    echo "exit" >> "$SCRIPTS"/update.sh
+    sed -i "s|^#placeholder_for_automatic_updates$|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
+    echo "#placeholder_for_automatic_updates" >> "$SCRIPTS"/update.sh
 fi
 
 msg_box "Please remember to keep backups in case something goes wrong, as you never know."
--- a/addons/automatic_updates.sh
+++ b/addons/automatic_updates.sh
@@ -42,7 +42,7 @@ touch $VMLOGS/update.log
 crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/update.log 2>&1"; } | crontab -u root -
 if yesno_box_yes "Do you want to reboot your server after every update? *recommended*"
 then
-    sed -i "s|exit|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
+    sed -i "s|^exit$|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
     echo "exit" >> "$SCRIPTS"/update.sh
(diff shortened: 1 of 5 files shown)


enoch85:
Why this change?

szaimen:
Because it is not impossible that the first backup (borgbackup.sh or rsyncbackup.sh) still runs when the scheduled update starts. nextcloud_update.sh will skip the execution in this case but if you have automatic updates enabled with restart, it will execute the restart in this case nonetheless which is bad for the first backup since that will not run successfully to the end and will leave the backup in a bad state.

Also, apart from this, I think it makes sense to exit update.sh if something in nextcloud_update.sh fails (aka if an exitcode gets returned or the exitcode of nextcloud_update.sh is unequal to 0.)

szaimen:
@enoch85 Any comment on my comment?

enoch85:
I don't have time to think right now, sorry. :/

nextcloud_update.sh
```diff
@@ -154,6 +154,9 @@ then
     rm -f $SCRIPTS/lib.sh
 fi
 
+# Allow update.sh to exit if the nextcloud_update.sh fails (2020-01-11)
+sed -i 's/^run_script GITHUB_REPO nextcloud_update$/run_script GITHUB_REPO nextcloud_update || exit 1/' "$SCRIPTS/update.sh"
```
enoch85, line 158: Please explain this in more detail

enoch85:
> Because it is not impossible that the first backup (borgbackup.sh or rsyncbackup.sh) 

That means that we will change code to adjust for stuff that aren't officially supported? I don't know if I like that idea tbh.

nextcloud_update.sh
enoch85: I get it now. Sorry.

static/update.sh
```diff
@@ -30,6 +30,8 @@ then
 fi
 
 # Delete, download, run
-run_script GITHUB_REPO nextcloud_update
+run_script GITHUB_REPO nextcloud_update || exit 1
+
+#placeholder_for_automatic_updates
 
 exit
```
enoch85, line 37: dunno if this is needed?

addons/automatic_updates.sh
```diff
@@ -42,8 +42,8 @@ touch $VMLOGS/update.log
 crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/update.log 2>&1"; } | crontab -u root -
 if yesno_box_yes "Do you want to reboot your server after every update? *recommended*"
 then
-    sed -i "s|exit|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
-    echo "exit" >> "$SCRIPTS"/update.sh
+    sed -i "s|^#placeholder_for_automatic_updates$|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
```
szaimen, line 45: unfortunately this won't work in already existing installations...

static/update.sh
```diff
@@ -30,6 +30,8 @@ then
 fi
 
 # Delete, download, run
-run_script GITHUB_REPO nextcloud_update
+run_script GITHUB_REPO nextcloud_update || exit 1
+
+#placeholder_for_automatic_updates
```
szaimen, line 35: I don't think that approach is better than substituting exit

static/update.sh
szaimen: yah, it might not be needed but is the current placeholder. I don't think that introduction a placeholder string is the better approach.

szaimen:
> That means that we will change code to adjust for stuff that aren't officially supported? I don't know if I like that idea tbh.

Yah, I've already imagined that this could be a problem.

Let me adjust some things for now

> That means that we will change code to adjust for stuff that aren't officially supported? I don't know if I like that idea tbh.

I've already tought that this might be a problem...

szaimen:
this was superseded by #1778

PR #1771, opened 2021-01-15
fix race condition

enoch85:
if the path is set but the file doen't exist, it just keeps on searching, this fixes that issue.

### DIFF
--- a/apps/fail2ban.sh
+++ b/apps/fail2ban.sh
@@ -57,6 +57,24 @@ fi
 # Create $VMLOGS dir
 mkdir -p "$VMLOGS"
 
+find_log() {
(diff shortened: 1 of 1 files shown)


szaimen (approved):
makes sense

question: shouldn't we break the loop directly after executing`find_log`?

enoch85:
> question: shouldn't we break the loop directly after executingfind_log?

Hmm, might have over looked that.

Will do some testing soon.

enoch85:
> question: shouldn't we break the loop directly after executingfind_log?

Just checked, it's already in the function.

szaimen:
> Just checked, it's already in the function.

No, it isn't. Return doesn't break the loop, afaik.
The break must be outside of the function to work.

PR #1778, opened 2021-01-20
better fix for #1764

szaimen:
better fix for #1764

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -30,10 +30,19 @@ root_check
 is_process_running apt
 is_process_running dpkg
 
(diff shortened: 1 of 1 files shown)


szaimen:
This is ready :)

szaimen:
The advantage of this approach in comparison to #1764 is that this will only affect instances that manually opted in here:
https://github.com/nextcloud/vm/blob/1a0d3a584151a7921f677aaab5ea95a3393c3850/nextcloud_install_production.sh#L57-L72
So this will not affect any standard installation :)

enoch85:
Exactly, much better!

PR #1781, opened 2021-01-23
ddclient - migrate to docker 

szaimen:


enoch85:
Thanks for your efforts, but I'm not sure Docker is the way to go.

IMHO it's much better to have everything "bare bone". Docker just creates an extra layer of both complication and is more recourse intensive.

To me, Docker is just a hype. Sure, it's great and all that, but not for *everything*.

Please explain your reasoning behind this.

szaimen:
I guess I've come to the conclusion that docker is the way to go mainly because of this:
1. independence from the host OS
2. possibility of having all configs in one place

This results in a few advantages:
1. Easier to restore/backup
2. There is no risk of breaking things when tinkering with the host OS
3. being able to update with one command independent from the host
4. remove the complete package without any leftover
5. a few more I cannot remember currently

Ans as far as I've seen the overhead of docker isn't that big and worth it because of the advantages.

szaimen:
#### BTW: those are the htop stats for the ddclient and msmtpd containers during idle:

---

#### ddclient: image

---

#### msmtp: image

PR #1790, opened 2021-02-01
move pihole update to the update script

szaimen:
I didn't want to add this to the nextcloud_update.sh script but because of https://github.com/nextcloud/vm/commit/16b5a6f1b73f11826ec7ee36860c5030b49e596b there is no other way.


szaimen:
This is ready 👍

enoch85:
In that case, let's always remove the old update script without asking instead.

enoch85:
https://github.com/nextcloud/vm/commit/862e95d9e84e24c708a606e702d69593ee7dda50

szaimen:
Sry but I need to reopen this.
The problem isn't that you were asked if you want to download the newest update.sh script but that if you've chosen to do so it is removing the update procedure for pihole. Actually I think it is quite nice to ask to do so since there could be something manually added in there.

To fix this we can either merge this PR or remove the download of the latest update.sh from main_menu completely.

enoch85:
The Pi-Hole script isn't one of the offifical ones, and nothing that we will consider breaking changes for it to work.

Sorry, but you have to find another way.

szaimen:
This is superseded by #1797

PR #1796, opened 2021-02-04
ask to download the latest update script again

szaimen:
reverts https://github.com/nextcloud/vm/commit/862e95d9e84e24c708a606e702d69593ee7dda50
Because the UX was better before.
Note that this change is not related to the pihole script!


szaimen:
This is ready 👍

PR #1802, opened 2021-02-08
Azure-tuned kernel for Hyper-V based VMs

schories:
- issue: https://github.com/nextcloud/vm/issues/1803
- check (based on dmidecode and dmesg results) and install Azure-tuned kernel for Hyper-V based VMs (both, Windows on-prem and also in Azure)
- using Ubuntu meta package "linux-azure"
- to benefit from improved performance by using latest virtual Hardware Enablement (HWE) kernel for up to date Linux Integration Services
- feel free to adapt/edit/change the PR as you please


enoch85:
Hey, any update here? :)

enoch85:
@schories All checks are passing now, but I don't think we need to over complicate things. 

Now when reading the code properly I can see that it would probably be sufficient to add the install command under the existing check, do you agree?

enoch85:
The checks do the same thing, but in a simpler way. So let's just add it for installation if "Microsoft Corpoaration" is detected.

enoch85:
@schories Welcome as a contributor to the Nextcloud project! :rocket:

PR #1811, opened 2021-02-10
don't enable disabled apps after update

### LINKED ISSUE #1810: the update script still enables apps that were disabled before the update
@enoch85 Am I allowed to fix this?

to reproduce:
1. install NcVM
2. install a few external (not system) apps
3. disable them (not remove them)
4. update Nextcloud to a minor release using the update script
5. all external apps (also those that were disabled before) are enabled (which they shouldn't!)

szaimen:
fix #1810


szaimen:
This was successfully tested by updating a testinstance to NC21-beta8 hence marking as ready for review

szaimen:
@enoch85 if you don't understand how this works, please fell free to ask. As I've already said, it was successfully tested and works as expected now. It only needs your review.

nextcloud_update.sh
```diff
@@ -593,7 +593,7 @@ else
     declare -Ag APPSTORAGE
     for app in $NC_APPS
     do
-        APPSTORAGE[$app]=$(nextcloud_occ_no_check config:app:get "$app" enabled)
+        APPSTORAGE[$app]=$(nextcloud_occ_no_check config:app:get "$app" enabled >/dev/null 2>&1)
```
szaimen, line 596: @enoch85 this will break the whole process. Did you wanted to apply this fix ad a different place? (around line 777?)

enoch85: Please change

enoch85: (it worked for me)

szaimen: I've reverted this

lib.sh
```diff
@@ -1048,7 +1048,7 @@ fi
 
 #example: is_app_installed documentserver_community
 is_app_installed() {
-if nextcloud_occ app:list | grep -wq "$1"
+if nextcloud_occ app:list | awk '{print$2}' | tr -d ':' | sed '/^$/d' | grep -q "^$1$"
```
enoch85, line 1051: why is this better?

enoch85: I mean, now you use 4 tools instead of just one...

szaimen: you are correct. the old method seems to work just fine.
I will revert this

enoch85:
Okay, tested with the last fix and this seems to work as expected.

szaimen:
Just tested this, too. Looks good from my side, too 👍

enoch85:
Tested in PROD, does what it's supposed to.

PR #1812, opened 2021-02-11
fix a detail with the new app restore process

szaimen:
Sry for not integrating that earlier but as I found out, the app:disable command actually does more than just setting the database entry for that app to no: https://github.com/nextcloud/server/blob/f124cb0ef11fa15f21217552bedc6dd4a8ff9b5d/lib/private/App/AppManager.php#L397-L419
So I guess it is better to use the official command in that case.
Follow up to #1811

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -778,17 +778,26 @@ then
     print_text_in_color "$ICyan" "Restoring the status of apps. This can take a while..."
     for app in "${!APPSTORAGE[@]}"
     do
-        if ! is_app_installed "$app"
+        if [ -n "${APPSTORAGE[$app]}" ]
         then
-            nextcloud_occ_no_check app:install "$app"
-        fi
-        if [ -n "${APPSTORAGE[$app]}" ] && [ "${APPSTORAGE[$app]}" != "yes" ] && is_app_enabled "$app"
-        then
-            if [ "${APPSTORAGE[$app]}" = "no" ]
+            # Check if the app is in Nextclouds app storage
+            if ! [ -d "$NC_APPS_PATH/$app" ]
             then
-                nextcloud_occ_no_check app:disable "$app"
-            else
-                nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+                # If the app is missing from the apps folder and was installed and enabled before the upgrade was done,
+                # then reinstall it
+                if [ "${APPSTORAGE[$app]}" = "yes" ]
+                then
+                    install_and_enable_app "$app"
+                # If the app is missing from the apps folder and was installed but not enabled before the upgrade was done, 
+                # then reinstall it but keep it disabled
+                elif [ "${APPSTORAGE[$app]}" != "yes" ]
+                then
+                    if ! is_app_installed "$app"
+                    then
+                        nextcloud_occ_no_check app:install "$app"
+                        nextcloud_occ_no_check app:disable "$app"
+                    fi
+                fi
             fi
         fi
     done
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -790,10 +790,18 @@ then
                     install_and_enable_app "$app"
                 # If the app is missing from the apps folder and was installed but not enabled before the upgrade was done, 
                 # then reinstall it but keep it disabled
-                elif [ "${APPSTORAGE[$app]}" != "yes" ]
+                elif [ "${APPSTORAGE[$app]}" = "no" ]
                 then
                     install_and_enable_app "$app"
                     nextcloud_occ_no_check app:disable "$app"
+                # Cover the case where the app is enabled for certain groups
+                else
+                    install_and_enable_app "$app"
+                    if is_app_enabled "$app"
+                    then
+                        # Only restore the group settings, if the app was enabled (and is thus compatible with the new NC version)
+                        nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+                    fi
                 fi
             fi
             # If the app still isn't enabled (maybe because it's incompatible), then at least restore from backup,
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -784,7 +784,12 @@ then
         fi
         if [ -n "${APPSTORAGE[$app]}" ] && [ "${APPSTORAGE[$app]}" != "yes" ] && is_app_enabled "$app"
         then
-            nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+            if [ "${APPSTORAGE[$app]}" = "no" ]
+            then
+                nextcloud_occ_no_check app:disable "$app"
+            else
+                nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+            fi
         fi
     done
 fi
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -780,11 +780,16 @@ then
     do
         if [ -n "${APPSTORAGE[$app]}" ]
         then
+            # Check if the app is in Nextclouds app storage
             if ! [ -d "$NC_APPS_PATH/$app" ]
             then
+                # If the app is missing from the apps folder and was installed and enabled before the upgrade was done,
+                # then reinstall it
                 if [ "${APPSTORAGE[$app]}" = "yes" ]
                 then
                     install_and_enable_app "$app"
+                # If the app is missing from the apps folder and was installed but not enabled before the upgrade was done, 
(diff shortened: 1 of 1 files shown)


szaimen:
It looks like this now:
```
Config value enabled for app weather_status set to ["admin"]
Config value enabled for app files_videoplayer set to ["admin"]
tasks 0.13.6 installed
tasks enabled
tasks 0.13.6 disabled
Config value enabled for app user_status set to ["admin"]
news 15.3.2 installed
news enabled
news 15.3.2 disabled
Config value enabled for app files_pdfviewer set to ["admin"]
notes 4.0.2 installed
notes enabled
notes 4.0.2 disabled
calendar 2.1.3 installed
calendar enabled
calendar 2.1.3 disabled
Config value enabled for app logreader set to ["admin"]
Config value enabled for app dashboard set to ["admin"]
contacts 3.4.3 installed
contacts enabled
contacts 3.4.3 disabled
maps 0.1.8 installed
maps enabled
maps 0.1.8 disabled
```

enoch85:
```
maps 0.1.8 installed
maps enabled
maps 0.1.8 disabled
```

Huh? Why even enable the app if you're about to disable it afterwards?

szaimen:
> Huh? Why even enable the app if you're about to disable it afterwards?

That was actually the reason why I raised #1810 because before #1811 the app was automatically installed but also enabled if it was disabled before. (I just sometimes like to install an app to try it out and disable it afterwards only to see it automatically become enabled after an update)

enoch85:
Naah, I don't get this. There must be a better way. 

Apps should only be enabled if they were enabled *before* the upgrade, if they wasn't activated they shouldn't even be called...

Sure, we live in 2021 and have enough computing power to go to Mars, but let's not use it if it's not necessary. :)

szaimen:
So do you want me to change it so that apps that disabled before the update don't even get installed after the update?
I am not sure if I like this because I want to have the apps state like it was before the update but this might only my personal preference...

szaimen:
Ah, it could be that we actually don't understand each other here:
`occ app:install app` also enables the app by default (and prints out that it enables the app)
If I want to have it installed but disabled like before the update, I need to install it first and disable it via `occ app:disable app` after running `occ app:install app`... Thats why you see `install -> enable -> disable`

enoch85:
> I need to install it first (if it is not installed) and disable it via occ app:disable app after running occ app:install app

Why do you need to disable the app? s it just in your case or in general?

enoch85:
I mean, apps that aren't enabled before the upgrade shouldn't even be installed. That's the logic we should use.

szaimen:
hm... I guess I still vote for restoring the same state including disabled apps because what if you update to a major version and one app gets disabled because it is not yet compatible and you don't enable it afterwards manually? If you update to the next minor version it will then get removed because it wasn't enabled before the update... By doing it like I suggested, the app would at least still be listed in your apps list after the minor update (and you wouldn't loose it completely from your apps list).

enoch85:
If it's not compatible then there's no use in enabling it at all as it would probably break your Nextcloud instance, or cause some strange bugs.

Let me take another look at the code and understand it better.

enoch85:
I tested a bit and I think this logic is better:

```
# Restore app status
# Fixing https://github.com/nextcloud/server/issues/4538
if [ "${APPSTORAGE[0]}" != "no-export-done" ]
then
    print_text_in_color "$ICyan" "Restoring the status of apps. This can take a while..."
    for app in "${!APPSTORAGE[@]}"
    do
        if [ -n "${APPSTORAGE[$app]}" ]
        then
            if [ "${APPSTORAGE[$app]}" = "yes" ]
            then
                if ! [ -d "$NC_APPS_PATH/$app" ]
                then
                    install_and_enable_app "$app"
                fi
            elif [ "${APPSTORAGE[$app]}" != "yes" ]
            then
                if ! is_app_installed "$app"
                then
                    nextcloud_occ_no_check app:install "$app"
                    nextcloud_occ_no_check app:disable "$app"
                fi
            fi
        fi
    done
fi
```

1. If the app is disabled it will get installed but then disabled
2. If the app is missing it will get installed and enabled if it was installed and enabled before
3. Everything is based on the status before the status *before* the update was done.

Feel free to improve.

enoch85:
This is a scenario:
```

nextcloud_occ app:disable contacts

# Backup app status
# Fixing https://github.com/nextcloud/server/issues/4538
print_text_in_color "$ICyan" "Getting and backing up the status of apps for later, this might take a while..."
NC_APPS="$(nextcloud_occ app:list | awk '{print$2}' | tr -d ':' | sed '/^$/d')"
if [ -z "$NC_APPS" ]
then
    print_text_in_color "$IRed" "No apps detected, aborting export of app status... Please report this issue to $ISSUES"
    APPSTORAGE="no-export-done"
else
    declare -Ag APPSTORAGE
    for app in $NC_APPS
    do
        APPSTORAGE[$app]=$(nextcloud_occ_no_check config:app:get "$app" enabled)
    done
fi

nextcloud_occ app:remove calendar
rm -R $NC_APPS_PATH/contacts

# Restore app status
# Fixing https://github.com/nextcloud/server/issues/4538
if [ "${APPSTORAGE[0]}" != "no-export-done" ]
then
    print_text_in_color "$ICyan" "Restoring the status of apps. This can take a while..."
    for app in "${!APPSTORAGE[@]}"
    do
        if [ -n "${APPSTORAGE[$app]}" ]
        then
            if ! [ -d "$NC_APPS_PATH/$app" ]
            then
                if [ "${APPSTORAGE[$app]}" = "yes" ]
                then
                    install_and_enable_app "$app"
                elif [ "${APPSTORAGE[$app]}" != "yes" ]
                then
                    if ! is_app_installed "$app"
                    then
                        nextcloud_occ_no_check app:install "$app"
                        nextcloud_occ_no_check app:disable "$app"
                    fi
                fi
            fi
        fi
    done
fi
```



Output:
```
Updating lib...
No such app enabled: contacts
Getting and backing up the status of apps for later, this might take a while...
calendar disabled
calendar 2.1.3 removed
Installing calendar...
contacts 3.4.3 installed
contacts enabled
contacts 3.4.3 disabled
```

enoch85:
I'm happy with this, please test it. @szaimen

enoch85:
I was thinking we could put the last commit in the `install_and_enable_app` but for some reason I decided a while back to change the `$BACKUP` dir to `/var` instead of `/mnt` which means more code in the function. The "correct path" is set in the update script before this is run.

I'm thinking that we should move back to `/mnt` for backups maybe, since the OS disk is usually quite small. But that's for another PR.

nextcloud_update.sh
```diff
@@ -778,13 +778,33 @@ then
...
     for app in "${!APPSTORAGE[@]}"
     do
-        if ! is_app_installed "$app"
+        if [ -n "${APPSTORAGE[$app]}" ]
         then
-            nextcloud_occ_no_check app:install "$app"
-        fi
-        if [ -n "${APPSTORAGE[$app]}" ] && [ "${APPSTORAGE[$app]}" != "yes" ] && is_app_enabled "$app"
-        then
-            nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
```
szaimen, line 787: We cannot get rid of this because it would break https://github.com/nextcloud/server/issues/4538 again

szaimen:
Sry but I don't like your new changes.
It makes everything much more complicated than it was at the beginning when I've started the PR.
Also it break restoring the groups for which each app is enabled again. (nextcloud/server#4538)

szaimen:
Since I don't want to revert you changes, I've recreated the initial changes of this PR here:
#1813

enoch85:
OK, but this is the way to go for restoring apps IMHO.

We can add `nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"` if you need that for groups.

szaimen:
> OK, but this is the way to go for restoring apps IMHO.
> 
> We can add `nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"` if you need that for groups.

Okay, I've just added a case to restore the group settings.
This should now work as expected although I still think it is not nearly as sleek as #1813 😅

enoch85:
> not nearly as sleek as

Well, it's more obvious now, and I like that. Code should be easy to read as well. :)

szaimen:
Sry I've just seen that my implementation was wrong

enoch85:
Aah, merged the same time you said that. I tested and it work for me.

Please make another PR.

szaimen:
@enoch85 I now get those errors during the update:
image

enoch85:
> > not nearly as sleek as
> 
> Well, it's more obvious now, and I like that. Code should be easy to read as well. :)

Cc @szaimen 

https://twitter.com/PovilasKorop/status/1361265259483451394?s=19

PR #1813, opened 2021-02-12
fix a detail with the new app restore process

szaimen:
try 2
This is still much better than the latest changes to #1812. Since I don't want to revert your changes there, I've recreated the initial PR here.

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -784,7 +784,12 @@ then
         fi
         if [ -n "${APPSTORAGE[$app]}" ] && [ "${APPSTORAGE[$app]}" != "yes" ] && is_app_enabled "$app"
         then
-            nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+            if [ "${APPSTORAGE[$app]}" = "no" ]
+            then
+                nextcloud_occ_no_check app:disable "$app"
+            else
+                nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
+            fi
         fi
     done
 fi


szaimen:
@enoch85 can we please discuss what you don't like about this PR here?
I can make it work so that disabled apps before the update don't get reinstalled after the update (I don't like this, but if you prefer it that way, I can make it work).

BTW: this is just an enhancement to the implementation that was made in #1811 (we already reinstall disabled apps there, and also set it to disabled but don't use the official `app:disable` command there.)

nextcloud_update.sh
```diff
@@ -784,7 +784,12 @@ then
         fi
         if [ -n "${APPSTORAGE[$app]}" ] && [ "${APPSTORAGE[$app]}" != "yes" ] && is_app_enabled "$app"
         then
-            nextcloud_occ_no_check config:app:set "$app" enabled --value="${APPSTORAGE[$app]}"
```
szaimen, line 787: This is what I mean with: 
> (we already reinstall disabled apps there, and also set it to disabled but don't use the official app:disable command there.)

(initially we didn't differentiate between the value beeing set to an array of enabled groups or to `no`. This PR fixes this by using the official command to disable the app if it was disable before the update was run) But it already does kind of the same: reinstalling and disabling apps that were disabled before the update was executed

szaimen:
Quoting you from #1812:
> 1. If the app is disabled it will get installed but then disabled
> 2. If the app is missing it will get installed and enabled if it was installed and enabled before
> 3. Everything is based on the status before the status before the update was done.

This is exactly what this PR does!
But it additionally includes:
4. restoring the groups for which certain apps were enabled before!

szaimen:
Closing because @enoch85 wants to do it like he has proposed in #1812

PR #1815, opened 2021-02-12
fix restore group settings for apps

szaimen:
follow up to #1812

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -800,10 +800,14 @@ then
             # and make sure it's disabled
             if ! [ -d "$NC_APPS_PATH/$app" ] && [ -d "$BACKUP/apps/$app" ]
             then
-                print_text_in_color "$ICyan" "Restoring $app from $BACKUP/apps..."
-                rsync -Aaxz "$BACKUP/apps/$app" "$NC_APPS_PATH/"
-                bash "$SECURE"
-                nextcloud_occ_no_check app:disable "$app"
+                if yesno_box_no "$app couln't be enabled. Do you want to restore it from backup?\n\nWARNING: It may result in failed integrity checks."
+                then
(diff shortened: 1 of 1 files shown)


szaimen:
This was successfully tested.

enoch85:
With this fix, apps now gets disabled even if they were enabled before.

enoch85:
It may be my testing VM, but this shouldn't happen:

```
root@nextcloud:~# bash test.sh 
Updating lib...
No such app enabled: contacts
Getting and backing up the status of apps for later, this might take a while...
calendar disabled
calendar 2.1.3 removed
Config value enabled for app nextcloud_announcements set to Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
yes
Config value enabled for app firstrunwizard set to Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
yes
Config value enabled for app checksum set to Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
yes
Config value enabled for app settings set to Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
yes
Config value enabled for app encryption set to Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
Config value enabled for app nextbackup set to Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
yes
Config value enabled for app theming set to Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
yes
```

It happened when I restored an app from backup and then upgraded with `occ upgrade`. I've tried to unset both `APPSTORAGE` and `app` to clear cache, but it still happens.

PR #1816, opened 2021-02-13
geoblock - allow some scans

szaimen:
Allow scans from scan.nextcloud.com and observatory.mozilla.org
(63.245.208.0/24) are all owned by mozilla based on https://www.lookip.net/
fixes https://help.nextcloud.com/t/nextcloud-security-scan-no-error-but-scan-not-updating/103503/9

### DIFF
--- a/network/geoblock.sh
+++ b/network/geoblock.sh
@@ -181,6 +181,9 @@ GEOIP_CONF+="  Allow from env=AllowCountryOrContinent
   Allow from 192.168.0.0/16
   Allow from 172.16.0.0/12
   Allow from 10.0.0.0/8
+  Allow from scan.nextcloud.com
+  # Allow scans from observatory.mozilla.org:
+  Allow from 63.245.208.0/24
   Order Deny,Allow
   Deny from all
 </Location>


szaimen:
@enoch85 This was successfully tested but do you think that allowing 63.245.208.0/24 is too permissive?
I've until now only seen scans coming from 63.245.208.129 and 63.245.208.130 but it is definitely possible that other ip-addresses are used by observatory.mozilla.org, hence I've allowed 63.245.208.0/24 (since all appear do be owned by Mozilla based on https://www.lookip.net/)

enoch85:
> allowing 63.245.208.0/24 is too permissive?

It's your call. I don't use it, and I actually recommend against it if not done very precise. ¯\_(ツ)_/¯

szaimen:
> It's your call.

Then I decide to do it like this 👍

> I don't use it, and I actually recommend against it

Kindly asking why? :)

szaimen:
@enoch85?

enoch85:
> Why? :)

I get several support tickets that TLS can't be renewed for example, it almost always is due to country restrictions.

szaimen:
> I get several support tickets that TLS can't be renewed for example, it almost always is due to country restrictions.

Intersting! In that case we should maybe allow those ip addresses?

enoch85:
It's impossible to know, since they probably change all the time. I figure that LE's network is pretty huge. :)

PR #1822, opened 2021-02-17
geoblock - add lets encrypt advice

szaimen:
fix https://github.com/nextcloud/vm/pull/1816#issuecomment-779309359


enoch85:
Yeah, better than nothing. :)

PR #1824, opened 2021-02-17
switch to beta release channel if new major version was found

szaimen:
I think this makes sense since most of the time you won't see the new major update in https://your-nc-domain/settings/admin/overview unless you set it to the beta channel.
And as you know, this is no problem since the update script doesn't rely on this channel either way.


szaimen:
@enoch85 what you think about this?

enoch85:
> @enoch85 what you think about this?

Let's leave it as is. I know it doesn't affect anything in the VM, but you never know what users might think. Keep it *stable* :D

enoch85:
Please delete the branch, thanks.

PR #1817, opened 2021-02-13
Introduce notify_push in master

szaimen:
fix #1809
Signed-off-by: szaimen <szaimen@e.mail.de>

PR #1825, opened 2021-02-17
fix pecl php extensions

szaimen:
while testing #1817 I found out that all php extensions that are installed via pecl except smbclient weren't enabled when listing them via `php -m`. This fixes this.

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -264,7 +264,7 @@ fi
 # Remove old redis
 if grep -qFx extension=redis.so "$PHP_INI"
 then
-    sed -i "s|extension=redis.so||g" "$PHP_INI"
+    sed -i "/extension=redis.so/d" "$PHP_INI"
 fi
 # Check if redis is enabled and create the file if not
 if [ ! -f $PHP_MODS_DIR/redis.ini ]
@@ -291,7 +291,7 @@ then
             # Remove old igbinary
             if grep -qFx extension=igbinary.so "$PHP_INI"
             then
-                sed -i "s|extension=igbinary.so||g" "$PHP_INI"
+                sed -i "/extension=igbinary.so/d" "$PHP_INI"
             fi
             # Check if igbinary is enabled and create the file if not
             if [ ! -f $PHP_MODS_DIR/igbinary.ini ]
@@ -324,7 +324,7 @@ then
             # Remove old smbclient
             if grep -qFx extension=smbclient.so "$PHP_INI"
             then
-                sed -i "s|extension=smbclient.so||g" "$PHP_INI"
+                sed -i "/extension=smbclient.so/d" "$PHP_INI"
             fi
         fi
         if pecl list | grep -q apcu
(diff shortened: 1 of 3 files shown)


enoch85:
Please grep for "extension" everywhere since we already add it in PHP-FPM's `php.ini` I'm not sure we need this here.

szaimen:
Please see the output of php -m from an instance that was installed two weeks ago:
```
ncadmin@nc:~$ sudo -i
[sudo] password for ncadmin:
root@nc:~# php -m
[PHP Modules]
bcmath
bz2
calendar
Core
ctype
curl
date
dom
exif
FFI
fileinfo
filter
ftp
gd
gettext
gmp
hash
iconv
imap
intl
json
ldap
libsmbclient
libxml
mbstring
openssl
pcntl
pcre
PDO
pdo_pgsql
pdo_sqlite
pgsql
Phar
posix
readline
Reflection
session
shmop
SimpleXML
smbclient
soap
sockets
sodium
SPL
sqlite3
standard
sysvmsg
sysvsem
sysvshm
tokenizer
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
```
## Conclusion:
We need this because it doesn't work currently.
(I've already tested that it works when running `phpenmod -v ALL extension` (and creating the extension ini file before that)

enoch85:
So since we now are using separate .ini files for each module, then maybe we should remove the stuff in `php.ini` to avoid "already loaded in..." messages in `nextcloud.log`?

szaimen:
> So since we now are using separate .ini files for each module, then maybe we should remove the stuff in `php.ini` to avoid "already loaded in..." messages in `nextcloud.log`?

done 👍

szaimen:
BTW the update was already successfully tested on a testmachine.
When runnin the update now, I get no error in the nextcloud log.
Additionally this is the output from `php -m` after the update:
```
root@nc:~# php -m
[PHP Modules]
apcu
bcmath
bz2
calendar
Core
ctype
curl
date
dom
exif
FFI
fileinfo
filter
ftp
gd
gettext
gmp
hash
iconv
igbinary
imap
intl
json
ldap
libsmbclient
libxml
mbstring
openssl
pcntl
pcre
PDO
pdo_pgsql
pdo_sqlite
pgsql
Phar
posix
readline
redis
Reflection
session
shmop
SimpleXML
smbclient
soap
sockets
sodium
SPL
sqlite3
standard
sysvmsg
sysvsem
sysvshm
tokenizer
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
```
## Conclusion
As you can see: apcu, redis and igbinary are finnally successfully loaded :)

nextcloud_update.sh
```diff
@@ -261,11 +261,22 @@ then
     yes no | pecl upgrade redis
     systemctl restart redis-server.service
 fi
-
-# Double check if redis.so is enabled
-if ! grep -qFx extension=redis.so "$PHP_INI"
+# Remove old redis
+if grep -qFx extension=redis.so "$PHP_INI"
+then
+    sed -i "s|extension=redis.so||g" "$PHP_INI"
```
enoch85, line 267: Use delete here instead. This works but leaves a line in the file.

`> sed '/redis.so/d' file`

enoch85:
OK, it looks good, but I just want to double check here: have you really tested this inside and out?

I don't have time for testing but I'll trust you if you say it works 100%

szaimen:
> OK, it looks good, but I just want to double check here: have you really tested this inside and out?
> 
> I don't have time for testing but I'll trust you if you say it works 100%

I've tested that the update works and doesn't produce any new issue. So concerning the update script I am 100% sure that it works 👍

PR #1830, opened 2021-02-19
Enh/fix pecl php extensions

enoch85:
this should work


enoch85:
We discussed backwards compatibility many times. ;)

szaimen:
Thanks for the fix!
Obviously I forgot older systems for a second. I am sorry!

PR #1836, opened 2021-02-21
Move phone region setting

enoch85:
If it's not fixed here we still have another shot in the update script.

Per default, this option is chosen during the first setup, since we release with US keyboard config.


enoch85:
Hmm, naah it's better to source in the startup script instead.

PR #1839, opened 2021-02-21
fix the --not-latest flag

szaimen:

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -161,16 +161,22 @@ nc_update
 DEBUG=0
 debug_mode
 
-# Nextcloud 21 is required
-lowest_compatible_nc 21
+# Nextcloud 20 is required until 21.0.3 is out. Then NC 21 will be required.
+lowest_compatible_nc 20
 
 # Import if missing and export again to import it with UUID
 zpool_import_if_missing
 
 # Set phone region
-if [ -n "$KEYBOARD_LAYOUT" ]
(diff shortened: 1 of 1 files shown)


enoch85:
Sorry, but I'd rather have it "breaking" than someone not wanting to install the latest stuff. 

If you really do want to not use the latest, then just hash out the lowest compatible, and *then* have stuff breaking. :)

enoch85:
Thanks for remembering though.

szaimen:
I don't get what your problem with that change is because with this change the flag works again. 
Why removing that functionality?

enoch85:
> Why removing that functionality?

I can live with a "breaking" (it's only that function that breaks for like 2-3 point releases) change for a question that came up 1 time in 6 years to solve the real issue with NC 21 needing phone region and an updated lib file.

Sure, the alternative is to hammer the lib file and updated it all the time, but I'd rather change the structure of the script - require NC 21 and not needing to fetch the lib again.

If you really want to fix this, then lower to NC 20 for compatibility and add back the check for NC 21 in the phone region function. No need to fetch the lib again with the current structure.

I was thinking less code, but sure, if this can't be avoided, why not.

szaimen:
> If you really want to fix this, then lower to NC 20 for compatibility and add back the check for NC 21 in the phone region function.

Isn't that exactly what I've done here? 🧐

> I was thinking less code, but sure, if this can't be avoided, why not.

Yes, would be great if we could restore the --not-latest flag functionality since I recommend to use it :)

szaimen:
Thanks! :)

PR #1841, opened 2021-02-22
introduce whiteboard app in master

szaimen:

### DIFF
--- a/apps/previewgenerator.sh
+++ b/apps/previewgenerator.sh
@@ -48,7 +48,7 @@ else
     # reset the cronjob
     crontab -u www-data -l | grep -v 'preview:pre-generate'  | crontab -u www-data -
     # Remove apps
-    APPS=(ffmpeg php-imagick libmagickcore-6.q16-3-extra)
+    APPS=(php-imagick libmagickcore-6.q16-3-extra)
     for app in "${APPS[@]}"
     do
         if is_this_installed "$app"
@@ -56,6 +56,10 @@ else
             apt purge "$app" -y
         fi
     done
+    if is_this_installed ffmpeg && ! is_app_installed integration_whiteboard
+    then
+        apt purge ffmpeg -y
+    fi
     apt autoremove -y
     # Show successful uninstall if applicable
     removal_popup "$SCRIPT_NAME"
--- a/apps/whiteboard.sh
+++ b/apps/whiteboard.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+# T&M Hansson IT AB © - 2021, https://www.hanssonit.se/
+
+true
+SCRIPT_NAME="Whiteboard for Nextcloud"
+SCRIPT_EXPLAINER="$SCRIPT_NAME makes it possible to collaborative work on a whiteboard.
+It integrates Spacedeck whiteboard server and lets Nextcloud users create .whiteboard files which \
+can then be opened in the Files app and in Talk. Those files can be shared to other users or via \
+public links. Everyone having access with write permissions to such a file can edit it collaboratively."
+# shellcheck source=lib.sh
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
+# Check for errors + debug code and abort if something isn't right
+# 1 = ON
+# 0 = OFF
+DEBUG=0
+debug_mode
+
+# Check if root
+root_check
+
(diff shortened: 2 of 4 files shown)


szaimen:
I'd like to test this but unfortunately because of https://github.com/nextcloud/integration_whiteboard/issues/3 I cannot install the app.

szaimen:
The script was successfully tested by installing it on a ubuntu 20.04 testmachine with NC21, hence marking as ready for review.

apps/previewgenerator.sh
```diff
@@ -48,14 +48,18 @@ else
...
     for app in "${APPS[@]}"
     do
         if is_this_installed "$app"
         then
             apt purge "$app" -y
         fi
     done
+    if is_this_installed ffmpeg && ! is_app_installed integration_whiteboard
+    then
+        apt purge ffmpeg -y
```
enoch85, line 61: Hmm, I why does that package need to be removed? And why in previewgenerator?

szaimen: Because ffmpeg also gets installed by the previewgenerator script (becausee it is needed for Movie previews)

enoch85: Wouldn't this cause a conflict if both Preview Generator and Whiteboard are installed at the same time?

szaimen: No, I don't see the problem here:
If you uninstall previewgenerator and integration_whiteboard is installed, it doesn't remove ffmpeg.
If you uninstall whiteboard, and movie previews are enabled, it doesn't remove ffmpeg.

Or are you refering to something else?

enoch85: As I see it:

1. Before the change `ffmpeg` was always removed. Probably beacuse it doesn't work with previewgenerator(?)
2. Now, you only purge it if the whiteboard app isn't installed

Result of 2:
If the Whiiteborad app s installed, ffpmeg will not be removed thus previewgenerator will not work?

szaimen: Since ffmpeg is like a library, there is no problem if two apps are using it at the same time...

apps/previewgenerator.sh
```diff
@@ -48,14 +48,18 @@ else
     # reset the cronjob
     crontab -u www-data -l | grep -v 'preview:pre-generate'  | crontab -u www-data -
     # Remove apps
-    APPS=(ffmpeg php-imagick libmagickcore-6.q16-3-extra)
```
enoch85, line 51: 1. You remove it here, in any case, no matter if Whiteborad app is installed or not.

apps/previewgenerator.sh
szaimen: Now I get it :)
No, ffmpeg gets removed when reinstalling previewgenerator but it needed for movie previews here as well. So in conclusion: we don't remove it here because it is incompatible but because we reinstall the previewgenerator!

apps/previewgenerator.sh
```diff
@@ -48,14 +48,18 @@ else
...
+    APPS=(php-imagick libmagickcore-6.q16-3-extra)
     for app in "${APPS[@]}"
     do
         if is_this_installed "$app"
         then
             apt purge "$app" -y
         fi
     done
+    if is_this_installed ffmpeg && ! is_app_installed integration_whiteboard
```
enoch85, line 59: 2. Here you *only* remove it if the Whiteboard app isn't installed.

Why? Doesn't Whiteboard app work without `ffmpeg`?

enoch85:
Please solve conflicts. Thanks!

apps/previewgenerator.sh
enoch85: OK, thanks!

Please solve the conflicts and I'll merge.

apps/previewgenerator.sh
szaimen: yes, the whiteboard app needs ffmpeg to work correctly, so this is correct👍

apps/previewgenerator.sh
enoch85: (I was lazy and didn't read everything) :man_technologist:

apps/previewgenerator.sh
szaimen: No, I've implemented a check for ffmpeg which checks if the whiteboard app is installed so that it doesn't remove ffmpeg in this case as you can see below.

szaimen:
> Please solve conflicts. Thanks!

done 👍

PR #1849, opened 2021-02-23
move extract to its own script

### LINKED ISSUE #1842: Move extract to its own script because of the dependencies?
Hey,

I wonder if we should move the extract app to its own script because of the dependencies?
So that you are able to install and uninstall it any time?
https://github.com/nextcloud/vm/blob/4dc4c79320f3cff212b3a6ecf8809067d0a69cfd/nextcloud_install_production.sh#L871-L877

szaimen:
fix #1842

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -848,6 +848,7 @@ $CHECKLIST_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Contacts" "" ON \
 "IssueTemplate" "" ON \
 "PDFViewer" "" ON \
+"Extract" "" ON \
 "Text" "" ON \
 "Mail" "" ON \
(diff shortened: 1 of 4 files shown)


nextcloud_install_production.sh
```diff
@@ -848,7 +848,6 @@ $CHECKLIST_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Contacts" "" ON \
 "IssueTemplate" "" ON \
 "PDFViewer" "" ON \
-"Extract" "" ON \
```
enoch85, line 851: It should still be default, of someone doesn't like it they can uninstall it with the menu.

szaimen: done

enoch85:
Thanks! :1st_place_medal:

PR #1850, opened 2021-02-24
make sure that all docker containers only listen on localhost

szaimen:


szaimen:

PR #1860, opened 2021-02-28
smtp-mail - make it possible to use the same settings in Nextcloud

szaimen:


szaimen:

PR #1862, opened 2021-02-28
add 2FA via Nextcloud Notification

szaimen:


szaimen:
This is ready 👍

szaimen:
hm... seems like it is not yet completely compatible with NC21
https://github.com/nickv-nextcloud/twofactor_nextcloud_notification/issues/474

enoch85:
Please let me know when it is and we'll merge it as an startup-script app app, not as a default app.

szaimen:
> and we'll merge it as an startup-script app app, not as a default app.

Where do I add it then, in which script at which place?

enoch85:
> in which script at which place?

I think in Nextcloud configuration as a seperate script or config in the if statement.

szaimen:
This PR itself was successfully tested. 
Unfortunately there is still no new release of the app for NC21 but it will work with NC20.
So I guess it is up to you to decide to merge this now or in a few days.

enoch85:
> no new release of the app for NC21 but it will work with NC20.

`lowest_compatible_nc 21`

szaimen:
> > no new release of the app for NC21 but it will work with NC20.
> 
> `lowest_compatible_nc 21`

Actually, it would be the opposite in this case 😉
`highest_compatible_nc 20`

enoch85:
> highest_compatible_nc 20 True :)

szaimen:
A new release with the fix for NC21 of the 2FA Nextcloud notification app was just created.
So this is now finally ready to go :)
(if you want to test it on NC21: it is possible that the new release isn't registered on all instances because of the appstore caching)

enoch85:
> (if you want to test it on NC21: it is possible that the new release isn't registered on all instances because of the appstore caching)

Let's wait a few days. No rush. :)

szaimen:
This was successfully tested on a NC21 instance 👍

enoch85:
But wait, just read the script again, this is just an app. Also it needs QR-codes and backup codes to be made, which we can't provide in the CLI "GUI". 

Isn't it better if they use the actual App Store for this?

I mean, we should only list apps that needs some extra CLI commands IMHO.

szaimen:
> Also it needs QR-codes

No that is the 2fa totp app.

> I mean, we should only list apps that needs some extra CLI commands IMHO.

Yes, you have a point 👍

PR #1870, opened 2021-03-02
add pdfdraw

szaimen:
Needed:
- [x] ~~Answer to https://github.com/strukturag/pdfdraw/issues/27~~ Just copy the server directory to another place upon installation.
- [x] Fix to https://github.com/strukturag/pdfdraw/issues/23
- [ ] Wait for a new release after https://github.com/strukturag/pdfdraw/issues/23 got fixed (then remove the TODO lines)
- [x] ~~Release for NC21 https://github.com/strukturag/pdfdraw/issues/21~~ No longer needed since it will only be available in the not-supported menu for now
- [x] ~~Get rid of npm as dependency? https://github.com/strukturag/pdfdraw/pull/28~~ Just uninstall it after running npm install.

### DIFF
--- a/apps/pdfdraw.sh
+++ b/apps/pdfdraw.sh
@@ -104,7 +104,7 @@ fi
 # Install dependencies
 # https://github.com/strukturag/pdfdraw/tree/master/server#requirements
 install_if_not nodejs
-install_if_not npm
+install_if_not npm # TODO: remove this in case https://github.com/strukturag/pdfdraw/pull/28 get merged
 install_if_not pdftk
 install_if_not python2
 install_if_not python-pypdf2
--- a/apps/pdfdraw.sh
+++ b/apps/pdfdraw.sh
@@ -71,7 +71,7 @@ else
     rm -f "$SERVICE_PATH"
     pip uninstall svglib -y &>/dev/null
     python2 -m pip uninstall pip -y &>/dev/null
-    for packet in nodejs npm pdftk pdftk-java python2-pip python-pypdf2
+    for packet in nodejs npm pdftk python-pypdf2
     do
         if is_this_installed "$packet"
         then
--- a/apps/pdfdraw.sh
+++ b/apps/pdfdraw.sh
@@ -1,0 +1,179 @@
+#!/bin/bash
+
+# T&M Hansson IT AB © - 2021, https://www.hanssonit.se/
+# Copyright © 2021 Simon Lindner (https://github.com/szaimen)
+
+true
+SCRIPT_NAME="PDF Annotations"
+SCRIPT_EXPLAINER="This script allows to easily install PDF Annotations, \
+a tool to annotate any PDF document collaboratively inside Nextcloud"
+# shellcheck source=lib.sh
+source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+
+# Check for errors + debug code and abort if something isn't right
+# 1 = ON
+# 0 = OFF
+DEBUG=0
+debug_mode
+
+# Check if root
+root_check
+
+# Variables
+SERVICE_PATH="/etc/systemd/system/pdfdraw.service"
+
+# Check requirements
+if is_app_enabled files_external || is_app_enabled groupfolders
+then
+    msg_box "The app is unfortunately not yet compatible with Nextcloud's external storage and groupfolders. \
+Cannot proceed!\n
+(The download of changed pdf files doesn't work.)
+You can track this here: 'https://github.com/strukturag/pdfdraw/issues/25'"
+    exit 1
+fi
+# Nextcloud Main Domain
+NCDOMAIN=$(nextcloud_occ_no_check config:system:get overwrite.cli.url | sed 's|https://||;s|/||')
+# Check if Nextcloud is installed
+print_text_in_color "$ICyan" "Checking if Nextcloud is installed..."
+if ! curl -s https://"$NCDOMAIN"/status.php | grep -q 'installed":true'
+then
+    msg_box "It seems like Nextcloud is not installed or that you don't use https on:
+$NCDOMAIN
+Please install Nextcloud and make sure your domain is reachable, or activate TLS \
+on your domain to be able to run this script.
+If you use the Nextcloud VM you can use the Let's Encrypt script to get TLS and activate your Nextcloud domain."
+    exit 1
+fi
+# Check apache conf
+if ! [ -f "$SITES_AVAILABLE/$NCDOMAIN.conf" ]
+then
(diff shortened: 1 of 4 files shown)


enoch85:
Sure adding aops is great, but the list is getting very long now.

Please be conservative. Is this really needed?

szaimen:
Yes, being able to annotate PDF files is a very desirable feature, IMO :)

szaimen:
I guess this is because of the many dependencies something for the not-supported folder for now.

ASerbinski:
Hi @szaimen 

Looks like a pretty ugly and complicated mess to get this added to your VM, I guess because of the need to include their back end server.

Their back end server actually doesn't have much in the way of useful functionality, rather it adds a lot of complexity and kills performance.

Consequently, I've been working on a related project. While I do borrow front end code, the differences are radical enough that it wouldn't be entirely fair to call it a fork.

https://gitlab.com/nextcloud-other/nextcloud-annotate

The differences:
1) the backend server is NOT required.
2) If two users annotate the same file at the same time, you won't see each other's cursors on your screen. I can't imagine why that would be useful actually.
3) instead of downloading the annotated PDF, it saves it back to nextcloud. Older versions are available through versions.
4) I fixed the problem with rotated pdf's not being correctly annotated.
5) its usable on small screen devices like phones and tablets. Very powerful to use on a tablet -- draw with a stylus!
6) very simple to install, just clone it into nextcloud/apps/pdfannotate/ and enable!

szaimen:
Hi @ASerbinski, thanks for your contribution! Your app looks really promising!
I am willing to create a new PR using your app if you could publish it to the apstore, would that be possible? :)
Also, I am sure that a lot of people would like to use it and it makes it much easier and better to find if they can install it via the appstore.

Also there are some things that I've notied while trying your app (but aren't a real deal breakers, imo):
- I've seen a RAM spike of about 1GB during saving (I am not sure if it was on the VM or the Client itself but I've noticed it)
- On mobile: I cannot scroll down
- Maybe it would make sense to change the download button to show that it saves back to nextcloud?
- Maybe the close button should show a warning when clicked that unsaved changes don't get saved?

ASerbinski:
I don't find "app stores" to be particularly compelling. You can feel free to publish it there if you wish, don't let me stand in the way of that, however, I'd suggest that you hold off until the dependency on svg2pdf and pdftk can be eliminated, since those installing from stores may not anticipate or be in a position to address them. You will note that at the moment, I'm still using the same approach for generating the pdf as pdfdraw uses.

Scrolling is a challenge that I should be able to address at some point, however it CAN be scrolled on mobile, you just have to put your finger down where it is not on the page. Zooming out can make this easier by increasing the amount of space beside the page.

Graphics... definitely a lower priority than functionality, but point noted.

Memory... would certainly be due to svg2pdf and/or pdftk. Processing graphics can be extremely memory intensive, so this doesn't surprise me. This also wouldn't be any different from pdfdraw.

It would probably be best to continue this discussion on my issue tracker for the project; https://gitlab.com/nextcloud-other/nextcloud-annotate/-/issues

szaimen:
Normally, we wouldn't consider any app that isn't published to the appstore but since it is that much better than the original app, I am going to make an exception here.

For the rest see https://gitlab.com/nextcloud-other/nextcloud-annotate/-/issues/1

I am going to close this PR and create a completely new one.

enoch85:
@ASerbinski 

> I don't find "app stores" to be particularly compelling.

If you want us to use your app in the VM (and get a super wide spread for thousands of users) then you might rethink your decision.

As @szaimen mentioned, the app store exists and is there for a reason. To us it's a receipt that it's trusted and would make it possible for us to check compatibility against the current Nextcloud version. Sure, it's a bit of maintenance, but everything comes with a cost.

I hope that we can find common ground. :+1:

ASerbinski:
Common ground would imply that our goals are tied together, when they actually are not. In fact, I work on the project for my own use above all else, but figured that if it would be beneficial to someone else, I would happily offer it. In fact, I don't need the burden, nor exposure to "thousands of users". If you absolutely require it to be offered through nextcloud's mechanisms, then that burden should fall on you or someone else who wants to do that.

enoch85:
@ASerbinski Hmm, I'm not sure I agree with your mindset and egoistic way of thinking.

I don't want to start a super long discussion here, but your answer made me decide - we don't add this.

szaimen:
> egoistic way of thinking.

I don't think that it is egoistig if he publishes this app for free. Also I am pretty sure that this statement won't convince him to publish the app to the app store.
In my opinion is the app still a viable and great option. Hence let's see if he or maybe somebody else is willing to publish the app in the app store.

---

@ASerbinski would you accept co-maintainers for the repository that will be able to create app-releases and publish the app to the store? I don't think that it is possible without that. If you aggree, I could start a new thread on https://helpt.nextcloud.com to look for any volunteer that could help you in doing this.

enoch85:
Sure, @ASerbinski maintains a repo for *his own* purposes for free.  But then he comes here and tell us that his way is much better, without being willing to actually contribute for others sake, only for himself. To me, that's egoistic and not the way most of the Nextcloud community is. 

> In fact, I don't need the burden

> ... then that burden should fall on you or someone else who wants to do that.

That explains a lot.

I understand though, maitaining an app means keeping up with changes in the server core, which I'm sure @ASerbinski doesn't care about. That in turn makes it hard for me to trust it.

ASerbinski:
You know something, before you start throwing around insults, you really should take a good long look in the mirror and ask yourself it the insult you are throwing might actually be a better fit for yourself. Egoist? The behavior you have exhibited is a very strong example of that. Do you actually believe that everyone who happens to have something of value should bend over backwards to comply with your needs above their own? Do you think that my *clients* needs should be put on the backburner so that I can worship you?

Let me make this very clear: I will not be bullied into complying with your demands. If there was any sliver of a chance that I might be willing to go through the song and dance that you say is needed for this to be included, then your attitude has made certain that it has been destroyed.

@szaimen ; I appreciate your level attitude and open mindedness. I am sorry that it looks like we won't be able to collaborate on this.

szaimen:
> Let me make this very clear: I will not be bullied into complying with your demands. If there was any sliver of a chance that I might be willing to go through the song and dance that you say is needed for this to be included, then your attitude has made certain that it has been destroyed.

That's comprehensible.

> @szaimen ; I appreciate your level attitude and open mindedness. I am sorry that it looks like we won't be able to collaborate on this.

Thanks for the kind words and yes, that's too bad!

szaimen:
> I can talk with people to make it happen if you decide you want to be an app maintainer.

@enoch85 Thanks for the idea but this was never an option for me!

enoch85:
@ASerbinski Let's agree to disagree then. :+1:

PR #1880, opened 2021-03-09
allow updating to next major version

enoch85:


enoch85:
@szaimen this was quick and dirty just to solve a urgent matter, but we should have a look at the NCBAD value in update.sh, I think it's wrong.

szaimen:
Hm... I wonder why make it optional and not the default when running the udpate script now that such an option exists?

PR #1882, opened 2021-03-09
fix versioning - take 2

enoch85:


enoch85:
@szaimen FYI, I think this will work

PR #1888, opened 2021-03-12
add some more info and fix typos

enoch85:
addons/desec.sh|45 col 96| "adress" is a misspelling of "address"
addons/desec.sh|47 col 87| "suport" is a misspelling of "support"
menu/nextcloud_configuration.sh|51 col 52| "behaviour" is a misspelling of "behavior"
menu/startup_configuration.sh|18 col 68| "completley" is a misspelling of "completely"
CHANGELOG.md|201 col 32| "occurances" is a misspelling of "occurrences"
issue_template.md|17 col 12| "behaviour" is a misspelling of "behavior"
issue_template.md|20 col 10| "behaviour" is a misspelling of "behavior"
apps/previewgenerator.sh|241 col 12| "behaviour" is a misspelling of "behavior"

### DIFF
--- a/addons/desec.sh
+++ b/addons/desec.sh
@@ -26,7 +26,7 @@ The only allowed characters for the username are:
         # Check for SOA record
         if host -t SOA "$DEDYNDOMAIN" >/dev/null 2>&1
         then
-            if ! yesno_box_yes "Sorry, but it seems like $DEDYNDOMAIN is taken. Do you want to try again?"
+            if ! yesno_box_yes "Sorry, but it seems like $DEDYNDOMAIN(.dedyn.io) is taken. Do you want to try again?"
             then
                exit
             fi
@@ -81,6 +81,7 @@ do
         fi
     else
        break
+       msg_box "$DEDYNDOMAIN(.dedyn.io) was sucessfully setup! Now please continue with the DDNS and TLS setup for the domain."
     fi
 done
 
--- a/addons/desec.sh
+++ b/addons/desec.sh
@@ -42,9 +42,9 @@ msg_box "You will now be prompted to enter an email address. It's very important
 
 Every 6 months you will get an email asking you to confirm your domain. If you don't react within a few weeks, your domain will be destroyed!"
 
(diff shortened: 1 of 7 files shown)


addons/desec.sh
```diff
@@ -81,6 +81,7 @@ do
         fi
     else
        break
+       msg_box "$DEDYNDOMAIN was successfully setup with deSEC! Now please continue with the DDNS and TLS setup for the domain."
```
szaimen, line 84: @enoch85 this msg_box will most likely never be shown because it is after the break keyword. (to get shown it must be before the keyword)

szaimen: Also, I think the word `setup` isn't correct here because it is a verb. And the verb for `setup` is `to set up`, afaik.

menu/server_configuration.sh
```diff
@@ -48,7 +48,7 @@ choice=$(whiptail --title "$TITLE" --checklist \
 $CHECKLIST_GUIDE\n\n$RUN_LATER_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "Static IP" "(Set static IP in Ubuntu with netplan.io)" OFF \
 "Security" "(Add extra security based on this http://goo.gl/gEJHi7)" OFF \
-"deSEC" "(Automatically get a domain from desec.io, configure ddclient and activate TLS)" "$STARTUP_SWITCH" \
+"deSEC" "(Automatically setup a dedyn.io domain, together with DDNS and TLS)" "$STARTUP_SWITCH" \
```
szaimen, line 51: Here the same for the word `setup` -> should be `set up`, imo

enoch85: Great feeback, please change to set up 

https://writingexplained.org/setup-vs-set-up-difference

addons/desec.sh
enoch85: actually, I was tired yesterday. ofc it should be after the break.

https://www.youtube.com/watch?v=eDMv3IHGpnA

PR #1893, opened 2021-03-15
improve https check

enoch85:


apps/notify_push.sh
```diff
@@ -50,13 +50,9 @@ If you use the Nextcloud VM you can use the Let's Encrypt script to get TLS and
...
 # Check apache conf
-if ! [ -f "$SITES_AVAILABLE/$NCDOMAIN.conf" ]
+if ! check_nextcloud_https "Notify Push"
 then
-    msg_box "The apache conf for $NCDOMAIN isn't available. This is not supported!"
-    exit 1
-elif ! grep -q "<VirtualHost \*:443>" "$SITES_AVAILABLE/$NCDOMAIN.conf"
-then
-    msg_box "The virtualhost config doesn't seem to be the default. Cannot proceed."
+    msg_box "Nextcloud doesn't seem to run on HTTPS. This is not supported!"
```
szaimen, line 55: Unfortunately, this check isn't enough for Notify Push because it will most likely fail the installation if not both checks (file and default configuration) are fullfilled. So this isn't going to work, unfortunately.
https://github.com/nextcloud/vm/blob/107f547d8b92cea5405eaa14d880f8f340f0f632/apps/notify_push.sh#L138

enoch85: I tested this, and it only checks if the VirtualHost config contains 443, right?

szaimen: The sed command that I've linked above inserts the needed code after the pattern match `<VirtualHost *:443>` into the `$SITES_AVAILABLE/$NCDOMAIN.conf` file. Without this, notify push will not work.

enoch85: OK, got it.

PR #1894, opened 2021-03-16
smart monitoring - test drives directly

szaimen:

### DIFF
--- a/disk/smart-monitoring.sh
+++ b/disk/smart-monitoring.sh
@@ -86,6 +86,7 @@ done
 if [ -z "$VALID_DRIVES" ]
 then
     msg_box "It seems like not even one drive supports smart monitoring/is healthy.
+This is completely normal if you run this script in a VM since virtual drives don't support smart monitoring.
 We will uninstall smart monitoring now since you won't get any helpful notification out of this going forward."
     apt purge smartmontools -y
     apt autoremove -y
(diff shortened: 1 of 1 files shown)


szaimen:

disk/smart-monitoring.sh
```diff
@@ -54,6 +62,37 @@ fi
...
+# Test drives
+print_text_in_color "$ICyan" "Testing if all drives support smart monitoring and are healthy..."
+mapfile -t DRIVES <<< "$DRIVES"
+for drive in "${DRIVES[@]}"
+do
+    echo '#########################'
+    print_text_in_color "$ICyan" "Testing /dev/$drive"
+    if ! smartctl -a "/dev/$drive"
+    then
+        print_text_in_color "$IRed" "/dev/$drive doesn't support smart monitoring or isn't healthy"
```
enoch85, line 75: is this intended? first print then msg_box?

szaimen: yes, for the output log of the script this makes it much more readable. But I also think that showing the msg_box here makes sense -> I do both

szaimen: I could also reverse the order -> show msg_box first and then print_text_in_color. What do you think?

enoch85: Whatever. Doesn't matter to me.

enoch85:
So, what does this do exactly?

It seems to test the disks somehow?

szaimen:
> So, what does this do exactly?
> 
> It seems to test the disks somehow?

Yes, it tests all connected drives directly and reports if a drive is unhealthy/doesn't support smart monitoring so that you know directly what is up with your drives :)

PR #1901, opened 2021-03-21
improve initial smart testing

szaimen:
Follow-up to #1894


szaimen:
This is a tremendous improvement over #1894 because it is now able to differentiate between smart support and health problems, only prints out the smart output if a drive doesn't support smart monitoring or is unhealthy and was successfully tested on a testmachine and VM, both with Ubuntu 20.04 👍

PR #1904, opened 2021-03-26
fix letsencryptrenew for all geoblock configurations

szaimen:

### DIFF
--- a/apps/bitwarden-rs.sh
+++ b/apps/bitwarden-rs.sh
@@ -200,6 +200,9 @@ else
     exit 1
 fi
 
+# Add Let's Encrypt! crontab
+add_crontab_le
+
 # Install docker
 install_docker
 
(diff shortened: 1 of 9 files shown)


szaimen:
This was successfully tested and is ready for review 👍

szaimen:
One change is still needed

enoch85:
Have you tested this by setting up a VPS and installed Nextcloud from the beginning, and then tried to activate TLS as usual?

I remember that the crontab wouldn't be set if run in a function, hence it's directly in the script.

szaimen:
I've just tested renewing and getting a certificate without allowing any country or continent and it worked. -> hence this PR isn't needed

enoch85:
Please add notify_admin_gui on line 66 if apache doesn't start.

Otherwise looks good, but not tested.

szaimen:
Do you mean in #1905?

enoch85:
Yes, sorry.

szaimen:
I think I've fixed it with https://github.com/nextcloud/vm/pull/1905/commits/3d435e4cff1c7afd97422592f6ca63379fa3eaf3

PR #1909, opened 2021-03-28
update clamav

szaimen:


apps/clamav.sh
```diff
@@ -287,7 +286,7 @@ chown clamav:clamav "$VMLOGS"/clamav-fullscan.log
 
 # Inform the user
 msg_box "The full scan was successfully setup.
-It will run on Sundays starting at 10:00 and will continue for a maximum of 12 hours (hardcoded).
+It will run on Sundays starting at 10:00.
```
szaimen, line 289: Unfortunately is the max-scantime option only for single files to prevent zip-bomb attacks as far as I found out. So it doesn't kill the scan after 12h...

PR #1912, opened 2021-03-31
speed up the network check if the network already works

szaimen:
This speeds up the initial network check of main_menu and other scripts by a lot.
If the network already works we don't need to reapply the netplan settings, do we?


szaimen:
Or maybe I should ask the other way around: am I correct that `netplan apply` is needed during the startup script?

enoch85:
I need to test this when I have time. I know I didn't want to `return` for a reason but I can't remember which reason it was.

enoch85:
Tested and works!

PR #1913, opened 2021-04-01
improve the built-in backup solution to work with normal NcVM configurations

### LINKED ISSUE #1910: improve the built-in backup solution to work with normal NcVM configurations?
Hey @enoch85 I wonder if I should invest the time to make the backup scripts compatible with the normal NcVM configuration?

What would be needed is the following:

0. Improve wording that you wont be able to use the built-in bavkup solution it you choose to not create the partition
1. make this partition to the default in any new installation: https://github.com/nextcloud/vm/blob/c6fab8acab4cbf1cfa8ffb2c13d5cc2627e50eae/nextcloud_install_production.sh#L67 
2. add a new option to modify a smbmount to be a backup mount here: https://github.com/nextcloud/vm/blob/c6fab8acab4cbf1cfa8ffb2c13d5cc2627e50eae/apps/smbmount.sh#L133-L134
3. allow those kind of mounts to be viable for the backup wizard script here: https://github.com/nextcloud/vm/blob/c6fab8acab4cbf1cfa8ffb2c13d5cc2627e50eae/not-supported/daily-backup-wizard.sh#L69-L70 (and don't fail if /mnt/ncdata is in mounttab) (don't allow custom mounts: can cause unexpectable problems)
4. modify the borgbackup script to create dedicated backups of the /mnt/ncdata partition
5. modify the system-restore script to be able to restore also the /mnt/ncdata partition from those backups
6. optional: modify the restore-backup script to allow to restore the /mnt/ncdata partition, too (also allow NcVM reserved in this case to not be present) (and don't check for /mnt/ncdata in the mounttab this case) also allow to use smbmount backup mounts without having to connect the device

Would you accept the changes that would be necessary? (especially point 1 is of importance)

szaimen:
As the title says
fix #1910

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -38,13 +38,21 @@ It is possible that a backup is currently running.\n
 Advice: don't restart your system now if that is the case!\n
 If you are sure that no backup is currently running, you can fix this by executing:
 'sudo lvrename /dev/ubuntu-vg/NcVM-snapshot-pending /dev/ubuntu-vg/NcVM-snapshot'"
-    # Don't exit the script while the snapshot exists to disable any automatic restart during backups
-    while does_snapshot_exist "NcVM-snapshot-pending"
-    do
-        print_text_in_color "$ICyan" "Waiting for NcVM-snapshot-pending to vanish... (press '[CTRL] + [C]' to cancel)"
-        sleep 5
-    done
-    print_text_in_color "$ICyan" "Continuing the update..."
+    # Kill all "$SCRIPTS/update.sh" processes to make sure that no automatic restart happens after exiting this script
+    PROCESS_IDS=$(ps aux | grep "$SCRIPTS/update.sh" | grep -v grep | awk '{print $2}')
+    if [ -n "$PROCESS_IDS" ]
+    then
+        mapfile -t PROCESS_IDS <<< "$PROCESS_IDS"
+        for process in "${PROCESS_IDS[@]}"
+        do
+            print_text_in_color "$ICyan" "Killing the process with PID $process to prevent a potential automatic restart..."
+            if ! kill "$process"
+            then
+                print_text_in_color "$IRed" "Couldn't kill the process with PID $process..."
+            fi
+        done
+    fi
+    exit 1
 fi
 
 # Create a snapshot before doing anything else
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -39,6 +39,7 @@ Advice: don't restart your system now if that is the case!\n
 If you are sure that no backup is currently running, you can fix this by executing:
 'sudo lvrename /dev/ubuntu-vg/NcVM-snapshot-pending /dev/ubuntu-vg/NcVM-snapshot'"
     # Kill all "$SCRIPTS/update.sh" processes to make sure that no automatic restart happens after exiting this script
+    # shellcheck disable=2009
     PROCESS_IDS=$(ps aux | grep "$SCRIPTS/update.sh" | grep -v grep | awk '{print $2}')
     if [ -n "$PROCESS_IDS" ]
     then
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -35,7 +35,9 @@ if does_snapshot_exist "NcVM-snapshot-pending"
 then
     msg_box "Cannot proceed with the update currently because NcVM-snapshot-pending exists.\n
 It is possible that a backup is currently running.\n
-Advice: don't restart your system now if that is the case!"
+Advice: don't restart your system now if that is the case!\n
+If you are sure that not backup is currently running, you can fix this by executing:
+'sudo lvrename /dev/ubuntu-vg/NcVM-snapshot-pending /dev/ubuntu-vg/NcVM-snapshot'"
     # Don't exit the script while the snapshot exists to disable any automatic restart during backups
     while does_snapshot_exist "NcVM-snapshot-pending"
     do
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -36,7 +36,7 @@ then
     msg_box "Cannot proceed with the update currently because NcVM-snapshot-pending exists.\n
 It is possible that a backup is currently running.\n
 Advice: don't restart your system now if that is the case!\n
-If you are sure that not backup is currently running, you can fix this by executing:
+If you are sure that no backup is currently running, you can fix this by executing:
 'sudo lvrename /dev/ubuntu-vg/NcVM-snapshot-pending /dev/ubuntu-vg/NcVM-snapshot'"
     # Don't exit the script while the snapshot exists to disable any automatic restart during backups
     while does_snapshot_exist "NcVM-snapshot-pending"
--- a/not-supported/backup-viewer.sh
+++ b/not-supported/backup-viewer.sh
@@ -190,15 +190,20 @@ then
     mv /root/.cache/borg/ /root/.cache/borg.bak
 fi
 
-# Mount the repository
+# Mount the drive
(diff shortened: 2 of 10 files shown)


nextcloud_update.sh
```diff
@@ -35,7 +35,9 @@ if does_snapshot_exist "NcVM-snapshot-pending"
 then
     msg_box "Cannot proceed with the update currently because NcVM-snapshot-pending exists.\n
 It is possible that a backup is currently running.\n
-Advice: don't restart your system now if that is the case!"
+Advice: don't restart your system now if that is the case!\n
+If you are sure that no backup is currently running, you can fix this by executing:
+'sudo lvrename /dev/ubuntu-vg/NcVM-snapshot-pending /dev/ubuntu-vg/NcVM-snapshot'"
```
enoch85, line 40: If the updates are run with automation, the end user will never see this message since `msg_box`es gets smashed since no TERM is set.

not-supported/backup-viewer.sh
```diff
@@ -48,10 +48,18 @@ fi
 # MC is needed for this
 if ! is_this_installed mc
 then
-    msg_box "For viewing backups we will need Midnight Commander, which is a command line file explorer.
-Please install it before you can continue with this script by running:
-'sudo bash /var/scripts/menu.sh' choose 'Main Menu => Additional Apps'"
-    exit 1
+    msg_box "For viewing backups we will need Midnight Commander, which is a command line file explorer."
```
enoch85, line 51: Well, you could view them with Webmin as well which is more user friendly IMHO.

not-supported/borgbackup.sh
```diff
@@ -21,6 +21,7 @@ root_check
 
 # Variables
 LVM_MOUNT="/system"
+ZFS_MOUNT="/ncdata"
```
enoch85, line 24: What's this doing?

not-supported/daily-backup-wizard.sh
```diff
@@ -47,16 +55,38 @@ Please rename or delete $BACKUP_SCRIPT_NAME if you want to reconfigure the backu
...
+    then
+        msg_box "The daily backup cannot get configured during the startup script.
+Please try again after it is finished by running: 
+'sudo bash $SCRIPTS/menu.sh' -> 'Server Configuration' -> 'Daily Backup Wizard'."
+        exit
+    fi
+    msg_box "You need to run the update script once before you can continue with creating the backup script."
+    if yesno_box_yes "Do you want to do this now?"
+    then
+        bash "$SCRIPTS"/update.sh minor
```
enoch85, line 71: Better instruct to use the menu script since the update script may change.

not-supported/backup-viewer.sh
szaimen: View them from the command line via webmin?
Actually, I recommend to do it like this:
https://szaimen.github.io/Nextcloud-NAS-Guide/docs/restore-files

not-supported/borgbackup.sh
szaimen: this is the directory where the ZFS snapshot gets mounted

not-supported/daily-backup-wizard.sh
szaimen: hm... but in this case we can simply change the command ourselves.
(I don't recommend to do it like this to the users but instead execute this command directly for the sake of simplicity.)

nextcloud_update.sh
szaimen: Yes I know. I just wanted to add this hint if someone is running the update script manually...

not-supported/backup-viewer.sh
szaimen: Midnight commander is useful to navigate through the folders and see if all files are backed up, so a very simple and fast way to check your backups. Also as long as midnight commander is open, the backup is mounted. Afterwards it gets directly unmounted, etc.

szaimen:
This is ready from my site 👍

not-supported/borgbackup.sh
enoch85: I think the folder name is confusing. Would be better to name it `zfs-backup-snapshots`

szaimen: No, This is the place where the newest snapshot gets mounted from where the data is read that gets backed up (I don't backup from the partition itself but from the snapshot ehich reduces the downtime to a very small timeframe.)

Actually, in my opinion is the name `zfs-backup-snapshots` much more confusing than `ncdata` when you open a backup via the backup viewer because you see the parent directory in the backup itself. 
Please trust me that this directory is the best mountpoint for the snapshot. I've thought about it a lot!

szaimen: Or do you mean the variable name?

not-supported/backup-viewer.sh
enoch85: For a windows user seeing MC, it will be like :exploding_head: :)

What I mean is that you have a file manager in Webmin as well.

not-supported/borgbackup.sh
enoch85: I mean the folder name. Someone not knowing might mix up the `/ncdata` with `/mnt/ncdata` if they explore the file system. It must be very clear that it's not the actual `ncdata` folder IMHO.

szaimen: It is clear because /ncdata is always empty. It will only be filled with data when the snapshot gets mounted there. I mean I could remove it after the backup is done. What do you think about this idea?

not-supported/backup-viewer.sh
szaimen: I see you haven't used it with my script before ;)
first I'll give some advices how to use it:
image
And after you confirm that you've understood and remember those rules, mc gets opened with the correct directory and looks like this:
image

I think this is completely doable and manageable for any user 😉

szaimen: Of course copy and move doesn't work as easily with MC but for that I recommend this https://szaimen.github.io/Nextcloud-NAS-Guide/docs/restore-files

szaimen: The same is of course possible with wrbmin but there is no way we can automate webmin to open the correct directory, etc.

szaimen: It looks like the directory is accessible by opening `https://192.168.178.86:10000/filemin/index.cgi?path=/tmp/borg` in a browser but before you can do this you must be logged in to webmin so it is really hard to automate this... Also it doesn't seem to support any normal mime type like pictures or movies. It only supports viewing text files as far as I can see...

szaimen: So how to proceed from here? I don't want to make webmin a requirement of this script because you can also use Remotedesktop for this which is much better, imo but won't work in all cases, of course...

szaimen: @enoch85 I hope that I fixed this review with 0d0e808

not-supported/borgbackup.sh
szaimen: I hope I've addresses this review with dd32a52

not-supported/backup-viewer.sh
szaimen: I've made it possible now to choose between different options. MC, Webmin and Remotedesktop are now possible. I hope this finally addresses this review perfectly.

not-supported/daily-backup-wizard.sh
enoch85: Seems to me that the minor switch doesn't work for some reason, but that's for another PR.

szaimen: Hm... I think the minor switch worked in my testing but yes, that's definitely something for a different PR.

PR #1915, opened 2021-04-05
smart-monitoring - fix initial disk check

### LINKED ISSUE #1914: S.M.A.R.T check instantly dooms VM disk as failed.
Since the last update the S.M.A.R.T  check thinks that the VM disk is bad on a new computer (my laptop) with a fresh VM.

This is wrong.

I think that the reason is that VM disks aren't supported to be run S.M.A.R.T checks on. In the previous code no real error was made clear, and it silently "failed" (aka didn't produce anything since it's badly supported running in VMs). Maybe that needs to be looked at. Right now everyone running the VM will have S.M.A.R.T  errors automatically.

@szaimen

szaimen:
Fix #1914

### DIFF
--- a/disk/smart-monitoring.sh
+++ b/disk/smart-monitoring.sh
@@ -71,7 +71,7 @@ do
     echo '#########################'
     print_text_in_color "$ICyan" "Testing /dev/$drive"
     OUTPUT=$(smartctl -a "/dev/$drive")
-    if ! echo "$OUTPUT" | grep -q '=== START OF INFORMATION SECTION ===' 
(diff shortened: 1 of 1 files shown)


szaimen:
Now I remember why I've initially tested the smart capability like this. I've tested it successfully on a VM 👍

enoch85:
Maybe it's better to `grep` for this?

image

enoch85:
This works btw, just tested. :+1:

szaimen:
> Maybe it's better to `grep` for this?
> 
> image

No, unfortunately this isn't always shown if a drive lacks smart capability. E.g. I own a drive that not even prints this message out.

Issue #734, opened 2018-12-08
Renew LE SSL DNS with Cloudflare and Dehytrated


enoch85:
https://github.com/sineverba/cfhookbash

- [ ] Create a Clodflare account with bash/python (not possible afaik)
- [ ] Change DNS to Cloudflare with bash/python (not possible afaik)
- [ ] Script the rest of the instructions from the link above.

enoch85:
Maybe this is possible with deSec API?

cc @peterthomassen

gpz1100:
Have you considered acme-dns?  https://github.com/joohoi/acme-dns

This works without using the cloudflare global api key or zone token.  In fact should be compatible with most registrars as no periodic records are actually updated at renewal time, just during initial setup.  Initial set up is more complicated, requiring port 53/udp inbound availability (also needed during updates).  Renewal script can open necessary ufw ports, start service, perform renewal, stop service, close ports.

https://gauravsohoni.wordpress.com/2020/07/21/renewing-letsencrypt-wildcard-ssl-certificate-with-acme-dns/ has a good guide on setting this up.

peterthomassen:
It is possible to use the deSEC API for set up the records for the Let's Encrypt ACME DNS challenge. Our API is also integrated in a bunch of libraries (e.g. acme.sh, lego). For a more detailed list, see https://talk.desec.io/t/tools-implementing-desec/11

> Maybe this is possible with deSec API?

Can you be a bit more specific what you mean by "this"? Is this about automating TLS certificates for users who already have connected their VM to a domain?

enoch85:
What I want to do here is to be able to automatically update the TXT records for LE TLS when renewed. Since the deSEC can do so much, can it also update TXT records grepped from the output of the LE TLS (DNS) challenge?

In that case we could still use certbot.

peterthomassen:
Yes, we have a hook script for that: https://github.com/desec-io/desec-certbot-hook It can be enabled using `--manual-auth-hook` / `--manual-cleanup-hook`. For details, see the README in the repo.

There's also a certbot plug-in here: https://github.com/akararsse/certbot-dns-desec However, we have not reviewed it yet, and I have not tried to use it. It's possible that it simplifies things, but it's also possible it's broken -- so in case you want to use it, I'd recommend to do some quality assurance first.

enoch85:
Thanks @peterthomassen!

Once we implemented the deSEC script here, we might take a look at this issue as well.

[...]

PR #1920, opened 2021-04-11
Add DNS updates with deSEC

enoch85:
@szaimen @peterthomassen 

This PR aims to fix DNS renewals, a long standing issue in this repo.

Please comment.

This will go into 21.0.01

Fix https://github.com/nextcloud/vm/issues/734

### DIFF
--- a/addons/desec.sh
+++ b/addons/desec.sh
@@ -106,7 +106,7 @@ then
     # Add DNS renewals with the deSEC hoock script
     print_text_in_color "$ICyan" "Preparing for DNS-renewals..."
     mkdir -p "$SCRIPTS"/deSEC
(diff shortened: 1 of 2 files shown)


peterthomassen (commented):
Overall, this looks nice!

You may want to consider copying the logic from our hook script into your own codebase, so that you don't depend on our script's "API" (and also reduce the number of dependencies you have).

enoch85:
> You may want to consider copying the logic from our hook script into your own codebase,

I actually prefer if someone else maintains that. ;) I get your point though. :+1:

enoch85:
This is pure magic! Tested on a local machine without any ports open at all, everything was setup like a charm. Wow!

PR #1921, opened 2021-04-16
catch error on regular run

enoch85:


enoch85:
@szaimen sorry, now I remember why we made it like that - Shellcheck!

lets-encrypt/activate-tls.sh
```diff
@@ -220,44 +220,34 @@ if [ -n "$DEDYNDOMAIN" ]
...
-    certonly
-else
-    generate_cert "$TLSDOMAIN"
-fi
-
-# Generate DHparams
-if [ -d "$CERTFILES" ]
-then
-    if [ ! -f "$DHPARAMS_TLS" ]
+    if certbot certonly --manual --text --rsa-key-size 4096 --renew-by-default --server https://acme-v02.api.letsencrypt.org/directory no-eff-email --agree-tos --preferred-challenges dns --manual-auth-hook "$SCRIPTS"/deSEC/hook.sh --manual-cleanup-hook "$SCRIPTS"/deSEC/hook.sh -d "$DEDYNDOMAIN"
```
szaimen, line 223: Why did you put all of this into one line? It is much harder to read like this, imho...

enoch85: to do the `if` statement.

szaimen: But this is also possible with using many lines. See for example: https://github.com/nextcloud/vm/blob/277908e21a2b48891c5e4a5f7c357c85e32b914c/not-supported/borgbackup.sh#L369-L370

szaimen: ping @enoch85

enoch85: Feel free to create a PR where that's sorted.

PR #1922, opened 2021-04-16
reintroduce line breaks to specific command in activate-tls

szaimen:
Follow-up to #1921


enoch85:
Need to test this, but are on vacation.

PR #1923, opened 2021-04-17
Add pdfannotate

szaimen:
Successor to #1870

---

needs:
- [ ] published to the appstore

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -513,6 +513,20 @@ then
     # Check for upgrades
     print_text_in_color "$ICyan" "Trying to automatically update all Nextcloud apps..."
     UPDATED_APPS="$(nextcloud_occ_no_check app:update --all)"
+    # Update pdfannotate
+    if [ -d "$NC_APPS_PATH/pdfannotate" ]
+    then
+        INFO_XML="$(curl -s https://gitlab.com/nextcloud-other/nextcloud-annotate/-/raw/master/appinfo/info.xml)"
+        if [ "$(echo "$INFO_XML" | grep -oP 'min-version="[0-9]+"' | grep -oP '[0-9]+')" -le "${CURRENTVERSION%%.*}" ] \
+&& [ "$(echo "$INFO_XML" | grep -oP 'max-version="[0-9]+"' | grep -oP '[0-9]+')" -ge "${CURRENTVERSION%%.*}" ]
+        then
+            print_text_in_color "$ICyan" "Updating the pdfannotate app..."
+            cd "$NC_APPS_PATH/pdfannotate"
+            git pull
+            chown -R www-data:www-data ./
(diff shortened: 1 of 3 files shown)


szaimen:
@ASerbinski This was already tested successfully. What do you think, how long will it take until you can get rid of the dependencies? Does it make sense to wait for it or just merge this PR as it is?

szaimen:
@ASerbinski Also I'd like to know how you would recommend to to do app updates in the future if the app isn't published to the app store?

lbdroid:
git pull in the application's directory would do it.
I can't give you any estimates, but probably won't be able to do to much with it until May.

ASerbinski:
git pull would certainly update it.
I can't commit to any timelines, and won't be able to get too much into it until after May 1.

szaimen:
> I can't commit to any timelines, and won't be able to get too much into it until after May 1.

Okay, thanks!

> git pull would certainly update it.

Hm... I have to discuss with @enoch85 if that is feasible.

enoch85:
> @enoch85 What do you think about this?

If the app is not in the app store, then I don't think this is a good idea.

szaimen:
> If the app is not in the app store, then I don't think this is a good idea.

I've edited the description so that we don't forget this going forward 👍

nextcloud_update.sh
```diff
@@ -513,6 +513,20 @@ then
...
+        if [ "$(echo "$INFO_XML" | grep -oP 'min-version="[0-9]+"' | grep -oP '[0-9]+')" -le "${CURRENTVERSION%%.*}" ] \
+&& [ "$(echo "$INFO_XML" | grep -oP 'max-version="[0-9]+"' | grep -oP '[0-9]+')" -ge "${CURRENTVERSION%%.*}" ]
+        then
+            print_text_in_color "$ICyan" "Updating the pdfannotate app..."
+            cd "$NC_APPS_PATH/pdfannotate"
+            git pull
+            chown -R www-data:www-data ./
+            chmod -R 770 ./
+        fi
+    fi
```
szaimen, line 529: This is how a manual update would look like.

ASerbinski: You could also do a ```sudo -u www-data git pull```
I'm also inclined to add tags for significant milestones, probably starting when the dependencies are removed.

szaimen: > You could also do a sudo -u www-data git pull

Yes, I know 👍

> I'm also inclined to add tags for significant milestones, probably starting when the dependencies are removed.

I am not sure if milestones help to figure out if the app is compatible with the current NC version...

enoch85:
Hey @ASerbinski, now when the dust has settled, any chance that you are willing to make this more maintainable - e.g. make it an Nextcloud app store app? :)

PR #1924, opened 2021-04-18
Add automatic snapshot unlocking upon reboot

szaimen:

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -34,10 +34,9 @@ is_process_running dpkg
 if does_snapshot_exist "NcVM-snapshot-pending"
 then
     msg_box "Cannot proceed with the update currently because NcVM-snapshot-pending exists.\n
-It is possible that a backup is currently running.\n
+It is possible that a backup is currently running or an update wasn't successful.\n
 Advice: don't restart your system now if that is the case!\n
-If you are sure that no backup is currently running, you can fix this by executing:
-'sudo lvrename /dev/ubuntu-vg/NcVM-snapshot-pending /dev/ubuntu-vg/NcVM-snapshot'"
+If you are sure that no update or backup is currently running, you can fix this by rebooting your server."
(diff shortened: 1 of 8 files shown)


nextcloud_update.sh
```diff
@@ -998,5 +1001,23 @@ Maintenance mode is kept on."
...
+            mapfile -t PROCESS_IDS_NEW <<< "$PROCESS_IDS_NEW"
+            for process in "${PROCESS_IDS_NEW[@]}"
+            do
+                print_text_in_color "$ICyan" "Killing the process with PID $process to prevent a potential automatic restart..."
+                if ! kill "$process"
+                then
+                    print_text_in_color "$IRed" "Couldn't kill the process with PID $process..."
+                fi
+            done
+        fi
```
szaimen, line 1020: I've copied this from here: https://github.com/nextcloud/vm/blob/caa3b163b8ccadf782d7b055193094b31fb5d2ae/nextcloud_update.sh#L41-L55

szaimen:
This is ready 👍

enoch85:
It would be nice if you would explain exactly what it does. Sure I can read code, but it's easier if you explain. :)

szaimen:
As you might know is the snapshot with the name NcVM-snapshot-pending currently preventing updates and backups & co to get executed if one of those already runs. It is kind of a blocker that prevents bad things from happening. If you shut down the VM while an update is running or while a backup is running, currently the next backup or update will be blocked from executing because the snapshot-pending still exists. To prevent those issues and make it easier to unlock the backup and update, I've added a cronjob that automatically unlocks backups and updates upon reboot.

enoch85:
@szaimen Urgent: https://github.com/nextcloud/vm/runs/2403856054

szaimen:
@enoch85 where does this error suddenly come from? It was all green when I've vreated the PR...

PR #1927, opened 2021-04-21
fix SC in master

szaimen:


szaimen:
This should fix SC in master but I wonder why this issue just popped up and SC didn't find it before.... 🤔

PR #1930, opened 2021-04-22
Revert "reintroduce line breaks to specific command in activate-tls"

enoch85:
Reverts nextcloud/vm#1922

Tested live and it doesn't work


enoch85:
```
usage: 
  certbot [SUBCOMMAND] [options] [-d DOMAIN] [-d DOMAIN] ...

Certbot can obtain and install HTTPS/TLS/SSL certificates.  By default,
it will attempt to use a webserver both for obtaining and installing the
certificate. 
certbot: error: unrecognized arguments: no-eff-email
```

Issue #1931, opened 2021-04-24
Apache log in root directory / with extreme sizes


markkco:
Hello!
I'm using Nextcloud VM installed with scripts (currently updated to 20.0.9). I stumbled on /access.log being few hundred MB's.

I checked why does Apache log out to /access.log. Could it be that /etc/apache2/sites-enabled/domain.com.conf
contain:
```
# Logs
LogLevel warn
CustomLog /access.log combined
ErrorLog /error.log
```
Should log files be located in /var/log/apache2/* ?

enoch85:
@markkco Hey, great find!

Which script is this related to? Collabora?

szaimen:
After a quick search I found out that probably all scripts with apache configs are effected since ${APACHE_LOG_DIR} isn't escaped anywhere and because it also doesn't exist in the lib.sh file, it just gets written as empty to the file...
So either we add the variable to the lib.sh file or we escape ${APACHE_LOG_DIR} in order to solve this...

szaimen:
Also we should probably delete the file in the root folder and/or rotate it regularly?

markkco:
Another strange thing connected to the issue is that /etc/apache2/envvars do not get loaded at startup. I'm linux newb so maybe this has to be this way. Apache should write logs to $APACHE_LOGDIR as defined in envvars.
Script which generates /etc/apache2/sites-available/domain.com.conf is [https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L682](url)
based on template default-ssl.conf (?):
```
ErrorLog ${APACHE_LOG_DIR}/error.log
CustomLog ${APACHE_LOG_DIR}/access.log combined
```
And, Daniel, thank You for creating NC VM scripts! Also to everyone working on them!

szaimen:
See #1932

enoch85:
The PR is now merged. 

I think maybe we should delete the old logs as @szaimen suggested. But at the same time, that means fixing all old installations as well, so naah, let's leave it.

enoch85:
Close?

szaimen:
Question is if there should be something like a logrotate in place?

enoch85:
>Question is if there should be something like a logrotate in place?

I think that's handled automatically since we use the default Apache log path.

szaimen:
> I think that's handled automatically since we use the default Apache log path.

If thats actually the case then great 👍

PR #1932, opened 2021-04-24
fix APACHE_LOG_DIR

szaimen:

### DIFF
--- a/apps/adminer.sh
+++ b/apps/adminer.sh
@@ -83,8 +83,8 @@ Listen 9443
 
     # Logs
     LogLevel warn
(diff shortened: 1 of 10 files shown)


szaimen:
This questions remains before merging (see https://github.com/nextcloud/vm/issues/1931#issuecomment-826157924):
> Also we should probably delete the file in the root folder and/or rotate it regularly?

enoch85:
Hmm, I'd actually set it in lib.sh, but I guess this is more "Linux like"-standard since the path is controlled in it's regular way.

Thanks @szaimen! 👍🏻

PR #1933, opened 2021-04-25
allow btrfs backup drives

szaimen:
This is the first of two PRs that allow BTRFS drives to be added into the backup scripts including scrubbing and snapshots.
I've divided them into two because this one needs to be reviewed because we decided to make the backup script official.


szaimen:
This was already successfully tested the last weeks hence marking as ready for review.

PR #1934, opened 2021-04-25
add btrfs options to the not-supported folder

szaimen:
This is the second of two PRs that allow BTRFS drives to be added into the backup scripts including scrubbing and snapshots as well as being the veracrypt encrypted.
This PR doesn't need as much review because it doesn't effect any official script. It may look like much but in I've basically copied from my other existing scripts.


szaimen:
This was already successfully tested the last weeks hence marking as ready for review.

enoch85:
Great job, but we don't use BTRFS, we use ZFS, so why add such an option at all?

szaimen:
> Great job

Thanks!

> but we don't use BTRFS, we use ZFS, so why add such an option at all?

I think I don't understand the question completely. The simple answer to this question would be because I think it makes sense.
Also, I thought that since this is the not-supported folder, you wouldn't have anything against it?

The longer answer is:
- because it is officially supported and included into the linux kernel and works like any other FS on Linux (without the IMO strange zpool create, export, import, etc.) (and thus can be added to the /etc/fstab file, etc.)
- because It has comparable features like ZFS and runs stable, as far as I can tell (the NextcloudPi project migrated to BTRFS around 4 years ago and there are no issues concerning this decision at all) (also it runs stable in my testing) (I don't think that ZFS is unstable but BTRFS is stable, too)
- It is supported by Veracrypt (which ZFS is not)
- It has a good driver for Windows that works after installing it out of the box by just connecting the drive without any needed command line magic whatsoever (see https://github.com/maharmstone/btrfs) (also with veracrypt encrypted BTRFS drives)

szaimen:
@enoch85 this is not meant to be a replacement for ZFS but for the NTFS scripts that I wrote.

enoch85:
I understand your arguments, but we chose between BTRFS and ZFS a long time ago. Sure we could have chosen BTRS back then but were told *not* to due to several reasons I can't remember now.

Putting it in the not-supported folder or not, this still means that we would open up for BTRFS, which is not the case. I'm totally against it and sees BTRFS as the "hobby version" of ZFS - stable or not.

szaimen:
Thanks for the clarification!

> I understand your arguments, but we chose between BTRFS and ZFS a long time ago. Sure we could have chosen BTRS back then but were told _not_ to due to several reasons I can't remember now.
> Putting it in the not-supported folder or not, this still means that we would open up for BTRFS, which is not the case. I'm totally against it and sees BTRFS as the "hobby version" of ZFS - stable or not.

Even if BTRFS could actually be considered as "hobby version" of ZFS, I would be very thankful if you would accept those scripts because BTRFS is the only option that is viable for myself (especially because of the veracrypt and Windows support). 
I understand that you don't like BTRFS but allowing BTRFS doesn't mean that we block ZFS, does it?

enoch85:
> for myself

Let's see what people think: https://twitter.com/tmhanssonit/status/1387147850497372167 :robot:

szaimen:
> Let's see what people think: https://twitter.com/tmhanssonit/status/1387147850497372167 🤖

I am not sure what you are trying to do here? Even if ZFS should be elected, it won't change this PR or my decision for BTRFS. 
Or are you trying to convince me to write scripts that use ZFS, instead?

szaimen:
What I still don't get is why you didn't have anything against NTFS which I've added a few months ago to this folder but now that I try to introduce a next gen FS, it isn't allowed?

enoch85:
> Or are you trying to convince me to write scripts that use ZFS, instead?

Actually, yes, that would be super nice! Then I'd really think about putting it in the production and released version. :+1:

enoch85:
> What I still don't get is why you didn't have anything against NTFS which I've added a few months ago to this folder but now that I try to introduce a next gen FS, it isn't allowed?

It's all about BTRFS being super unstable and unpredictable. Maybe my studies are from way back, but all the information I gathered pre my decision to use ZFS pointed to that. I've also lost large data drives when using BTRFS myself - because I actually tested it before I opted for ZFS - which I never had any issues with.

Regarding NTFS, it is what it is. It's Windows, and we want a broader use.

szaimen:
Thanks for discussing this with me! :)

> It's all about BTRFS being super unstable and unpredictable. Maybe my studies are from way back, but all the information I gathered pre my decision to use ZFS pointed to that. I've also lost large data drives when using BTRFS myself - because I actually tested it before I opted for ZFS - which I never had any issues with.

Now it makes sense why you don't like BTRFS. I already expected something like this but wasn't sure.
I am sure that it had it's issues when it was new but it runs pretty stable with any recent kernel, afaik. As I said, if this wouldn't be the case, the NextcloudPi project and other projects would have dropped it long time ago. (Red Hat Enterprise Linux is maybe an exception but as I found out, the FS didn't fit into their business plan and they had used a pretty old kernel which seemingly was too old for stable BTRFS support.)

> Regarding NTFS, it is what it is. It's Windows, and we want a broader use.

Since a good BTRFS driver for Windows exists, I think that we actually keep this usecase also with BTRFS. Hence, I originally planned to drop the NTFS scripts in the long term.

---

>> Or are you trying to convince me to write scripts that use ZFS, instead?
>
> Actually, yes, that would be super nice! Then I'd really think about putting it in the production and released version. 👍

Let's do it like this:
If you allow those BTRFS scripts (including the needed changes in #1933) in the not-supported folder, I might write the scripts for ZFS, too. If you decide to not allow them, I will not be the one that writes the ZFS scripts.

enoch85:
image

PR #1935, opened 2021-04-27
clarification

jospoortvliet:
We don't want customers to think we can support the VM as we don't.


jospoortvliet:
good to merge from my side.

Issue #1942, opened 2021-05-05
NCBACKUP-OLD takes a huge amount of space


ishmumx:
First of all I would like to extend many thanks to the creators of NCVM. What a phenomenal piece of software that I have been using for years! I have a little bug for you: When you mount an SMB share, it tends to download the SMB contents. Weird...or not? You tell me. Thank you!
-->
## Steps to reproduce
1. Mount an SMB share with 100GB of data.
2. Go to disk usage
3. /mnt/smbshares has the SMB mounts all downloaded, resulting in 100GB of data in the VM too.

## Expected behavior
It should just browse the SMB share, not download the entirety of the contents.

## Actual behavior
It downloads the content, resulting in a huge uptick in disk usage. My VM actually failed because the VM is inside a 128GB SSD, and the SSD filled up from the SMB shares.

#### Log file (/var/log/nextcloud/nextcloud.log)
```
https://0bin.net/paste/RNUD6XuL#SpbmFWbJVx35H4Ijln2MiXoEpodAEdbQZmuSLLoqXMb
```

szaimen:
Hi, thanks for your feedback and report! Could you please post the output from `sudo lsblk` and `sudo df -h` here? Thanks!

UNlDAN:
sudo lsblk:

![image](https://user-images.githubusercontent.com/37858032/117185089-4fc1e300-ad8e-11eb-8f7f-8f2af01d6eb6.png)

sudo df -h

![image](https://user-images.githubusercontent.com/37858032/117185178-67996700-ad8e-11eb-9da7-7c08f5ac0ac1.png)


Thank you!

szaimen:
Hm... the output looks fine. What does `du -sh /mnt/ncdata/*` return?

---

> When you mount an SMB share, it tends to download the SMB contents.

I wonder how did you come to this conclusion?

> Go to disk usage
> /mnt/smbshares has the SMB mounts all downloaded, resulting in 100GB of data in the VM too.

Could you post a screenshot from this disk usage overview?

UNlDAN:
> Hm... the output looks fine. What does `sudo du -sh /mnt/ncdata/*` return?
> 
![image](https://user-images.githubusercontent.com/37858032/117200431-58bbb000-ada0-11eb-925b-32e52a6a55a1.png)

> > When you mount an SMB share, it tends to download the SMB contents.
> 
> I wonder how did you come to this conclusion?

[...]

PR #1946, opened 2021-05-07
only store one old backup and one current backup during the update script to limit the used storage

szaimen:
Fix https://github.com/nextcloud/vm/issues/1942
See my comment in https://github.com/nextcloud/vm/issues/1942#issuecomment-834812405

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -735,7 +735,6 @@ print_text_in_color "$ICyan" "Backing up data..."
 if [ -d "$BACKUP" ]
 then
     rm -rf "$BACKUP"-OLD
-    mkdir -p "$BACKUP"-OLD
     mv "$BACKUP" "$BACKUP"-OLD
     rm -rf "$BACKUP"
     mkdir -p "$BACKUP"
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -736,8 +736,7 @@ if [ -d "$BACKUP" ]
 then
     rm -rf "$BACKUP"-OLD
     mkdir -p "$BACKUP"-OLD
-    install_if_not rsync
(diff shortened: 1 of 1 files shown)


nextcloud_update.sh
```diff
@@ -732,12 +732,12 @@ fi
...
 print_text_in_color "$ICyan" "Backing up data..."
-DATE=$(date +%Y-%m-%d-%H%M%S)
 if [ -d "$BACKUP" ]
 then
-    mkdir -p "$BACKUP"-OLD/"$DATE"
+    rm -rf "$BACKUP"-OLD
+    mkdir -p "$BACKUP"-OLD
     install_if_not rsync
-    rsync -Aaxz "$BACKUP"/ "$BACKUP"-OLD/"$DATE"
+    rsync -Aaxz "$BACKUP"/ "$BACKUP"-OLD
```
szaimen, line 740: What about renaming the backup folder here instead of rsyncing the content?

enoch85:
Thanks!

I think a better implementation would be to delete `older than X days`.

This  specific commit would also delete all old backups for current users, I'm not sure we want that.

enoch85:
Maybe something like: `find "$BACKUP"-OLD/*" -mtime +90 -exec rm {} \;`

Another option would be to move all the data to the /mnt/ncdata instead... That's the original location but is changed during the script for some reason I can't remember: https://github.com/nextcloud/vm/blob/b6edbac199cd2a3c5ed640b460459b98dfda9fe1/nextcloud_update.sh#L703

szaimen:
@enoch85 as you wanna do it pretty differently than my proposal, I suggest that you open a new PR and implement there how you would do it and then I can have a look if I am okay with this. What do you think?

PR #1950, opened 2021-05-11
set retention_obligation to default

szaimen:
I think 30 days makes more sense because it is the default in many applications like other clouds or mail-providers and also Nextcloud itself. 
What do you think about this?


szaimen:
Hm... Maybe instead of changing it here, should be configurable by the admin, instead? So is this something for the nextcloud_configuration menu?

PR #1951, opened 2021-05-11
move backups location and delete old backups

enoch85:
@szaimen let's continue here


enoch85:
Tested and works, let's do this!

PR #1952, opened 2021-05-12
allow to install both - ONLYOFFICE and Collabora

### LINKED ISSUE #1163: Install of both ONLYOFFICE and Collabora?
nextcloud_install_production.sh uninstalls ONLYOFFICE when installing Collabora, I wonder if it would be possible to have both installed by the script?

If no what are the issues preventing a simultaneous install?

szaimen:
fix #1163
I also moved the integrated documentservers to the old folder since after thinking about it I think it is easy enough to install them through the webpanel if needed, the docker documentservers are better in terms of stability and performance and we can get rid of two scripts 🚀


enoch85:
> I also moved the integrated documentservers to the old folder since after thinking about it I think it is easy enough to install them through the webpanel if needed,

There's a reason for the current logic. 

I don't know if you remember but the document server apps can't be installed if it's taking too long from the app store, since there are a limit on how long an app can be downloaded. There's no limit with the `occ` command. That's why we put them here in the first place. :)

This also seems to remove all the logic for checks if "the other" is already installed.

Sorry, but I don't like this.

szaimen:
Next try in #1953

PR #1953, opened 2021-05-13
allow to install both - ONLYOFFICE and Collabora - 2

### LINKED ISSUE #1163: Install of both ONLYOFFICE and Collabora?

szaimen:
Supersedes #1952
Fix #1163

### DIFF
--- a/apps/collabora_docker.sh
+++ b/apps/collabora_docker.sh
@@ -68,55 +68,6 @@ else
     removal_popup "$SCRIPT_NAME"
 fi
 
-# Check if OnlyOffice is previously installed
-# If yes, then stop and prune the docker container
-if does_this_docker_exist 'onlyoffice/documentserver'
-then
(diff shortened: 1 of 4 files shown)


szaimen:
@enoch85 do you have some thoughts on this?

szaimen:
@enoch85 whats the problem here?

enoch85:
> @enoch85 whats the problem here?

I don't like it. 

It's no playground for testing them out side by side. Either you choose Collabora or OnlyOffice.

I'm also not sure about the side effects. Just because it's possible it doesn't mean it's best practice.

szaimen:
So then please go ahead and close the Issue and this PR since you don't want to let it happen!

PR #1960, opened 2021-05-17
choose port with deSEC

enoch85:
TODO: 

~~- Check for NONO ports~~

### DIFF
--- a/apps/talk.sh
+++ b/apps/talk.sh
@@ -83,21 +83,27 @@ lowest_compatible_nc 19
 # Check if Nextcloud is installed with TLS
 check_nextcloud_https "Nextclod Talk"
 
-# Let the user choose port. TURN_PORT in msg_box is taken from lib.sh and later changed if user decides to.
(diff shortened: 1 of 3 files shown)


enoch85:
FYI @peterthomassen @szaimen

peterthomassen:
I like the idea of choosing a port for Talk -- but I'm not sure: what does it have to do with deSEC or dedyn.io?

enoch85:
> I like the idea of choosing a port for Talk -- but I'm not sure: what does it have to do with deSEC or dedyn.io?

I needed to fix the Talk script as well to be able to make the check a function.

We will now offer users to choose port directly in the setup so that they don't have to manually change the listening port in Apache - which defaults to 433.

enoch85:
@peterthomassen I guess this is the what you're looking for: https://github.com/nextcloud/vm/pull/1960/files#diff-30b7465f5e2121a291f1b4ddb20cd4af5d64189ec2001a9ad0a5ed4e0c69d568R230

Issue #1680, opened 2020-11-14
Make every update log specific with date


enoch85:
Instead of just adding stuff to the update.log we should append `$date` to the filename so that we don't grow a single file.

It's better because:
1. The latest stuff is appended to the end of the log, and after a while it becomes several thousands of lines. This fix would only print a single update run
2. We can delete old logs, let's say 30 logs back since they are probably not needed

I also think that we should put all update logs in `$VMLOGS/update/update-$(date).log`.

szaimen:
@enoch85 I would propose creating a new file during the update and merging it into one in the end.
Also we should get rid of spinner_loading in the update script, because it really spams any logfioe.

enoch85:
> I would propose creating a new file during the update and merging it into one in the end. 

Isn't that the same as now, once single big file. That's what we don't want. :)

> Also we should get rid of spinner_loading

As a Linux novice you'd then think that nothing is happening during update and abort the script. Not a good idea. Been there, done that. :/

PR #1974, opened 2021-05-20
add dates to automatic updates log

enoch85:
Fix https://github.com/nextcloud/vm/issues/1680

### DIFF
--- a/addons/automatic_updates.sh
+++ b/addons/automatic_updates.sh
@@ -39,8 +39,8 @@ fi
 
 # Install automatic updates
 DATE='`date +%Y-%m-%d_%H:%M`'
-mkdir -p $VMLOGS/updates
-crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/updates/update-$DATE.log 2>&1"; } | crontab -u root -
+mkdir -p "$VMLOGS"/updates
+crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/updates/update-${DATE}.log 2>&1"; } | crontab -u root -
 if yesno_box_yes "Do you want to reboot your server after every update? *recommended*"
 then
     sed -i "s|exit|/sbin/shutdown -r +1|g" "$SCRIPTS"/update.sh
--- a/addons/automatic_updates.sh
+++ b/addons/automatic_updates.sh
@@ -4,6 +4,7 @@
 
 true
 SCRIPT_NAME="Automatic Updates"
+# Disable https://github.com/koalaman/shellcheck/wiki/SC2016 since we don't want to expand DATE for the crontab
 # shellcheck source=lib.sh disable=2016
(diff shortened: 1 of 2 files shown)


addons/automatic_updates.sh
```diff
@@ -38,8 +40,9 @@ else
 fi
 
 # Install automatic updates
-touch $VMLOGS/update.log
-crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/update.log 2>&1"; } | crontab -u root -
+DATE='`date +%Y-%m-%d_%H:%M`'
+mkdir -p "$VMLOGS"/updates
+crontab -u root -l | { cat; echo "0 $AUT_UPDATES_TIME * * 6 $SCRIPTS/update.sh minor >> $VMLOGS/updates/update-${DATE}.log 2>&1"; } | crontab -u root -
```
szaimen, line 45: @enoch85 I didn't test this but it looks like all updates just get written to an update log file that is called e.g.
/var/log/nextcloud/updates/update-2021-05-20_22:16.log instead of getting written to /var/log/nextcloud/update.log.

enoch85: That will expand with the current date in the file name, not the date that it's being run.

This PR is 100% intentional. I will delete old logs in another PR as well.

Thanks for the pointers! 👍🏻

szaimen: yes, but your PR doesn't work as intended, as far as I can see.
Please see my suggestion.

szaimen: After thinking about it it might work excactly like my proposal.

enoch85: Have you tested it?

szaimen: No but I am able to read the code ;)

enoch85: OK, everything works as intended anyway. 

I had missed to espace the date variable correctly (did it in my testing but not in the final PR) but that's fixed now.

The intention here is to get a new file for each run which is how it works now.

By doing as you suggest the date variable would be expanded in the crontab filename directly, and always get the same date, to the same file - which is wrong.

szaimen: > The intention here is to get a new file for each run which is how it works now.

Yes, that is also what my proposal is doing

> By doing as you suggest the date variable would be expanded in the crontab filename directly, and always get the same date, to the same file - which is wrong.

I though this would be the mistake that you've done in your PR but obviously we've found two solutions to the same idea ;)
Concerning my proposal: By escaping the date command, every log gets its own log file.

enoch85: Tried that already, didn't work. Hence the ' ' in the date variable. 

Your output: 
/var/log/nextcloud/updates/update-$(2021-05-21_08:27).log 2>&1

What we want (current solution):
/var/log/nextcloud/updates/update-`date +\%Y-\%m-\%d_\%H:\%M`.log 2>&1

enoch85: OK, rest my case. Works without the back ticks.

szaimen: hm... the escaping should technically do it... strange

PR #1979, opened 2021-05-21
POC for regular zfs snapshot prune

### LINKED ISSUE #1971: Remove zfs-auto-snapshots?
Snapshots are pretty space consuming, especially on smaller disks. I also don't think that users actually even know how to restore from a data snapshot. If they would, then the database would have to be restored as well due to file cache and what not.

So basically, we provide a function that I don't think many users actually use. Maybe it would be better to:

- Make it an option instead?
- Remove it completely?
- Release larger disks as default?
- Remove it for now, and reintroduce it when you can put ZFS on the main disk (OS) as well? Not until then we could actually provide a clean way to restore directly.

As of today you can find it in these places:

```
nextcloud_update.sh:if is_this_installed zfs-auto-snapshot
nextcloud_update.sh:        sed -i "s|date --utc|date|g" /usr/sbin/zfs-auto-snapshot
nextcloud_update.sh:        check_command zfs-auto-snapshot -r ncdata
disk/prune_zfs_snaphots.sh:    if is_this_installed zfs-auto-snapshot
disk/format-chosen.sh:        install_if_not zfs-auto-snapshot
disk/format-chosen.sh:        sed -i "s|date --utc|date|g" /usr/sbin/zfs-auto-snapshot
disk/format-chosen.sh:Automatic snapshots are taken with 'zfs-auto-snapshot'. You can list current snapshots with:
disk/format-chosen.sh:http://manpages.ubuntu.com/manpages/focal/man8/zfs-auto-snapshot.8.html
disk/format-sdb.sh:        install_if_not zfs-auto-snapshot
disk/format-sdb.sh:        sed -i "s|date --utc|date|g" /usr/sbin/zfs-auto-snapshot
disk/format-sdb.sh:Automatic snapshots are taken with 'zfs-auto-snapshot'. You can list current snapshots with:
disk/format-sdb.sh:http://manpages.ubuntu.com/manpages/focal/man8/zfs-auto-snapshot.8.html
not-supported/borgbackup.sh:    sed -i "s|date --utc|date|g" /usr/sbin/zfs-auto-snapshot
not-supported/borgbackup.sh:    if ! zfs-auto-snapshot -r ncdata
```

What do you think?

szaimen:
Fix #1971
Disclaimer: I didn't test this but it should theoretically work.

### DIFF
--- a/disk/change-to-zfs-mount-generator.sh
+++ b/disk/change-to-zfs-mount-generator.sh
@@ -65,3 +65,22 @@ then
     print_text_in_color "$ICyan" "/etc/zfs/zfs-list.cache/$POOLNAME is empty, setting values manually instead."
     zfs list -H -o name,mountpoint,canmount,atime,relatime,devices,exec,readonly,setuid,nbmand,encroot,keylocation > /etc/zfs/zfs-list.cache/"$POOLNAME"
 fi
+
+# Add daily snapshot prune
+cat << PRUNZE_ZFS > "$SCRIPTS/daily-zfs-prune.sh"
+#!/bin/bash
+
+# Source the library
+source /var/scripts/fetch_lib.sh
+
+# Check if root
+root_check
+
+# Run the script
+run_script DISK prune_zfs_snaphots
+PRUNZE_ZFS
+
+# Add crontab
+chmod +x "$SCRIPTS/daily-zfs-prune.sh"
+crontab -u root -l | grep -v "$SCRIPTS/daily-zfs-prune.sh"  | crontab -u root -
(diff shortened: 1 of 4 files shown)


enoch85:
I get your idea, nice. But this will also clear trashbin since it's running the cleanup script.

But hmm, only since it's 70%... Yeah, this might be OK.

I do think that the script is created in the wrong place though. Make it a static script in the repo instead.

szaimen:
Do you mean like this? 0155949

enoch85:
Yes exactly, but put the creation in the installation script instead. :+1:

szaimen:
done 👍

PR #1982, opened 2021-05-22
Adjust umask for new files

szaimen:
Adjust umask for new files (defaults to 0133 which gives everyone read access). Now only www-data members and user will be able to read and change new files.

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -666,6 +666,22 @@ fi
 restart_webserver
 fi
 
+# Adjust umask for new files (defaults to 0133 which gives everyone read access)
+# https://stackoverflow.com/questions/39290444/how-to-set-umask-for-www-data-user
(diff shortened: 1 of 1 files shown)


enoch85:
My question is, if the UMASK is set to a default, it's for a reason. So what's the reason behind changing it?

Security sure, but are we 100% sure about the side effects?

szaimen:
I must admit that I cannot say why it is set to 0133 but it is definitely insecure.
BTW: file creation, deletion, etc. works still fine with my PR which we should care about, no?

enoch85:
What's the background for you adding this here then? 

Have you tested a full installation with this, all the apps and all the other things? I mean PHP-FPM is the core in the VM - if it breaks it's disaster.

szaimen:
The idea was that all new instances profit from this security benefit. I for myself edited it in the correct file but I currently don't have the time to test sufficiently so I am going to close this for now.

PR #1987, opened 2021-05-22
update mimetype list

szaimen:
Sometimes the mimetype list gets updated and as far as I know doesn't Nextcloud get those automatically so they need to be refreshed manually.

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -819,6 +819,8 @@ then
     nextcloud_occ upgrade
     # Optimize
     print_text_in_color "$ICyan" "Optimizing Nextcloud..."
+    nextcloud_occ maintenance:mimetype:update-js
+    nextcloud_occ maintenance:mimetype:update-db
     yes | nextcloud_occ db:convert-filecache-bigint
     nextcloud_occ db:add-missing-indices
     CURRENTVERSION=$(sudo -u www-data php $NCPATH/occ status | grep "versionstring" | awk '{print $3}')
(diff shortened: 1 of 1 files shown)


enoch85:
OK, is this version specific? When was this introduced in Nextcloud?

szaimen:
It exsits since forever. This is the oldest commit that references the command    that I could find:
image
https://github.com/nextcloud/server/commit/a4b19a5b1e4079752e33d6eb75c72a47ce048bde#diff-27115c2bceab8180cc3f88647640c9e690d125591b7fb70500d9f9d0c011293e

enoch85:
I changed the location to the "optimize" thing.

szaimen:
That's good, problem is I don't know if it works while maintenance mode is on... I guess we should test this... Let me do that now...

szaimen:
I've tested it and it seems like both commands works while maintenance mode is on 👍

enoch85:
What was the specific reason for adding this? I just had a customer which the whole files app broke, and he couldn't see his files at all.

szaimen:
> Sometimes the mimetype list gets updated and as far as I know doesn't Nextcloud get those automatically so they need to be refreshed manually.

For example: the pdf icon that was updated in one year ago by the server team didn't show up in the files app. 
After running those commands the new icon shows up.

> I just had a customer which the whole files app broke, and he couldn't see his files at all.

I am interested: how did you fix the problem on his instance? I am pretty sure that it is not related to the command itself but due to a misconfiguration of his instance.

szaimen:
> I just had a customer which the whole files app broke, and he couldn't see his files at all.

Can I see the update log of the update that broke this for him/her?

enoch85:
Sorry, no log since it wasn't automatically done.

szaimen:
hm... based on the poor documentation that I could find about those commands, maybe we should consider moving them back to the place where I originally planned them to be? It seems like they need database access and the database isn't available in the maintenance mode, afaik. This could potentially fix such errors in the future...

enoch85:
It's already off: https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L829

szaimen:
> It's already off: https://github.com/nextcloud/vm/blob/master/nextcloud_update.sh#L829

Ah you are right. So this wasn't the problem in his/her case then.

enoch85:
We could leave it for now, if it happens again we could make it an option instead under Server Configuration.

szaimen:
> we could make it an option instead under Server Configuration.

Not sure about that because you never know when the mimetypes change/get updated in a new server update...

PR #1989, opened 2021-05-23
reorganize office scripts

szaimen:

### DIFF
--- a/apps/collabora_integrated.sh
+++ b/apps/collabora_integrated.sh
@@ -46,16 +46,6 @@ then
 We will now remove the old docker and install the app from Nextcloud instead."
     # Remove docker image
     docker_prune_this 'collabora/code'
-    # Disable RichDocuments (Collabora App) if activated
-    if is_app_installed richdocuments
-    then
-        nextcloud_occ app:remove richdocuments
-    fi
-    # Disable OnlyOffice (Collabora App) if activated
-    if is_app_installed onlyoffice
-    then
-        nextcloud_occ app:remove onlyoffice
-    fi
     # Revoke LE
     SUBDOMAIN=$(input_box_flow "Please enter the subdomain you are using for Collabora, e.g: office.yourdomain.com")
     if [ -f "$CERTFILES/$SUBDOMAIN/cert.pem" ]
@@ -172,6 +162,8 @@ if ! is_app_installed richdocuments
 then
     msg_box "The Collabora app failed to install. Please try again later."
 fi
+
+nextcloud_occ config:app:set richdocuments public_wopi_url --value="$(nextcloud_occ_no_check config:system:get overwrite.cli.url)"
 
 # Just make sure the script exits
 exit
--- a/apps/collabora_integrated.sh
+++ b/apps/collabora_integrated.sh
@@ -20,6 +20,25 @@ debug_mode
 # Check if root
 root_check
 
+# Check if Collabora is installed using the new method
+if ! is_app_installed richdocumentscode
+then
+    # Ask for installing
+    install_popup "$SCRIPT_NAME"
+else
+    # Ask for removal or reinstallation
+    reinstall_remove_menu "$SCRIPT_NAME"
+    # Removal
+    nextcloud_occ app:remove richdocumentscode
+    # Disable Collabora App if activated
+    if is_app_installed richdocuments
+    then
+        nextcloud_occ app:remove richdocuments
+    fi
+    # Show successful uninstall if applicable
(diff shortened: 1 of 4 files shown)


szaimen:
I've tested onlyoffice_integrated successfully. The rest was a simple reorganization. 👍

enoch85:
I think it's better if you explain what's going on here instead of me trying to figuring it out.

Thanks!

szaimen:
I just try to make them better readable to use use the same logic and improve some small things while doing that.
Especially the integrated onlyoffice is now much better readable, imo!

enoch85:
By readable, do you actually mean "possible to install both collabora and onlyoffice"?

szaimen:
> "possible to install both collabora and onlyoffice"?

No, thats not what I intended.
If you would actualy look at the code, you would find that out ;)

enoch85:
As far as I can see you remove Collabora instead of OnlyOffice in the Collabora script, and the opposite in the OnlyOffice script. End result = both Collabora and OnlyOffice will be installed.

Please correct me if I'm wrong.

enoch85:
> If you would actualy look at the code, you would find that out ;) 

I looked at the code. ;)

enoch85:
Wrong button, I will triple check. 

Need to install a test environment.

szaimen:
> As far as I can see you remove Collabora instead of OnlyOffice in the Collabora script, and the opposite in the OnlyOffice script. End result = both Collabora and OnlyOffice will be installed.

You are wrong! Please read the code correctly!

enoch85:
OK, so installed a fresh VPS and:

1. Installed Collabora docker (no issues)
2. Instaled Collabora integrated (no issues)
3. Installed OnlyOffice docker (Collabora app still there, but not working)
4. Installed OnlyOffice integrated (Collabora removed, OnlyOffice integrated *doesn't work*) Wrong domain

5. Uninstalled everything
6. Installed Collabora docker (no issues)
7. Installed Onlyoffcie docker (Noticed that the templates folder from Collabora is still there, other than that - all good!)
8. Installed Collabora integrated (installed but not working, maybe due to HSTS? - worked with DEMO server
9. Installed OnlyOffice Integrated (worked this time with the correct domain)

10. Installed Onlyoffice docker (no issues and domain updated correctly)
11. Installed Onlyoffice Integrated (no issues)
12. Installed Collabora docker (no issues)
13. Installed Collabora Integrated (domain value from docker still in the settings and it didn't choose the integrated option, but when I changed to use the integrated server it worked. Maybe we can force to choose the integrated version when that's installed?)
14. Installed OnlyOffice docker (tested number 3 above again, still same issue, Collabora app is still installed but not working)
15. Installed Collabora integrated (no issues)

**Errors to fix:**
1. Remove Collabora app completely when running OnlyOffice docker and integrated)
2. Remove Templates folder when uninstalling/reinstalling Collabora
3. Force to use the integrated version of Collabora when that's installed and Collabora Docker was installed before. Maybe the same for OnlyOffice?

szaimen:
> 1. Remove Collabora app completely when running OnlyOffice docker and integrated)

Should be done now

> 2. Remove Templates folder when uninstalling/reinstalling Collabora

Thats out of scope for this PR

> 3. Force to use the integrated version of Collabora when that's installed and Collabora Docker was installed before. Maybe the same for OnlyOffice?

integrated Onlyoffice should already work, collabora now too.

apps/collabora_integrated.sh
```diff
@@ -46,16 +46,6 @@ then
...
-    # Disable RichDocuments (Collabora App) if activated
-    if is_app_installed richdocuments
-    then
-        nextcloud_occ app:remove richdocuments
-    fi
-    # Disable OnlyOffice (Collabora App) if activated
-    if is_app_installed onlyoffice
-    then
-        nextcloud_occ app:remove onlyoffice
-    fi
```
enoch85, line 39: Why did you remove both of these?

szaimen: because they already get removed down below: https://github.com/nextcloud/vm/pull/1989/files#diff-9307a1a930d542adcec66849575564e9e6def7462586ff5a71da9c5eb985dc17R124-R135

szaimen: https://github.com/nextcloud/vm/blob/1caa815a8a2910cabfb9d3e19cb5fdff46bb6083/apps/collabora_integrated.sh#L124-L134

szaimen: ping @enoch85

PR #1990, opened 2021-05-23
make trusted_domain removal and addition its own function

szaimen:
Follow up to #1989

### DIFF
--- a/apps/collabora_docker.sh
+++ b/apps/collabora_docker.sh
@@ -17,6 +17,41 @@ debug_mode
 # Check if root
 root_check
 
+# TODO: remove both functions with NC21.0.3 release
+add_to_trusted_domains() {
+    local element="$1"
+    local count=0
+    print_text_in_color "$ICyan" "Adding $element to trusted domains..."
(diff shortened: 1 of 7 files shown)


enoch85:
@szaimen Is this ready?

szaimen:
no, not yet

szaimen:
Now it is ready 👍

szaimen:
Oh, we will probably run into some issue during the startup script with that because the lib will be old and not have the new needed library. How do we handle this @enoch85 ?

enoch85:
> Oh, we will probably run into some issue during the startup script with that because the lib will be old and not have the new needed library. How do we handle this @enoch85 ?

We wait to merge it until a new release is out

enoch85:
Or keep the stuff that we moved to lib for now, and then remove it in the next release instead.

szaimen:
I need to think about this

szaimen:
I think this is the way to go. With the next release we can then simply remove the functions from each script.

enoch85:
Will have a look at this when time allows and I'm not too tired. :)

PR #1991, opened 2021-05-23
Make it possible to add subdomains to deSEC

enoch85:
This script makes it possible to add a subdomain to an already existing deSEC account.

It also checks if the domain already exists, and deletes it with the option to activate it again, and thus updating the IP address with DDclient - or completley removes it.

Rate-limiting is taken into account as well.

### DIFF
--- a/addons/desec_subdomain.sh
+++ b/addons/desec_subdomain.sh
@@ -1,6 +1,6 @@
 source /var/scripts/fetch_lib.sh
 
-if [ -f "$SCRIPTS"/deSEC/.dedynauth ] && [ -f [ /etc/ddclient.conf ]
+if [ -f "$SCRIPTS"/deSEC/.dedynauth ] && [ -f /etc/ddclient.conf ]
 then
     DEDYN_TOKEN=$(grep DEDYN_TOKEN "$SCRIPTS"/deSEC/.dedynauth | cut -d '=' -f2)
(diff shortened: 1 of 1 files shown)


enoch85:
cc @peterthomassen Any comments?

cc @szaimen This is super cool!

peterthomassen:
> cc @peterthomassen Any comments?

If you had given some time to respond before merging, I would have suggested to consider using CNAME records on the subdomains, pointing one level up, so that you don't need to touch the subdomains when the IP changes.

enoch85:
@peterthomassen I tried CNAMEs but never got it working. 

I also though that RRsets are better for you guys at deSEC as it removes the need for more domains, and for the user, as the user always will try against an already existing domain.

But, I'm curious, how would you add a CNAME?

peterthomassen:
```bash
curl -X POST https://desec.io/api/v1/domains/"$DEDYN_NAME"/rrsets/ \
        --header "Authorization: Token $DEDYN_TOKEN" \
        --header "Content-Type: application/json" --data @- <<EOF
    {
      "subname": "$SUBDOMAIN",
      "type": "CNAME",
      "ttl": 3600,
      "records": ["$DEDYN_NAME."]
    }
EOF
```

enoch85:
> ["$DEDYN_NAME."]

Aah, that was the little thing I missed all along!

Thanks! Will try it out.

PR #1993, opened 2021-05-23
add TLS with the new function

enoch85:


enoch85:
@szaimen I need your brainstorming expertise in this.

Current option is to put the domain in a file under the deSEC folder. It works, but it's not smart so it doesn't know which one to choose. Imagine you want to enable Talk and Collabora, how would the function know which domain to choose? Also, we need to bring the port change if it's not default (443) to the Apache2 configs.

Another option would be to check if deSEC is installed (with the new function for that - `is_desec_installed`) in the regular `generate_cert` function and only run the DNS validation against the hook. Problem here is that you need the domain first. Should we generate that domain directly in the `generate_cert` function, and if yes - how? One option would be to change the current `desec_subdomain.sh` and check if `$1` is set so that it can be called from a function. But then again, we have the issue with the port...

A third option would be to do as in option two - change the `desec_subdomain.sh` script so that it can be called from any function or script with a subdomain option. Like if you wanted to generate a cert for Collabora then run something like `generate_desec_cert SUBDOMAIN` and that the script would take care of everything, from creating the domain, to generateing the cert.

I sat up until late this morning and tested a few options, but I don't know which is best. I added two functions so far in master, which I'm sure can be improved:
1. https://github.com/nextcloud/vm/commit/4a4edc508ba92bd02d8c3c50e8f75f20d3ca833a
2.  https://github.com/nextcloud/vm/commit/dfcb88613883a5cbec8fa122037fa241a889edc4

What do you think, maybe my brain is too tired to think right now... :D

CC @peterthomassen if you have any ideas...?

enoch85:
This was my testing (not done yet) with `collabora_docker.sh`: https://gist.github.com/enoch85/b52d8eec54a4f4c5ad55d527d4025dc8

szaimen:
Before changing the office scripts here, can we please merge #1989?
I don't want to need to handle any conflict that would probably arise!

peterthomassen:
I do not have enough context to judge your various proposals, and it's also not clear what's needed from the DNS side to set up (e.g.) Talk. Would that need a subdomain, or a full new domain?

enoch85:
Let's merge this for now, I'll come up with a better flow later on.

Points to consider.

1. Maybe we don't need to add TLS now when it's a CNAME?
2. Maybe the flow could be to run this script, then add the new domain to the app script?

enoch85:
Thought so as well, but it doesn't seem like it's generated when looking in `/etc/letsencrypt/live`.

This is a test basically anyway, will figure out something better for production.

PR #1997, opened 2021-05-26
Preparations for the next NC22.0.0 NcVM release

szaimen:
Follow up to #1990
TODO: merge this shortly before the next VM release.


szaimen:
Here the same like https://github.com/nextcloud/vm/pull/1990#issuecomment-848943167

enoch85:
@szaimen Only add the function for now and we'll use the function in the next release.

szaimen:
@enoch85 this needs to be merged shortly before the next VM release.

enoch85:
Thanks!

Can you add the changelog into this PR as well?

szaimen:
> Can you add the changelog into this PR as well?

Done 👍

CHANGELOG.md
```diff
@@ -15,6 +15,36 @@ If you run Hyper-V or want 500 GB, 1 TB or 2 TB VM you can download it from [T&M
...
+- Make sure sudo and software-properties-common is installed
+- Fixed password generation in edge cases
+- Reworked the cookielifetime script
+- SMB-server - added option to automatically empty recylce bins
+- SMB-server - added option to empty all recycle bins
+- Minor bugfixes and improvements
+- And more...
+
+### Full changelog:
+- https://github.com/nextcloud/vm/compare/21.0.2..21.0.3
```
szaimen, line 45: Hm... I don't think that this will work. Do I need to change this to https://github.com/nextcloud/vm/compare/21.0.2..enh/function-removal-docker-documentserver ?

enoch85: I'm not sure what you mean?

szaimen: What I mean is the compare link
https://github.com/nextcloud/vm/compare/21.0.2..21.0.3

I dont think that it will work because it compares branches.
So I think it would need to be https://github.com/nextcloud/vm/compare/21.0.2..enh/function-removal-docker-documentserver in this case, no?

enoch85: Does this link work? https://github.com/nextcloud/vm/compare/21.0.0..21.0.1

In other words, it will work once it's merged. :)

szaimen: Hm... I don't think so since this branch isn't called 21.0.3

szaimen: Okay, seems like you are not comparing branches but release tags.

CHANGELOG.md
```diff
@@ -37,6 +37,8 @@ PR's are more than welcome. Happy Nextclouding!
 - Fixed password generation in edge cases
 - Reworked the cookielifetime script
 - Updated geoblock database - June 2021
+- Added option to check for 0-byte files
+- Changed from apt to apt-get
```
enoch85, line 41: thanks @szaimen!

enoch85:
Now when thinking about it, we put in some safe-measures for NC 19 I think, maybe it would a good idea to remove them now, or should we keep them?

szaimen:
> Now when thinking about it, we put in some safe-measures for NC 19 I think

Can you send some pointers? I am not sure if I know which ones you are talking about.

enoch85:
NVM, can't find it except this: https://github.com/nextcloud/vm/blob/master/lib.sh#L114

But ppl are still on NC 13 (that I'm migrating), so let's keep it.

PR #2001, opened 2021-05-29
facerecognition - use curl instead of wget

### LINKED ISSUE #2000: Face Recognition install Script Error.
the script fail to add the key automatically.

+ echo 'deb https://repo.delellis.com.ar focal focal'
+ sudo apt-key add -
+ wget -qO - https://repo.delellis.com.ar/repo.gpg.key
gpg: no valid OpenPGP data found.

szaimen:
Fix #2000

### DIFF
--- a/apps/face-recognition.sh
+++ b/apps/face-recognition.sh
@@ -90,7 +90,7 @@ fi
 # Install requirements
 # https://github.com/matiasdelellis/facerecognition/wiki/Installation#ubuntu-focal
 echo "deb https://repo.delellis.com.ar focal focal" > /etc/apt/sources.list.d/20-pdlib.list
-wget -qO - https://repo.delellis.com.ar/repo.gpg.key | sudo apt-key add -
+curl -fsSL https://repo.delellis.com.ar/repo.gpg.key | sudo apt-key add -
 apt update -q4 & spinner_loading
 install_if_not php7.4-pdlib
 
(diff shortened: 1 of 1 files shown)


apps/face-recognition.sh
```diff
@@ -90,7 +90,13 @@ fi
 # Install requirements
 # https://github.com/matiasdelellis/facerecognition/wiki/Installation#ubuntu-focal
 echo "deb https://repo.delellis.com.ar focal focal" > /etc/apt/sources.list.d/20-pdlib.list
-wget -qO - https://repo.delellis.com.ar/repo.gpg.key | sudo apt-key add -
+if ! curl -fsSL https://repo.delellis.com.ar/repo.gpg.key | sudo apt-key add -
```
enoch85, line 93: I was thinking the exact same thing actually, again. :D

But, I think maybe we should use the `curl_to_dir` function instead and improve that function for all the scripts with something like this.

szaimen: hm... can we do that in a follow up? 
It is honestly a bit out of scope for this PR and should better be done in a different PR, imo.

szaimen:
This is ready 👍

enoch85:
This was what I was thinking, or similar: https://github.com/nextcloud/vm/pull/2006

szaimen:
Will be handled in #2006

Issue #2002, opened 2021-05-29
Feedback via Mail


szaimen:
I purchased the 2TB script yesterday for NExtCloud and BItwarden. I have run into a number of errors and have spent almost 2 full work days trying to get your scripts to function correctly.

I purchased the 2TB NextCloud OVA. Imported the OVA into vCenter. Increased CPU to 8 vcpu and 12gb vram. Power on the VM.

Geoblock error

![image](https://user-images.githubusercontent.com/42591237/120068828-10cc3800-c083-11eb-9dfa-54d448a713cc.png)

SMTP Relay for reference for later Bitwarden Mail Server Config Bugs, Manual setup, NO Encryption, Port 25, No username or Password. Test email does arrive.

![image](https://user-images.githubusercontent.com/42591237/120068872-2b9eac80-c083-11eb-91aa-e4db38bff3d0.png)

These are the apps package options

![image](https://user-images.githubusercontent.com/42591237/120068876-31948d80-c083-11eb-9847-f78390d6d227.png)

Bitwarden Setup Options

![image](https://user-images.githubusercontent.com/42591237/120068883-378a6e80-c083-11eb-9164-604f67c578df.png)

I followed the bitwarden instructions. Made sure to say no in the correct places. The apache based TLS Lets Encrypt cert installs fine. I pull up the bitwarden page as it says to in the settings and I create my first account. This completes with no issue but when I try to log that account in we see this error: 

![image](https://user-images.githubusercontent.com/42591237/120068891-3f4a1300-c083-11eb-9b52-120dc3a39c7f.png)

I waited a few minutes tried again and could login.

Then the Bitwarden Mail Config. I have tried this using the default global config as already configured and just hit OK (just leave the box empty as it states in instructions)

![image](https://user-images.githubusercontent.com/42591237/120068895-496c1180-c083-11eb-9580-e94a4818e83a.png)

But then it errors. Empty is not allowed….

![image](https://user-images.githubusercontent.com/42591237/120068899-4ec95c00-c083-11eb-9810-ada514c59d4e.png)

[...]

PR #2004, opened 2021-05-29
bitwarden - add hint that you need to wait

szaimen:
Fix part of #2002


szaimen:
This is ready 👍

apps/bitwarden_mailconfig.sh
```diff
@@ -207,8 +204,8 @@ then
 fi
 
 # Start Bitwarden
-start_if_stopped bitwarden
+check_command systemctl restart bitwarden
 msg_box "Your Bitwarden mailserver settings should be successfully changed by now.
-
+Bitwarden is now restarting. This can take a few minutes. Please wait until it is done.\n
```
enoch85, line 209: I think it would be smarter to add a check for the PID to go away, or to check the status of docker (stopped or running) and then continue.

szaimen: Maybe it is smarter but I currently don't know how to implement that...
Maybe you can have a look?

enoch85: https://stackoverflow.com/a/43723174

szaimen: Please go ahead 👍

enoch85:
Not tested, but that's how I imagine it would work

enoch85:
image

szaimen:
> image

I've reverted my commit.
But in this case it probably needs to be changed in bitwarden_registration.sh, too?

enoch85:
Hmm, is that only restarting?

szaimen:
> Hmm, is that only restarting?

Yes

PR #2005, opened 2021-05-29
refactor the bitwarden_mailconfig script

### LINKED ISSUE #1453: adjust the bitwarden-mailconfig script to fit the new input_box_flow function
follow up to #1449

### LINKED ISSUE #2002: Feedback via Mail

szaimen:
Follow up to #2004
Fix #1453
Fix #2002 (last part of it)
Disclaimer: I didn't test this but it should already work.

### DIFF
--- a/apps/bitwarden_mailconfig.sh
+++ b/apps/bitwarden_mailconfig.sh
@@ -166,6 +166,7 @@ fi
 
 # Stop bitwarden
 systemctl stop bitwarden
+countdown "Waiting for Bitwarden to stop..." "30"
 
 # Write to files
 # mailserver
--- a/apps/bitwarden_mailconfig.sh
+++ b/apps/bitwarden_mailconfig.sh
@@ -120,7 +120,9 @@ then
     MAIL_PASSWORD=$(input_box_flow "Please enter the SMTP password to your email provider.")
 fi
 
-ADMIN_ACCOUNT=$(input_box_flow "Please enter mail accounts, that should have access \
+while :
+do
+    ADMIN_ACCOUNT=$(input_box "Please enter mail accounts, that should have access \
 to the Bitwarden admin-panel, reachable under https://your-bitwarden-domain/admin/.
 They don't have to be registered Bitwarden accounts.
 To make this setting work, your Bitwarden mailserver settings have to be correct.
@@ -128,6 +130,17 @@ You can enter just one e-mail address or enter more than one like so:
 'bitwarden@example.com,bitwarden2@example1.com,bitwarden3@example2.com'
 If you want to keep the admin accounts that are already configured inside the \
 global.override.env-file, just leave the box empty.")
+
+    if [ -n "$ADMIN_ACCOUNT" ]
+    then
+        if yesno_box_yes "Does this look correct: $ADMIN_ACCOUNT"
+        then
+            break
+        fi
+    else
+        break
+    fi
+done
 
 # Present what we gathered, if everything okay, write to files
 msg_box "These are the settings that will be used. Please check that everything seems correct.
--- a/apps/bitwarden_mailconfig.sh
+++ b/apps/bitwarden_mailconfig.sh
@@ -58,61 +58,68 @@ then
     echo "globalSettings__mail__smtp__startTls=false" >> "$BITWARDEN_HOME"/bwdata/env/global.override.env
 fi
 
-# Enter mailserver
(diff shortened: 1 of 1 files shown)


enoch85:
One error with this is that `input_box_flow` can't be empty here: https://github.com/nextcloud/vm/pull/2005/files#diff-a6390883bb894de29dc43c3c4337f2d8b4b5c08a0057076a91344414939dd8ffR130

enoch85:
Just tested and it seems to work.

Will fix the same fix as in the static IP script: https://github.com/nextcloud/vm/blob/master/network/static_ip.sh#L141

szaimen:
> One error with this is that `input_box_flow` can't be empty here: https://github.com/nextcloud/vm/pull/2005/files#diff-a6390883bb894de29dc43c3c4337f2d8b4b5c08a0057076a91344414939dd8ffR130

Will fix that later 👍

enoch85:
It's now fixed. If argument tested and works.

I feel confident with merging this now.

enoch85:
@szaimen Please give your OK before I merge.

apps/bitwarden_mailconfig.sh
```diff
@@ -58,125 +58,125 @@ then
...
-        then
-            DEFAULT_PORT=""
-            break
-        else
-            msg_box "The answer wasn't correct. Please type in 'SSL', 'STARTTLS', 'none' or leave the input box empty."
         fi
+    else
+        break
     fi
 done
```
szaimen, line 143: Do we really want allow to leave this empty? Otherwise I would just use input_box_flow here and remove the option to leave this empty.

szaimen:
Apart from my comment it is ready to go 👍

apps/bitwarden_mailconfig.sh
enoch85: Yeah, let's do it like this.

PR #2006, opened 2021-05-29
add more functionality to curl_to_dir

enoch85:


lib.sh
```diff
@@ -511,7 +511,14 @@ then
     mkdir -p "$3"
 fi
     rm -f "$3"/"$2"
-    curl -sfL "$1"/"$2" -o "$3"/"$2"
+    if ! curl -sfL "$1"/"$2" -o "$3"/"$2"
+    then
+        msg_box "It seems like the server for the download isn't reachable, or that a temporary error occurred, please try again later."
+        if yesno_box_yes "Do you want to exit this script?"
```
szaimen, line 517: We shouldn't allow to not exit here, imho. Because it probably would break some scripts.

enoch85: that's why it's a question, we could add a warning like: "CAUTION, answering 'yes' will break the current operation!"

apps/face-recognition.sh
```diff
@@ -87,10 +87,20 @@ then
...
+This is most likely just a temporary downtime of the server where the key comes from.
+Please try again later!"
+    exit
+fi
+
 # Install requirements
 # https://github.com/matiasdelellis/facerecognition/wiki/Installation#ubuntu-focal
 echo "deb https://repo.delellis.com.ar focal focal" > /etc/apt/sources.list.d/20-pdlib.list
-wget -qO - https://repo.delellis.com.ar/repo.gpg.key | sudo apt-key add -
+curl_to_dir https://repo.delellis.com.ar repo.gpg.key . | sudo apt-key add -
```
szaimen, line 103: Why not add it under `/etc/apt/trusted.gpg.d/` ?

enoch85: Yeah why not? :)

Please change.

lib.sh
enoch85: Made it a while loop instead

szaimen:
@enoch85 does the while loop work when running the update script via cronjob or does it end in an endless loop?

enoch85:
> @enoch85 does the while loop work when running the update script via cronjob or does it end in an endless loop?

Good point! 

Will think about it some more. Please come up with a better suggestion if you have one.

apps/face-recognition.sh
szaimen: Can we do the facerecognition part in https://github.com/nextcloud/vm/pull/2001?

enoch85:
the latest commit makes it retry for 10 attempts, tgen just continue as normal, adding some info that you wouldn't get if you just ran the normal `curl` flags (that does the same thing)

apps/face-recognition.sh
enoch85: this improves all the scripts, and when we merge this we would change that part anyway so I think it's better to continue here.

enoch85:
tested and works (the function)

enoch85:
@szaimen Ok with you to merge?

szaimen:
Should be ready to go now 👍

lib.sh
```diff
@@ -511,7 +511,21 @@ then
...
+        if ! curl -sfL "$1"/"$2" -o "$3"/"$2"
+        then
+            msg_box "We just tried to fetch '$1/$2', but it seems like the server for the download isn't reachable, or that a temporary error occurred. We will now try again.
+
+Please report this issue to $ISSUES"
+            retries=$((retries+1))
+            print_text_in_color "$ICyan" "$retries of 10 retries."
+            countdown "Trying again in 30 seconds..." "30"
+        else
+            break
```
szaimen, line 526: @enoch85 we need to exit here, or at least return 1, no?

enoch85: No, this breaks if it works.

szaimen: You are right. But it continues after trying 10 times...

enoch85: Yes it does. As you said it might not be a good idea to just exit, which makes sense (but at the same time not, because if the file isn't there then the script will break anyway).

How should we handle this?

szaimen: See #2009 how I would handle that

PR #2007, opened 2021-05-29
docker documentserver - don't restart docker

szaimen:
The restart is unnecessary here and just leads to needing to wait for the container to come online, imo.

### DIFF
--- a/apps/collabora_docker.sh
+++ b/apps/collabora_docker.sh
@@ -366,10 +366,6 @@ then
     add_to_trusted_domains "$SUBDOMAIN"
     # Add prune command
     add_dockerprune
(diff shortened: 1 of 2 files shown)


enoch85:
IIRC we need to restart the docker containers for Collabora and OnlyOffice, else they won't work. 

We added for a reason anyway, I'm 100% sure of that.

szaimen:
Okay but do we really need to restart the docker service completely?
Isn't the restart of the docker container enough?

enoch85:
I'm fine with only restarting the container.

szaimen:
Done 👍

enoch85:
Perfect!

PR #2009, opened 2021-05-30
curl_to_lib - exit if tried 10 times

szaimen:


enoch85:
Sure, this works, but it's not nice to try for the wrong file 10 times and then ask the user to choose.

I think we should have a look at the function calling `curl_to_dir()`

szaimen:
@enoch85 Please have a look at https://github.com/nextcloud/vm/pull/2010 for the fix that you are looking for

PR #2010, opened 2021-05-30
Fix downloading other files than sh files

### LINKED ISSUE #2008: master/static/index.sh
## Steps to reproduce
1. Nextcloud Install script
2.
3.

## Expected behavior
We just tried to fetch 'https://raw.githubusercontent.com/nextcloud/vm/master/static/index.sh', but it seems like the server for the download isn't reachable, or that a temporary error occurred. We will   │
│ now try again.  

## Actual behavior
Tell us what happens instead

## Server configuration

**PHP Version** (e.g. 7.4)

## Network
**Do you use DHCP?**

```
Insert your nextcloud.log file here
```

```
Insert the CLI output here
```

szaimen:
Fix #2008

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -1265,6 +1265,10 @@ rm -f releases
 # Use it for functions like download_static_script
 download_script() {
     rm -f "${SCRIPTS}/${2}.sh" "${SCRIPTS}/${2}.php" "${SCRIPTS}/${2}.py"
+    if [ -n "$4" ]
+    then
+        curl_to_dir "${!1}" "${2}.$4" "$SCRIPTS"
+    fi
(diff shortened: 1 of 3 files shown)


enoch85:
Also works, but is just a workaround. There must be a better way.

szaimen:
Please have a look yourself then. Thats the best I was able to come up with.

szaimen:
In my opinion it is not really a workaround because those are the only places where we have other files than .sh files. So we should be able to change the download_script function to only download sh files if not explicitely defined.

enoch85:
Yes, but the whole idea with `download_script` and `run`script` is to be able to call any of the file endings we add. 

Right now I'm looking into if we can somehow check if any of those functions are used in `curl_to_dir`, or if we can check by the download URL if it contains php or something else. Then we could make it generic.

PR #2011, opened 2021-05-30
fix curl_to_dir for use in function

enoch85:
This adds a variable to the `download_script` function which is then picked up in `curl_to_dir` and executes differently if set.

Tested and works!


enoch85:
@szaimen Please test this

enoch85:
I'll merge it since I've tested several times and the scripts currently are broken.

PR #2012, opened 2021-05-31
restart notify push in some situations

szaimen:


szaimen:
This is ready 👍

enoch85:
Just curios, why is this needed?

szaimen:
> Just curios, why is this needed?

Because no automatic restart of the service is implemented in the service itself, afaik. So when upgrading the app or Nextcloud and/or stopping Nextcloud, I suppose the service will not work anymore and needs to be restarted.

PR #2014, opened 2021-06-08
fix passwords

szaimen:
Having an equal character at the beginning of a password can sometimes lead to invalid password authentification.


szaimen:
It seems like Github workflows have a hard time today but it is definitely not related to my fix.

PR #2020, opened 2021-06-11
rework cokielifetime

szaimen:

### DIFF
--- a/addons/cookielifetime.sh
+++ b/addons/cookielifetime.sh
@@ -25,29 +25,29 @@ $MENU_GUIDE\n\n$RUN_LATER_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 "1800s" "30 minutes" \
 "7200s" "2 hours" \
 "43200s" "12 hours" \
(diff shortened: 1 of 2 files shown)


addons/cookielifetime.sh
```diff
@@ -68,14 +68,15 @@ You can not set a value below 30 minutes (1800 seconds).")
...
             fi
         done
     ;;
     *)
     ;;
 esac
+
+# Set the value
+nextcloud_occ config:system:set remember_login_cookie_lifetime --value="$COOKIE_LIFETIME"
+nextcloud_occ config:system:set session_lifetime --value="$COOKIE_LIFETIME"
```
szaimen, line 81: After a more indepth testing I found out that you get logged out if either of those values gets reached due to inactivity. This was so long not clear because session_lifetime is by default set to 24h which makes testing this a bit hard. But it gets obvious if you test with 5min or 10min intervals.

szaimen:
This is ready 👍

enoch85:
Makes sense.

Issue #1534, opened 2020-10-06
Report 0 byte files


enoch85:
```
filesystems[0]="$NCDATA"
cd "$NCDATA"

emSub="Warning: Nextcloud Contains 0 Byte Files"

for fs in "${filesystems[@]}"; do

        while read -d '' -r; do
                arr+=( "$REPLY\n" )
                done < <(find $fs -size 0 -print0)

done

if [[ ! -z ${arr[@]} ]]; then
        send_mail "emSub" "${arr[@]}"
fi
```

Credits to @Wdavery

enoch85:
Maybe add a function to check all users in `NCDATA` as well?

If we're going to scan the whole `NCDATA` folder, it might take time!

szaimen:
> send_mail "emSub" "${arr[@]}"

I guess this should be
`send_mail "$emSub" "${arr[@]}"`

szaimen:
Is this still needed?

enoch85:
It's a feature to check if Nextcloud produces 0-byte files, which actually happens from time to time with the desktop client. You should probably check for your self. 

We could add it in one of the menus, with a big warning that it will take time, especially if the data folder is large.

szaimen:
So this is not meant to run continuosly in the bacground?

enoch85:
Correct. :+1:

enoch85:
This is the issue: https://github.com/nextcloud/server/issues/3056

szaimen:
I've just checked my personal folder via SMB and it seems like at least my account is not affected...

enoch85:
> my personal folder

Do you mean your complete data mount?

> via SMB

Do you use SMB as the main data drive mounted in `/etc/fstab`?

szaimen:
> > my personal folder
> 
> Do you mean your complete data mount?

Yes

> > via SMB
> 
> Do you use SMB as the main data drive mounted in `/etc/fstab`?

All my data is on an external BTRFS drive that is encrypted with Veracrypt. Every user has a personal folder on this drive.
I have created a SMB-share for each personal folder on this drive. The same directory also gets mounted to Nextcloud using the local external storage feature. This way I can access the same files via Nextcloud and SMB.
Here is a bit more indepth explation on this: https://szaimen.github.io/Nextcloud-NAS-Guide/docs/create-shares

enoch85:
Most of the reporters in that issue are saying that it's due to the Desktop client, or Android client. 

[...]

PR #2026, opened 2021-06-17
check for 0-byte files

enoch85:
Fix https://github.com/nextcloud/vm/issues/1534

### DIFF
--- a/addons/0-byte-files.sh
+++ b/addons/0-byte-files.sh
@@ -26,11 +26,14 @@ done
 # NEWARR="$(echo ${arr[*]} | sed s/'appdata'//)"
 
 # Notify!
-if [[ -n "${arr[*]}" ]]
+if [ -n "${arr[*]}" ]
 then
-    notify_admin_gui "$emSub" "${arr[*]}"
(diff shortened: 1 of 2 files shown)


addons/0-byte-files.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# TODO
+# Remove known 0-byte files
+# Everything in appdata isn't important
+# NEWARR="$(echo ${arr[*]} | sed s/'appdata'//)"
+
+# Notify!
+if [ -n "${arr[*]}" ]
+then
+    send_mail "$emSub" "${arr[*]}"
```
szaimen, line 43: @enoch85 this won't work without having smtp-mail set up. I'd recommend to test if it is installed by simply doing something like this: 
https://github.com/nextcloud/vm/blob/5b04abac6e54e0f6383f96001615af1c13ac655a/not-supported/daily-backup-wizard.sh#L154-L158

enoch85: At the same time it won't give any output if not installed, so if they have it it works, if not, then too bad. :)

Issue #2015, opened 2021-06-09
`apt` is used by scripts instead of `apt-get`


paulmenzel:
`apt` should be used in interactive scenarios, while `apt-get` should still be used in scripts.

```
$ git grep 'apt install'
apps/clamav.sh:apt install clamav clamav-freshclam clamav-daemon -y
lib.sh:    apt update -q4 & spinner_loading && apt install dnsutils -y
lib.sh:    apt update -q4 & spinner_loading && apt install net-tools -y
lib.sh:    apt update -q4 & spinner_loading && RUNLEVEL=1 apt install "${1}" -y
network/ddclient-configuration.sh:DEBIAN_FRONTEND=noninteractive apt install ddclient -y
nextcloud_install_production.sh:    apt install curl -y
nextcloud_install_production.sh:    apt install lshw -y
nextcloud_install_production.sh:    apt install net-tools -y
nextcloud_install_production.sh:    apt install whiptail -y
nextcloud_install_production.sh:apt install postgresql -y
nextcloud_install_production.sh:check_command apt install apache2 -y
nextcloud_install_production.sh:check_command apt install -y \
nextcloud_install_production.sh:        apt install -y --install-recommends \
not-supported/firewall.sh:    DEBIAN_FRONTEND=noninteractive apt install ufw -y --no-install-recommends
not-supported/remotedesktop.sh:            apt install gnome-session --no-install-recommends -y
not-supported/remotedesktop.sh:                    apt install makemkv-oss makemkv-bin -y
not-supported/restore-backup.sh:    apt install veracrypt --no-install-recommends -y
not-supported/restore-backup.sh:    DEBIAN_FRONTEND=noninteractive apt install samba -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"
not-supported/tpm2-unlock.sh:apt install clevis-tpm2 clevis-luks clevis-initramfs -y
not-supported/veracrypt-btrfs.sh:    apt install veracrypt --no-install-recommends -y
not-supported/veracrypt-ntfs.sh:    apt install veracrypt --no-install-recommends -y
```

enoch85:
You run this on Debian, while we made if for Ubuntu. If you want to maintain and develop a VM with Debian, that would be super cool. If not, please use Ubuntu as intended.

Thanks!

paulmenzel:
> You run this on Debian, while we made if for Ubuntu.

[...]

PR #2027, opened 2021-06-18
change to apt-get instead of apt

enoch85:
This fixes https://github.com/nextcloud/vm/issues/2015

Let's go back to what's best practice.


enoch85:
@szaimen @paulmenzel Any comments?

PR #2032, opened 2021-06-22
Upgrade to PostgreSQL 13

enoch85:
It's now supported in NC 21: https://docs.nextcloud.com/server/21/admin_manual/installation/system_requirements.html

https://www.postgresql.org/about/news/postgresql-13-released-2077/

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -341,10 +341,10 @@ $MENU_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
 done
 
 # Install PostgreSQL
-# sudo add-apt-repository "deb http://apt.postgresql.org/pub/repos/apt/ bionic-pgdg main"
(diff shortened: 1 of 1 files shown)


szaimen:
Are you sure about adding a 3rd party repository to the sources?
I thought better stable than bleeding edge? 😉

enoch85:
> I thought better stable than bleeding edge? wink

It's supported by Nextcloud, so not bleeding edge IMHO.

> Are you sure about adding a 3rd party repository to the sources?

it's PostgreSQL's repo. I use it when migrating DB and it's basically the same thing as Ubuntu if not better.

What do you think?

szaimen:
I think it is fine from a security standpoint. Question is if it might have some implications on backups/updates.

enoch85:
Hmm, let's keep it as is. 

If one wants to upgrade it's possible, but stable stable stable or more important than "nice to have" - even if it brings some amazing futures.

Thanks for your POV.

Issue #2039, opened 2021-07-01
Memcache \OC\Memcache\APCu not available for local cache


ozinfotech:
## Steps to reproduce
1. Open PuTTY and log in to Nextcloud
2. Update the VM with the script: `sudo bash /var/scripts/menu.sh`
3. Attempt a minor release update

## Expected behavior
The update usually completes without any hiccups.

## Actual behavior
The script errors out with the following message:
```
An unhandled exception has been thrown:
OC\HintException: [0]: Memcache \OC\Memcache\APCu not available for local cache (Is the matching PHP module installed and enabled?)
```

## Server configuration

**Nextcloud server version:** 21.0.2

**Server OS** 20.04.2 LTS

**PHP Version** 7.4

**How did you install the VM?** (Scripted install from master **OR** Released version)
Purchased from TechAndMe.se

## Network
**Do you use DHCP?**
No

**Is port 80 and/or 443 open?**
Yes to both

#### Log file (/var/log/nextcloud/nextcloud.log)
```
{"reqId":"HlrzHllugT6koEfqzcoJ","level":2,"time":"2021-06-23T18:55:02-05:00","remoteAddr":"","user":"--","app":"OC\\Log\\Rotate","method":">
{"reqId":"wxFT1zguKoQ62axvDr3R","level":3,"time":"2021-06-23T19:02:40-05:00","remoteAddr":"213.70.130.70","user":"6DE4C6DA-602E-486D-8389-A>
{"reqId":"WLafzNgke46jPytF22Zl","level":3,"time":"2021-06-29T14:11:28-05:00","remoteAddr":"213.70.126.219","user":"6DE4C6DA-602E-486D-8389->
{"reqId":"muUEtUp6kTGOsc4fZQBk","level":2,"time":"2021-06-30T10:47:23-05:00","remoteAddr":"213.69.71.49","user":"0269457C-E7F6-432A-A0A5-18>
{"reqId":"E1pzMrN6IQocSuucHLEo","level":3,"time":"2021-06-30T12:24:46-05:00","remoteAddr":"162.41.74.90","user":"6DE4C6DA-602E-486D-8389-A>
{"reqId":"FUU64O3lpfJ6QsJt3c0p","level":3,"time":"2021-06-30T12:28:44-05:00","remoteAddr":"162.41.74.210","user":"6DE4C6DA-602E-486D-8389->
[...]

PR #2040, opened 2021-07-02
Remove APCu and replace with Redis instead

enoch85:
Fix https://github.com/nextcloud/vm/issues/2039

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -55,6 +55,16 @@ If you are sure that no update or backup is currently running, you can fix this 
     exit 1
 fi
 
+# Change from APCu to Redis for local cache
+# https://github.com/nextcloud/vm/pull/2040
+if pecl list | grep redis >/dev/null 2>&1
+then
+    sed -i "/memcache.local/d" "$NCPATH"/config/config.php
+    nextcloud_occ config:system:set memcache.local --value='\OC\Memcache\Redis'
+else
+   nextcloud_occ config:system:delete memcache.local
+fi
+
 # Create a snapshot before doing anything else
 check_free_space
 if ! [ -f "$SCRIPTS/nextcloud-startup-script.sh" ] && (does_snapshot_exist "NcVM-startup" \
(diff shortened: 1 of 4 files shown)


szaimen:
Shouldn't we fix this in old installations via the update script, too? I have the feeling that this will fix some issues in the future, too...

szaimen:
Question is then if switching from apcu to redis during the update will produce some issues, itself...

enoch85:
> Question is then if switching from apcu to redis during the update will produce some issues itself...

We have to test this before of course.

I will fire up the latest release now just to see if it works. Can you do the same please? @szaimen

szaimen:
What is still missing, afaics is switching from 'memcache.local' => '\\OC\\Memcache\\APCu', to 'memcache.local' => '\\OC\\Memcache\\Redis', in the config.php file during the update script...

nextcloud_update.sh
```diff
@@ -324,13 +324,41 @@ then
     check_command phpenmod -v ALL redis
 fi
 
-# Upgrade APCu and igbinary
-#if [ "${CURRENTVERSION%%.*}" -ge "17" ]
-if [ "${CURRENTVERSION%%.*}" -ge "23" ]
+# Change from APCu to Redis for local cache
+if pecl list | grep redis >/dev/null 2>&1
+then
+    nextcloud_occ config:system:set memcache.local --value='\OC\Memcache\Redis'
```
enoch85, line 330: @szaimen

szaimen: ah sorry, didn't see this.
But shouldn't more likely be called
`nextcloud_occ config:system:set memcache.local --value='\\OC\\Memcache\\Redis'`

(notice the double slashes)

enoch85: Nope :)

enoch85:
Tested and works!

enoch85:
Tested on multiple instances now, all is good. Nice!

matthieu-labas:
I wanted to comment on issue #2039 but it's locked so I figured I can say it here: it *can* work with APCu if you add a `--define apc.enable_cli=1` to the php command line, whatever it is (in my case it was the cron).

From https://help.nextcloud.com/t/occ-wont-run-with-memcache-apcu/119724/3:

    sudo -u www-data php --define apc.enable_cli=1 -f "/opt/nextcloud/cron.php"

enoch85:
@matthieu-labas I'm sure it works, but now we decided to not use it. We have been running Redis successfully for years, and then one of the core devs told us that APCu is better to run for local cache (yes it felt snappier) but then the major bug happened where it just stopped working (Nextcloud bug).

In other words, APCu as cache doesn't feel reliable, and we don't want to maintain something that is broken in Nextcloud, and constantly come up with workarounds to fix it. Redis on the other hand never failed, there are _0_ bug reports regarding it, and we made this VM since 2016.

enoch85:
If you had issues with Redis, I really think you should try this: https://github.com/nextcloud/vm/blob/master/addons/redis-server-ubuntu.sh ;)

PR #2045, opened 2021-07-11
Update and rename bitwarden-rs.sh to vaultwarden.sh

ke-lin:
Since bitwarden_rs was renamed to vaultwarden and the docker bitwarden_rs image is now deprecated, I have changed everything to the new image and name. The changes include:
- New script name
- Added information on how to update to new image
- Check if vaultwarden is already installed
- Changed all file location / configs to vaultwarden
- Changed docker image to vaultwarden


enoch85:
Cc @szaimen

enoch85:
@ke-lin Thanks a ton for your PR!

Would you mind signing it so that DCO passes?

Thanks!

szaimen:
Please don't merge this yet. 

It is actually not how I would approach this change. I will create a PR which will address this in a different way.

enoch85:
>  I will create a PR which will address this in a different way.

@szaimen Please work in the same PR if possible.

szaimen:
@ke-lin either way, many thanks from my side, too! :)

enoch85:
@szaimen Could you please comment the latest changes?

ke-lin:
@enoch85 the commits are signed now.

szaimen:
So my approach on this one is to not rename the internally used name and directory and instead just change the publicly used name. Doing it differently will need a lot of fixes at many places.

ke-lin:
> 
> 
> So my approach on this one is to not rename the internally used name and directory and instead just change the publicly used name. Doing it differently will need a lot of fixes at many places.

Wouldn't it make more sense to create a dedicated script, which handles migration or upgrade, which can be triggered if the already installed check is true?
This would keep a clean naming convention, especially for new installation, e.g. fail2ban commands. 
At the end we need to change only internally used name and directories for the docker data location and fail2ban config or do I miss something?

szaimen:
> Wouldn't it make more sense to create a dedicated script, which handles migration or upgrade, which can be triggered if the already installed check is true? 
> This would keep a clean naming convention, especially for new installation, e.g. fail2ban commands.
> At the end we need to change only internally used name and directories for the docker data location and fail2ban config or do I miss something?

Such a dedicated script already exists but the problem is that it is possible that the script isn't run regularly hence the old path and old configuration could still exsits hence we would need to check for the old and the new configuration everywhere.
Just a few additional pointers that come to my mind (apart from the admin-panel and the install script itself):
https://github.com/nextcloud/vm/blob/ed0c57abddb76c878b97cfcbfe4dc1773f167f6d/nextcloud_update.sh#L483
https://github.com/nextcloud/vm/blob/ed0c57abddb76c878b97cfcbfe4dc1773f167f6d/apps/fail2ban.sh#L37
https://github.com/nextcloud/vm/blob/ed0c57abddb76c878b97cfcbfe4dc1773f167f6d/not-supported/restore-backup.sh#L439

ke-lin:
> 
> 
> > Wouldn't it make more sense to create a dedicated script, which handles migration or upgrade, which can be triggered if the already installed check is true?
> > This would keep a clean naming convention, especially for new installation, e.g. fail2ban commands.
> > At the end we need to change only internally used name and directories for the docker data location and fail2ban config or do I miss something?
> 
> Such a dedicated script already exists but the problem is that it is possible that the script isn't run regularly hence the old path and old configuration could still exsits hence we would need to check for the old and the new configuration everywhere.
> Just a few additional pointers that come to my mind (apart from the admin-panel and the install script itself):
> 
> https://github.com/nextcloud/vm/blob/ed0c57abddb76c878b97cfcbfe4dc1773f167f6d/nextcloud_update.sh#L483
> 
> https://github.com/nextcloud/vm/blob/ed0c57abddb76c878b97cfcbfe4dc1773f167f6d/apps/fail2ban.sh#L37
> 
> https://github.com/nextcloud/vm/blob/ed0c57abddb76c878b97cfcbfe4dc1773f167f6d/not-supported/restore-backup.sh#L439

Thanks for the inside.
If we would put the upgrading and renaming logic into nextcloud_update.sh we woulnd`t have this issue. Since everybody, which do not run the script will be on the old name and image, everybody who already has run the script, is on the new one. Right? 
And yes, we might need to include a double check in fail2ban.sh and restore-backup.sh (maybe in other scripts too), but would this be a problem?
I just thought having a clean naming convention would be beneficial, especially in the long term. However, if you have any concerns in regards to renaming, we should not proceeded like this.

szaimen:
> I just thought having a clean naming convention would be beneficial, especially in the long term. 

In general, is this the way to go, yes.

> However, if you have any concerns in regards to renaming, we should not proceeded like this.

I let @enoch85 decide if it is worth the effort.

enoch85:
@ke-lin @szaimen I think you can decide what's best.

My POV, take the best stuff from both PRs and co-op on one of the two. Preferably this one so that @ke-lin gets the contributor status he/she deserves.

ke-lin:
@szaimen if it is fine with you, I would take your approach in regards to the nextcloud_update.sh script and amend it with the following:
- copy and rename data location  
- update container config with new name
- update fail2ban config

Afterwards i would also add references to fail2ban.sh and restore-backup.sh and check, if there are further location it might be referenced at.
What do you think?

szaimen:
the plan sounds good to me! Please make sure to always check for the old and new config.
e.g. here 
https://github.com/nextcloud/vm/blob/ed0c57abddb76c878b97cfcbfe4dc1773f167f6d/apps/fail2ban.sh#L37
it should be 
```
if ! does_this_docker_exist bitwarden_rs && ! does_this_docker_exist vaultwarden
```
now, that we make everything consistant.
I also add some things that will not work below.

szaimen:
Here is a different plan since what I wrote in the comment before will be too laborious after thinking about it a bit more:
I will make all needed changes that I think are necessary in order to make this PR mergable.
This will include reverting some smaller changes that you made.

@ke-lin are you fine with that?

ke-lin:
@szaimen sure, that is absolutely fine with me.

szaimen:
Okay, then I'll go ahead but I will probably need some days until I get to it.

enoch85:
@szaimen Hey, any update on this?

I would like to merge this pretty soon, and if you don't have time, maybe we can merge it as is and do changes later?

szaimen:
Sry, I was unfortunatley pretty busy the last weeks.
Problem is that the PR is unfortunatley not mergeable in the current state, imho, as we loose all backwards compatibility.

szaimen:
I'll try to work on this this weekend

szaimen:
I'll work on this now

szaimen:
This should be it but I haven't everything

enoch85:
@szaimen I trust your judgment on this one.

enoch85:
@ke-lin Can you please test this properly now once it's merged? 

1. Install Bitwarden RS, then run the update script. (Bitwarden RS should be updated)
2. Run the new Vaultwarden script to reinstall the Docker with Vaultvarden instead. (Bitwarden RS should be replaced with Vaultwarden)
3. Run the update script (Vaultwarden should be updated)

Please let us know if everything works as expected.

Thanks!

ke-lin:
> @ke-lin Can you please test this properly now once it's merged?
> 
>     1. Install Bitwarden RS, then run the update script. (Bitwarden RS should be updated)
> 
>     2. Run the new Vaultwarden script to reinstall the Docker with Vaultvarden instead. (Bitwarden RS should be replaced with Vaultwarden)
> 
>     3. Run the update script (Vaultwarden should be updated)
> 
> 
> Please let us know if everything works as expected.
> 
> Thanks!

@enoch85 sure! I will try to test it tonight, but I cannot promise, since I am quite busy at work right now.
If not tonight I will do it tomorrow night.

Issue #2018, opened 2021-06-10
Service systemd-sysctl fails with Linux kernel package for virtual machines


paulmenzel:
Using Ubuntu 20.04 with a Linux kernel tailored for a virtual machine, in my case [*linux-image-kvm*](https://packages.ubuntu.com/focal/linux-image-kvm) for QEMU/KVM, certain options are not selected, causing the service `systemd-sysctl` to fail.

```
ncadmin@nextcloud:~$ systemctl status systemd-sysctl.service
● systemd-sysctl.service - Apply Kernel Variables
     Loaded: loaded (/lib/systemd/system/systemd-sysctl.service; static; vendor>
     Active: failed (Result: exit-code) since Thu 2021-06-10 17:40:21 CEST; 4h >
       Docs: man:systemd-sysctl.service(8)
             man:sysctl.d(5)
    Process: 145 ExecStart=/lib/systemd/systemd-sysctl (code=exited, status=1/F>
   Main PID: 145 (code=exited, status=1/FAILURE)

Warning: some journal files were not opened due to insufficient permissions.
ncadmin@nextcloud:~$ sudo /lib/systemd/systemd-sysctl
Couldn't write '176' to 'kernel/sysrq', ignoring: No such file or directory
Not setting net/ipv4/conf/all/promote_secondaries (explicit setting exists).
Not setting net/ipv4/conf/default/promote_secondaries (explicit setting exists).
Couldn't write 'fq_codel' to 'net/core/default_qdisc', ignoring: No such file or directory
Couldn't write '4194304' to 'kernel/pid_max': Invalid argument
ncadmin@nextcloud:~$ ls /etc/sysctl.d/
10-console-messages.conf   10-magic-sysrq.conf       30-postgresql-shm.conf
10-ipv6-privacy.conf       10-network-security.conf  99-sysctl.conf
10-kernel-hardening.conf   10-ptrace.conf            README.sysctl
10-link-restrictions.conf  10-zeropage.conf
ncadmin@nextcloud:~$ grep -R 176 /etc/sysctl.*
/etc/sysctl.d/10-magic-sysrq.conf:kernel.sysrq = 176
ncadmin@nextcloud:~$ grep SYSRQ /boot/config-5.4.0-1009-kvm
# CONFIG_MAGIC_SYSRQ is not set
```

At least one of these seem to be generic Ubuntu package problems though.

[...]

PR #2047, opened 2021-07-15
Check for KVM and install kernel

enoch85:
cc @paulmenzel Your take on this?

### DIFF
--- a/nextcloud_install_production.sh
+++ b/nextcloud_install_production.sh
@@ -896,6 +896,14 @@ then
     fi
 fi
 
+# Fix https://github.com/nextcloud/vm/issues/2018
+if [[ "$SYSVENDOR" == "QEMU" || "$SYSVENDOR" == "Red Hat" ]];
+then
+    # KVM
+    apt-get install -y --install-recommends \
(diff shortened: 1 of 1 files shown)


nextcloud_install_production.sh
```diff
@@ -896,6 +896,14 @@ then
     fi
 fi
 
+# Fix https://github.com/nextcloud/vm/issues/2018
+if [[ "$SYSVENDOR" == "QEMU" || "$SYSVENDOR" == "Red Hat" ]];
+then
+    # KVM
+    apt-get install -y --install-recommends \
+    linux-image-kvm
```
paulmenzel, line 904: Do you know what recommends are that? I normally wouldn’t install them.

paulmenzel:
I do not think, that it fixes issue https://github.com/nextcloud/vm/issues/2018, as that is about problems caused by installing *linux-image-kvm*.

enoch85:
Hmm, but this adds kernel support for KVM right? 

I'm not on KVM myself, so you're the expert here.

enoch85:
No feedback --> close.

PR #2065, opened 2021-07-26
only update the updatenotification script if a new Nextcloud update is available

szaimen:
As the title says


szaimen:
This is ready 👍

PR #2067, opened 2021-07-29
Remove deSEC

enoch85:
Add a script for removal of deSEC accounts


enoch85:
@peterthomassen Any comments?

PR #2071, opened 2021-08-03
add hostname to notification

enoch85:


enoch85:
@szaimen I'm not sure about those SC errors.

Could you please have a look?

PR #2074, opened 2021-08-07
view hostname during notifications

enoch85:
Fix https://github.com/nextcloud/vm/pull/2071


enoch85:
Thanks for the suggestion @szaimen, much cleaner!

PR #2079, opened 2021-08-14
fix typo + update spamhaus script

### LINKED ISSUE #2078: /bin/sh: 1: /var/scripts/spamhaus_crontab.sh: not found
Hello,
I installed thursday the nextcloud-vm via script and now getting every night an error via e-mail
`/bin/sh: 1: /var/scripts/spamhaus_crontab.sh: not found` from the Cron Deamon.
I searched the `/var/scripts` folder and only found `spamhaus_cronjob.sh` instead of `spamhaus_crontab.sh`!
Is it save to change this in the crontab? Think this is an typo in the install-script!?
Can't exactely determine where it belongs but I think it is the add-on "enchanced security".

enoch85:
Fix #2078

### DIFF
--- a/addons/security.sh
+++ b/addons/security.sh
@@ -37,9 +37,11 @@ else
     rm -rf /var/log/apache2/evasive
     rm -f "$ENVASIVE"
     a2dismod reqtimeout
-    bash "$SCRIPTS"/spamhaus_cronjob.sh deletechain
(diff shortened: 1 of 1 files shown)


enoch85:
This PR relies on that https://github.com/wallyhall/spamhaus-drop/pull/3 is merged.

PR #2090, opened 2021-08-20
update geoblock database - august

szaimen:


enoch85:
Can't we script this in some way?

szaimen:
> Can't we script this in some way?

Probably yes but since this only takes not even 5min every 2months I thing it doesn't need to be automated...

enoch85:
Something like this would be nice: https://github.com/mitchellkrogza/nginx-ultimate-bad-bot-blocker

PR #2096, opened 2021-08-22
improve the notification on running the update script

szaimen:


enoch85:
Thing is, it must happen before any occ command is run whatsoever, so that's why it's in the top.

szaimen:
Fine by me 👍

PR #2103, opened 2021-09-09
only overwrite daemon.json if the overlay isn't set correctly

szaimen:


szaimen:
This is ready 👍

PR #2117, opened 2021-10-15
don't update Bitwarden for now

enoch85:


szaimen:
@enoch85 I see that you disabled updates for all docker containers not only bitwarden? Was this intentional?

enoch85:
No, but never mind. We'll update them again when Bitwarden is fixed.

PR #2130, opened 2021-11-11
Fix SC

enoch85:
@szaimen


szaimen:
Not sure where the failure suddenly comes from but should be fixed with 220bfc0

enoch85:
> Not sure where the failure suddenly comes from but should be fixed with 220bfc0

I've noticed that it's more thorough when the actual script is modified, that doesn't happen always with the lib, and the code behind the actual shellcheck might have been updated as well.

PR #2132, opened 2021-11-12
change IP host

enoch85:


lib.sh
```diff
@@ -38,22 +37,19 @@ KEYBOARD_LAYOUT=$(localectl status | grep "Layout" | awk '{print $3}')
 # HYPERVISOR=$(dmesg --notime | grep -i hypervisor | cut -d ':' -f2 | head -1 | tr -d ' ') TODO
 SYSVENDOR=$(cat /sys/devices/virtual/dmi/id/sys_vendor)
 # Network
-first_iface() {
```
szaimen, line 41: if you remove this function you might also remove it everywhere where it is used...

enoch85: This is the official VM, it's a bit different.

Need to double check this though.

szaimen: ah okay

szaimen: I didn't see that it is going to be merged into the official vm ;)

enoch85: In the official VM the lib isn't renewed every time, it stays local, and is released with every release.

So this should work.

PR #2140, opened 2021-11-18
simplify logic of apps re-enabling

### LINKED ISSUE #2113: Improve apps re-enabling
From this issue: https://github.com/nextcloud/vm/issues/2109

*Instead of modifying the mentioned logic, I'd vote for simplifying it instead after nextcloud/server#29020 is merged, backported and released...*

szaimen:
Close #2113

Disclaimer: I didn't test this.

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -890,7 +890,7 @@ then
     msg_box "Backup was not OK. Please check $BACKUP and see if the folders are backed up properly"
     exit 1
 else
-    printf "${IGreen}\nBackup OK!${Color_Off}\n"
+    print_text_in_color "$IGreen" "Backup OK!"
 fi
 
 # Download and validate Nextcloud package
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -812,22 +812,6 @@ then
     check_command lvrename /dev/ubuntu-vg/NcVM-snapshot /dev/ubuntu-vg/NcVM-snapshot-pending
 fi
 
-# Backup app status
-# Fixing https://github.com/nextcloud/server/issues/4538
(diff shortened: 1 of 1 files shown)


enoch85:
Aah, ok, so we'll let Nextcloud handle this, and only copy unsupported apps to the apps folder for easier re-anbling by the user?

Are you sure we don't need to restore apps-status and all that?

szaimen:
> Aah, ok, so we'll let Nextcloud handle this, and only copy unsupported apps to the apps folder for easier re-anbling by the user?
> 
> Are you sure we don't need to restore apps-status and all that?

Yes, thats excactly what https://github.com/nextcloud/server/pull/29020 has finally fixed ;)

enoch85:
Tested and working!

enoch85:
@szaimen Been having this in master now for a while and I have noticed that even core apps are missing/not enabled during update.

We now got several reports regarding this, so I really think we should improve this.

Something like:

1. Backup all current apps before the upgrade
2. Do the update
3. Check which apps that aren't enabled from the list in 1
4. Try to enabled all apps from the list in 3
5. If it doesn't succeed, ask to restore them (disabled)

That would guarantee that users easily can at least decide to enable the apps again through the GUI after a major upgrade were multiple apps are disabled. I don't trust Nextclouds app update/restore mechanism,.. :)

szaimen:
I am not sure about this. Shouldn't https://github.com/nextcloud/vm/blob/0a3f0c661924f26ecb428ffa6fb4be52cbe3d088/nextcloud_update.sh#L1076-L1096 already do as suggested?

enoch85:
You're right, sorry should have tested first.

PR #2151, opened 2021-11-30
also fix old installations from lool to cool

enoch85:
@small1


nextcloud_update.sh
```diff
@@ -591,6 +591,18 @@ $DOCKER_RUN_OUTPUT"
     docker_update_specific 'plex' "Plex Media Server"
 fi
 
+# Fix Collabora change too coolwsd
+if grep -r loolwsd $SITES_AVAILABLE/*
+then
```
enoch85, line 596: maybe add a loop here if there are more than one, shouldn't be, but you never know.

enoch85:
I've tested this and it works, but could you please confirm @szaimen?

PR #2158, opened 2021-12-04
add temporary fix in two places

enoch85:
This is so that we avoid to make a brand new release in case anything needs fixing in beginning or end


enoch85:
cc @small1

small1 (approved):

PR #2168, opened 2021-12-19
curl -sS; replace systemctl to service

vazhnov:
In https://gitlab.com/vazhnov/devuan_nextcloud_install_production.sh I've made an experiment with some changes, so now it is possible to use `nextcloud_install_production.sh` on non-systemd systems like Devuan. It would be great to push these changes into upstream, they look harmless. Here is only a part of changes.

Also, let's use `curl -sS` instead of `curl -s`.

I can split this to 2 separate merge requests, if you wish.


enoch85:
I'm not sure about this.

Sure it's great to be able to run the installer with other than Ubuntu, but the whole thing is assuming this is run on Ubuntu. Providing this possibility would mean that we would have to maintain it, and in the end that person would be me. :)

enoch85:
Btw, thanks for your efforts!

enoch85:
And, if anything - I've been thinking about switching to Debian.

vazhnov:
@enoch85 , non-systemd version can be separate (for example, in my repository). I want to keep the minimal difference with your original script.

vazhnov:
Switch to Debian — great idea! So, I can try to do changes for Devuan, and you can use them for Debian version. (In my opinion, Devuan is just a better version of Debian, because you can choose init system).

enoch85:
> Devuan

I'm sure it is, but we also have to think about our audience. We need a distro that is easy to maintain for a noob Linux user. Hence we chose Ubuntu from the get go. :)

enoch85:
@vazhnov We will probably not merge this, I just kept it here for thinking...

Thing is that `systemd` is the new way of doing things (like it or not), and we aim to keep the releases as close to default as possible for easier debugging. We don't want to reinvent the wheel.

With that said, I like the idea of using service instead to make it compatible with other Linux-distros but I'm not sure it's the right way forward.

I'm sorry, but I will close this. Thank you very much for your efforts.

PR #2174, opened 2021-12-27
fix janus conf

szaimen:
Before this fix, it would overwrite the whole line with the key. Now it works in my testing. I am using the same syntax like two lines below that.


enoch85:
Not tested, but ok. Seems legit.

Issue #2179, opened 2022-01-06
Logs empty


trhr:
### Steps To Reproduce

Install nextcloud

### Expected Result

`cat /var/log/nextcloud/nextcloud.log` should return results

### Actual Result

/var/log/nextcloud/nextcloud.log is empty, but /var/log/nextcloud.log.1 etc are being compressed.

### Additional Context

Solved by adding the **copytruncate** directive into /etc/logrotate.d/nextcloud.log.conf

### Build Version

23

### Environment

By using the scripts

### Environment Details

Ubuntu LXC

PR #2180, opened 2022-01-06
Fix for #2179

trhr:
See #2179


enoch85:
Thanks for this PR!

Please resolve DCO.

PR #2185, opened 2022-01-17
implement a notification for outdated Nextcloud versions

szaimen:

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -626,7 +626,7 @@ then
 fi
 
 # Update updatenotification.sh (gets updated after each nextcloud update as well; see down below the script)
-if [ -f $SCRIPTS/updatenotification.sh ] && ! grep -q "Check for supported Nextcloud 21" "$SCRIPTS/updatenotification.sh"
+if [ -f $SCRIPTS/updatenotification.sh ] && ! grep -q "Check for supported Nextcloud version" "$SCRIPTS/updatenotification.sh"
 then
     download_script STATIC updatenotification
     chmod +x $SCRIPTS/updatenotification.sh
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -625,6 +625,13 @@ then
     fi
 fi
 
+# Update updatenotification.sh (gets updated after each nextcloud update as well; see down below the script)
+if [ -f $SCRIPTS/updatenotification.sh ] && ! grep -q "Check for supported Nextcloud 21" "$SCRIPTS/updatenotification.sh"
+then
+    download_script STATIC updatenotification
+    chmod +x $SCRIPTS/updatenotification.sh
+fi
+
 # Update all Nextcloud apps
 if [ "${CURRENTVERSION%%.*}" -ge "15" ]
 then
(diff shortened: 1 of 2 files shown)


nextcloud_update.sh
```diff
@@ -625,6 +625,13 @@ then
     fi
 fi
 
+# Update updatenotification.sh (gets updated after each nextcloud update as well; see down below the script)
+if [ -f $SCRIPTS/updatenotification.sh ] && ! grep -q "Check for supported Nextcloud 21" "$SCRIPTS/updatenotification.sh"
```
enoch85, line 629: Shouldn't 21 be a $var?

szaimen: > Shouldn't 21 be a $var?

you  mean in lib.sh?

enoch85: "Check for supported Nextcloud 21"

With this we need to change for each version, use the function to check old versions instead and add that instead of hardcorded "21".

static/updatenotification.sh
```diff
@@ -25,6 +25,16 @@ NCMIN=$(curl -s -m 900 $NCREPO/ | sed --silent 's/.*href="nextcloud-\([^"]\+\).z
 REPORTEDMAJ="$CURRENTVERSION"
 REPORTEDMIN="$CURRENTVERSION"
 
+# Check for supported Nextcloud 21
+if [ "${CURRENTVERSION%%.*}" -lt 21 ]
+then
+    notify_admin_gui \
+        "Your Nextcloud version is End of life! Please upgrade as soon as possible!" \
```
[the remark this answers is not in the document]
szaimen: done

nextcloud_update.sh

enoch85: Nope, like this: https://github.com/nextcloud/vm/blob/master/lib.sh#L131

Already exists. ;)

szaimen: Hope that works now...

static/updatenotification.sh
```diff
@@ -25,6 +25,16 @@ NCMIN=$(curl -s -m 900 $NCREPO/ | sed --silent 's/.*href="nextcloud-\([^"]\+\).z
 REPORTEDMAJ="$CURRENTVERSION"
 REPORTEDMIN="$CURRENTVERSION"
 
+# Check for supported Nextcloud version
+if [ "${CURRENTVERSION%%.*}" -lt "$NCBAD" ]
```
szaimen, line 29: Hope that works now

enoch85: Looks better yes. :+1: 

But is this really needed? https://github.com/nextcloud/vm/blob/287c83682425654c40e92c0649aa8cc4be1a164a/static/updatenotification.sh#L24-L26

szaimen: > But is this really needed?
> 
> https://github.com/nextcloud/vm/blob/287c83682425654c40e92c0649aa8cc4be1a164a/static/updatenotification.sh#L24-L26

You mean if the linked code is needed?

enoch85: Exactly.

szaimen: It is indeed needed for getting the next minor update version so that we are able to send a notification if either a new minor or a major version is found. See the logic below.

szaimen: Reportedmaj and reportedmin store tje already reported version. Which is need so thqt you don't get an updatenotificarion each day but only one time for each version...

Issue #1794, opened 2021-02-01
Elasticsearch migration for FTS?


Ark74:
With the news of AWS forking Elastichsearch because of the licence change.
- https://www.elastic.co/pricing/faq/licensing
- https://aws.amazon.com/es/blogs/opensource/stepping-up-for-a-truly-open-source-elasticsearch/

For Fulltextsearch, if possible a transparent migration (7.10), should we move into it?
- https://opendistro.github.io

@daita 
@enoch85

enoch85:
For me it doesn't matter, as long as we follow licensing regulations.

If a migration is the best option, then go for it. :+1:

enoch85:
@Ark74 What do you think?

Ark74:
I'll try to research and do some testing this weekend to see if this is even possible, then we might be able to drop some ideas on it.

enoch85:
Thanks!

Ark74:
I've done some testing.
Opendistro for Elasticsearch bundles several plugins and provides a security functionality redundant to readonly rest.

It would require to re-structure the way it works so far, but on the good side of things the images will come ready from upstream. I'll have to study more the subject in order to connect it to the Nextcloud VM.

@enoch85 if you allow me, let this open and I'll update according the progress I make, and hopefully submit a PR.

enoch85:
Fine with me. Great job @Ark74!

enoch85:
Hi @Ark74 Any progress here?

Ark74:
Las time there where issues with the app.
I've seen there are updates, will try again.

enoch85:
Hi @Ark74 Still issues with the app?

enoch85:
Is this still valid?

Ark74:
Sorry @enoch85  this project got buried over so many other stuff, I'll send a MR if I can move it forward.
Closing for now.

enoch85:
ping @Ark74

PR #2197, opened 2022-01-27
Replace elasticsearch in favor of opensearch as fts engine

Ark74:
Close https://github.com/nextcloud/vm/issues/1794

To try this commit, please install a base VM without FTS.
Then manually get and run this PR [fulltextsearch.sh](https://raw.githubusercontent.com/Ark74/vm/add_opensearch_fts_engine/apps/fulltextsearch.sh) app

That should get the app and engine working.
Some values are hard coded in order to make the PR work.

Regards.


Ark74:
Closing as latest "fix" introduced a regression.
Will report soon.

Ark74:
Ready for review :+1:

apps/fulltextsearch.sh
```diff
@@ -1,18 +1,20 @@
...
+# SwITNet Ltd © - 2022, https://switnet.net/
 
 true
 SCRIPT_NAME="Full Text Search"
-SCRIPT_EXPLAINER="Full Text Search provides Elasticsearch for Nextcloud, which makes it possible to search for text inside files."
+SCRIPT_EXPLAINER="Full Text Search provides OpenSearch for Nextcloud, which makes it possible to search for text inside files."
 # shellcheck source=lib.sh
-source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+#source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
+source <(curl -sL https://raw.githubusercontent.com/Ark74/vm/add_opensearch_fts_engine/lib.sh)
```
szaimen, line 11: todo: needs to be reverted

apps/fulltextsearch.sh
```diff
@@ -87,51 +89,167 @@ fi
 
 # Check & install docker
 install_docker
+install_docker_compose
```
szaimen, line 92: Would it be possible to do it without requiring docker compose or is that not feasible anymore?

apps/fulltextsearch.sh
```diff
@@ -87,51 +89,167 @@ fi
...
---ulimit nofile=65536:65536 \
--p 127.0.0.1:9200:9200 \
--p 127.0.0.1:9300:9300 \
--v esdata:/usr/share/elasticsearch/data \
--v /opt/es/readonlyrest.yml:/usr/share/elasticsearch/config/readonlyrest.yml \
--e "discovery.type=single-node" \
--e "bootstrap.memory_lock=true" \
--e ES_JAVA_OPTS="-Xms1024M -Xmx1024M -Dlog4j2.formatMsgNoLookups=true" \
--i -t $nc_fts
+chmod 744 -R  $OPNSDIR
```
szaimen, line 224:
```suggestion
chmod 740 -R  $OPNSDIR
```

apps/fulltextsearch.sh
```diff
@@ -87,51 +89,167 @@ fi
...
+
+# Launch docker-compose
+cd $OPNSDIR
+docker-compose up -d
 
 # Wait for bootstrapping
-docker restart $fts_es_name
+docker restart $fts_node
```
szaimen, line 231: do we really need a restart here?

lib.sh
```diff
@@ -167,13 +167,20 @@ SPAMHAUS=/etc/spamhaus.wl
...
+    fts_node="fts_os-node"
+}
+create_certs(){
+    rm "$OPNSDIR"/"$1"
+    curl -s "$STATIC"/"$1" > "$OPNSDIR"/"$1"
+    sed -i "s|__NCDOMAIN__|$NCDOMAIN|" "$OPNSDIR"/"$1"
+    cd "$OPNSDIR"
+    bash "$1"
+}
+[ -n "$opensearch_install" ] && opensearch_install # TODO: remove this line someday
```
szaimen, line 183: I think we can get rid of this line finally

apps/fulltextsearch.sh
```diff
@@ -87,51 +89,167 @@ fi
...
+  fts_os-data:
+
+networks:
+  fts_os-net:
+
+#[1] https://github.com/opensearch-project/performance-analyzer
+YML_DOCKER_COMPOSE
+
+# Prepare certs
+#create_certs opensearch_certs.sh
```
szaimen, line 216: todo?

szaimen (commented):
Thanks for doing this! looks like a pretty immense change.
Just had a quick look but didn't test.

apps/fulltextsearch.sh
```diff
@@ -47,13 +49,13 @@ else
         fi
     done
     # Removal Docker image
-    docker_prune_this "$nc_fts"
```
szaimen, line 50: we should probably leave this here for cleaning up any potential old installation

apps/fulltextsearch.sh
```diff
@@ -87,51 +89,167 @@ fi
 
 # Check & install docker
 install_docker
+install_docker_compose
 set_max_count
-mkdir -p "$RORDIR"
-docker pull "$nc_fts"
-
-# Create configuration YML
-cat << YML_CREATE > /opt/es/readonlyrest.yml
```
szaimen, line 95: we should probably remove that file on old installations if they reinstall with this script?

apps/fulltextsearch.sh
szaimen: we should probably also remove the esdata volume in that case

apps/fulltextsearch.sh
```diff
@@ -47,13 +49,13 @@ else
         fi
     done
     # Removal Docker image
-    docker_prune_this "$nc_fts"
+    docker_prune_this "$opens_fts"
```
szaimen, line 52: what about also resetting the fts_os-data volume?

apps/fulltextsearch.sh
Ark74: Oh! i didn't realized I've removed that. I'll restore it.
IIRC docker system prune -a should remove all images and volumes, right?

szaimen: > system prune -a should remove all images

it removes all dangling images and networks, afaik. volumes need to be purge with docker volume prune but I am not sure if thats the best way to do it. deleting this volume specifically is probably better...

apps/fulltextsearch.sh
Ark74: yeah the content is below.

apps/fulltextsearch.sh
szaimen: what I meant was if it is not possible to simply use a docker run command for the container instead of relying on docker-compose?

apps/fulltextsearch.sh
Ark74: not really, i just replace the old variable.

szaimen: okay, would be cool to get rid of that then :)

apps/fulltextsearch.sh
szaimen: okay

apps/fulltextsearch.sh
szaimen: yes but as I said i think it is better to just remove the specific volume here...

apps/fulltextsearch.sh
Ark74: we are not running a vanilla image, there are some addons installed, that will require a third party to host and maintain that new image, with docker-compose there is no need to do that.

apps/fulltextsearch.sh
Ark74: That would require testing

apps/fulltextsearch.sh
szaimen: okay, that is a good reason. Thanks for making that clear!

apps/fulltextsearch.sh
Ark74: this is  a no go.

lib.sh
Ark74: done

apps/fulltextsearch.sh
Ark74: fixed by 61bd632

apps/fulltextsearch.sh
Ark74: solved by 74616ea

apps/fulltextsearch.sh
Ark74: solved by 61bd632f40c6bdd2a20cec8075104f0aa1e27b2e

lib.sh
```diff
@@ -167,13 +168,21 @@ SPAMHAUS=/etc/spamhaus.wl
...
+    OPNSREST=$(gen_passwd "$SHUF" "A-Za-z0-9")
     nc_fts="ark74/nc_fts"
-    fts_es_name="fts_esror"
+    opens_fts="opensearchproject/opensearch"
+    fts_node="fts_os-node"
+}
+create_certs(){
+    rm "$OPNSDIR"/opensearch_certs.sh
+    curl -s https://raw.githubusercontent.com/Ark74/vm/add_opensearch_fts_engine/static/opensearch_certs.sh > "$OPNSDIR"/opensearch_certs.sh
+    #curl -s "$STATIC"/opensearch_certs.sh > "$OPNSDIR"/opensearch_certs.sh
```
Ark74, line 181: TODO: replace when merged

apps/fulltextsearch.sh
```diff
@@ -48,12 +50,18 @@ else
     done
     # Removal Docker image
     docker_prune_this "$nc_fts"
+    docker_prune_this "$opens_fts"
+    docker_prune_this "esdata"
+    docker_prune_this "fts_os-data"
```
szaimen, line 55: is docker_prune_this really able to prune volumes?

apps/fulltextsearch.sh
```diff
@@ -27,7 +29,7 @@ root_check
 lowest_compatible_nc 18
 
 # Check if Full Text Search is already installed
-if ! does_this_docker_exist "$nc_fts" || ! is_app_installed fulltextsearch
+if ! does_this_docker_exist "$opens_fts" || ! is_app_installed fulltextsearch
```
szaimen, line 32:
```suggestion
if ! does_this_docker_exist "$nc_fts" || ! does_this_docker_exist "$opens_fts" || ! is_app_installed fulltextsearch
```

szaimen (commented):
Thanks, looks already much better!

apps/fulltextsearch.sh
```diff
@@ -27,7 +29,7 @@ root_check
 lowest_compatible_nc 18
 
 # Check if Full Text Search is already installed
-if ! does_this_docker_exist "$nc_fts" || ! is_app_installed fulltextsearch
+if ! does_this_docker_exist "$nc_fts" || ! does_this_docker_exist "$opens_fts" || ! is_app_installed fulltextsearch
```
szaimen, line 32:
```suggestion
if ! does_this_docker_exist "$nc_fts" && ! does_this_docker_exist "$opens_fts" && ! is_app_installed fulltextsearch
```

szaimen: Explanation: we only want to show the install_popup if neither of those containers/app was found. If it was, we want to show the option for reinstallation or uninstallation.

apps/fulltextsearch.sh
Ark74: actually it only stop them, then it prunes everything is not working.
https://github.com/nextcloud/vm/blob/74616ea74ea9ee3fa34ec7875e71d02f94b0805d/lib.sh#L1636-L1647

szaimen: ah okay, fine by me then

apps/fulltextsearch.sh
Ark74: yeah, ok.

Ark74:
Please check the latest changes and let me know if it works for you.
Regards

szaimen:
Thanks! lets wait for a review from @enoch85

enoch85:
Hey guys, I will have a look at this next week. I'm sick in Covid here (second time) and need to rest. A little dead right now after been working anyway the whole day.

szaimen:
> Hey guys, I will have a look at this next week. I'm sick in Covid here (second time) and need to rest. A little dead right now after been working anyway the whole day.

Rest and get well soon!

Ark74:
Hello, sorry to read that.
I was like that last week, same 2nd time.
This week I'm dealing only the small symptoms.
Get well soon.

szaimen:
- [ ] TODO: update the image during in the nextcloud_update script

enoch85:
OK @Ark74, you wanted me to implement something in lib?

apps/fulltextsearch.sh
```diff
@@ -27,7 +29,7 @@ root_check
 lowest_compatible_nc 18
 
 # Check if Full Text Search is already installed
-if ! does_this_docker_exist "$nc_fts" || ! is_app_installed fulltextsearch
+if ! does_this_docker_exist "$nc_fts" && ! does_this_docker_exist "$opens_fts" && ! is_app_installed fulltextsearch
```
enoch85, line 32: Sometimes the docker exists, but not the app. Sometimes the app exists, but not the docker. It should check for all ocurrances seperatly and ask to reinstall if any of them are true.

static/opensearch_certs.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
```
enoch85, line 1: Can you please move this to /apps instead?

lib.sh
```diff
@@ -167,13 +168,21 @@ SPAMHAUS=/etc/spamhaus.wl
...
-    ROREST=$(gen_passwd "$SHUF" "A-Za-z0-9")
+    OPNSREST=$(gen_passwd "$SHUF" "A-Za-z0-9")
     nc_fts="ark74/nc_fts"
-    fts_es_name="fts_esror"
+    opens_fts="opensearchproject/opensearch"
+    fts_node="fts_os-node"
+}
+create_certs(){
+    rm "$OPNSDIR"/opensearch_certs.sh
+    curl -s https://raw.githubusercontent.com/Ark74/vm/add_opensearch_fts_engine/static/opensearch_certs.sh > "$OPNSDIR"/opensearch_certs.sh
```
enoch85, line 180: change to: `run_script APPS opensearch_certs.sh`

lib.sh
enoch85: replace it now, we can fix bugs when merged. I trust that you tested it already.

enoch85 (changes requested):

Ark74:
@enoch85 it's already there, 
https://github.com/nextcloud/vm/blob/a7fa9a2a11bf8585f463a32c739f2032e8e12d6c/lib.sh#L178-185

just make sure that 
* https://github.com/nextcloud/vm/blob/a7fa9a2a11bf8585f463a32c739f2032e8e12d6c/lib.sh#L180 and
* https://github.com/nextcloud/vm/blob/a7fa9a2a11bf8585f463a32c739f2032e8e12d6c/lib.sh#L181

do the same thing

enoch85:
> do the same thing

Yes, but we want to use existing functions if possible.

apps/fulltextsearch.sh
szaimen: Yes, that is what it is doing now. If any of the rules apply, it will start the reinstallation. Only if all are false, it will show the install screen  :)

Ark74:
> Yes, but we want to use existing functions if possible.

well, create_certs is not done before.

enoch85:
> well, create_certs is not done before.

It will be when we merge it. :)

Just tested the current version, and it installs fine, but I couldn't get it to find text inside documents (PDF).

Ark74:
Interesting, maybe is just a db sync timing.

I'll setup both a new elastic and a opensearch later today to do some testing, it's amazingly late or amazingly early around here :sweat_smile:

lib.sh
```diff
@@ -177,11 +177,11 @@ opensearch_install() {
...
 create_certs(){
     rm "$OPNSDIR"/opensearch_certs.sh
-    curl -s https://raw.githubusercontent.com/Ark74/vm/add_opensearch_fts_engine/static/opensearch_certs.sh > "$OPNSDIR"/opensearch_certs.sh
-    #curl -s "$STATIC"/opensearch_certs.sh > "$OPNSDIR"/opensearch_certs.sh
-    sed -i "s|__NCDOMAIN__|$1|" "$OPNSDIR"/opensearch_certs.sh
-    cd "$OPNSDIR"
-    bash opensearch_certs.sh
+    curl -s https://raw.githubusercontent.com/Ark74/vm/add_opensearch_fts_engine/static/opensearch_certs.sh > "$SCRIPTS"/opensearch_certs.sh
+    #download_script APPS opensearch_certs
+    sed -i "s|__NCDOMAIN__|$1|" "$SCRIPTS"/opensearch_certs.sh
```
Ark74, line 182: Maybe 
mv "$SCRIPTS"/opensearch_certs.sh "$OPNSDIR"

enoch85: I've arleady fixed it in your fork

lib.sh
enoch85: Let's merge first, then I can have a look with find/replace

lib.sh
Ark74: ok, seems like this could break things.
I'll be back in some hours. o/

apps/fulltextsearch.sh
```diff
@@ -1,18 +1,19 @@
...
+SCRIPT_EXPLAINER="Full Text Search provides OpenSearch for Nextcloud, which makes it possible to search for text inside files."
 # shellcheck source=lib.sh
 source /var/scripts/fetch_lib.sh || source <(curl -sL https://raw.githubusercontent.com/nextcloud/vm/master/lib.sh)
 
 # Get all needed variables from the library
 ncdb
 nc_update
-es_install
+opensearch_install
+NCDOMAIN=$(nextcloud_occ_no_check config:system:get overwrite.cli.url | sed 's|https://||;s|/||')
```
enoch85, line 16: @szaimen aren't we using this in more places? :thinking: 

Maybe move to lib?

lib.sh
```diff
@@ -105,6 +106,8 @@ NCUSER=ncadmin
 UNIXUSER=$SUDO_USER
 UNIXUSER_PROFILE="/home/$UNIXUSER/.bash_profile"
 ROOT_PROFILE="/root/.bash_profile"
+# Name in trusted_config
+NCDOMAIN=$(nextcloud_occ_no_check config:system:get overwrite.cli.url | sed 's|https://||;s|/||')
```
szaimen, line 110: are you sure it is a good idea to get this value every time we get the library?

enoch85: We might use it in more places, and I think it was a good variable.

szaimen: maybe wrap it in a function then?

enoch85: Sure, go ahead!

enoch85 (approved):

lib.sh
enoch85: I'll merge this now, we can fix details later.

enoch85:
@szaimen If you want to fix the NCDOMAIN, please make a new PR, thanks! :pray:

Ark74:
@enoch85 
> Just tested the current version, and it installs fine, but I couldn't get it to find text inside documents (PDF).

Seems like even when launching `fulltextsearch:index` it waits for a cron job to sync the db, the only way to force realtime results is using,

`occ fulltextsearch:live`

The only difference is that it will require to manually exit the application, as it remains open, while `fulltextsearch:index`, closes automatically once the index is complete and awaits the cron job for the update.

I think that leaving the default behavior is ok.

enoch85:
Hmm, maybe something for a crontab, or service...

Ark74:
I think it uses the nextcloud php cron job

PR #2199, opened 2022-01-27
fix zero byte files

### LINKED ISSUE #2123: SetEnv proxy-sendcl 1
https://github.com/nextcloud/vm/issues/1534#issuecomment-954613649

szaimen:
Close #2123

@Zealot2000 did I understand correctly that you simply did this?

### DIFF
--- a/lets-encrypt/activate-tls.sh
+++ b/lets-encrypt/activate-tls.sh
@@ -209,6 +209,10 @@ then
 ### EXTRAS ###
     SSLUseStapling On
     SSLStaplingCache "shmcb:logs/ssl_stapling(32768)"
+
+    # Fix zero file sizes 
+    # See https://github.com/nextcloud/server/issues/3056#issuecomment-954209565
+    SetEnv proxy-sendcl 1
 TLS_CREATE
 fi
 
--- a/lets-encrypt/activate-tls.sh
+++ b/lets-encrypt/activate-tls.sh
@@ -121,6 +121,13 @@ then
     TLS13="+TLSv1.3"
 fi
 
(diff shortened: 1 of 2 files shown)


Zealot2000:
@szaimen Two actions are needed for this to work:
- Setting "SetEnv proxy-sendcl 1"
- Updating Apache to 2.4.47 or newer

(see https://bz.apache.org/bugzilla/show_bug.cgi?id=57087)

szaimen:
Thanks for the heads-up!

Since we are not going to update apache from other sources, I guess this will only be useful with ubuntu 22.04. So I guess @enoch85 should decide if we merge this now or wait for the switch to ubuntu 22.04...

enoch85:
@szaimen  Thanks!

Add a check for Ubuntu 22.04 so that we can merge now. 👌🏻

szaimen:
@enoch85 done 👍

enoch85:
@szaimen Thanks, but please use the `version` check instead.

szaimen:
> @szaimen Thanks, but please use the `version` check instead.

do you have an example how it should look like in this case?

enoch85:
@szaimen 

https://github.com/nextcloud/vm/blob/master/lets-encrypt/activate-tls.sh#L118-L122

enoch85:
@szaimen Can you please add the same fix in `nextcloud_install_production`?

szaimen:
@enoch85 do you know if it fixes the bug on Ubuntu 22.04? I was not able to confirm this yet...

enoch85:
No, I don't know. Haven't tested this specifically

PR #2206, opened 2022-01-29
add some docs how to increase the size of the root partition

szaimen:


Loomaanaatii:
Is it safe to run the command on an active nextcloud installation? The default 32GB / partition in the nextcloud VM template seems to be too small and keeps filling up. I'd like to expand it to 64gb.

For reference, this is the command I'd run:

`'sudo pvresize \$(sudo pvs | grep ubuntu-vg | grep -oP \"/dev/sda[0-9]\")'`

PR #2209, opened 2022-01-30
better check for LDAP

enoch85:
@szaimen

### DIFF
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -960,7 +960,11 @@ fi
 ## TODO; remove in NC 23.0.2
 if is_app_installed user_ldap
 then
-    nextcloud_occ app:disable user_ldap
+    if is_app_enabled user_ldap
+    then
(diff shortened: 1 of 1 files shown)


nextcloud_update.sh
```diff
@@ -960,7 +960,11 @@ fi
 ## TODO; remove in NC 23.0.2
 if is_app_installed user_ldap
 then
-    nextcloud_occ app:disable user_ldap
+    if is_app_enabled user_ldap
+    then
+        nextcloud_occ app:disable user_ldap
+        touch /tmp/ldap_enabled
```
szaimen, line 966: We could also store that in a variable

szaimen (approved):
Except my small comment it looks fine now 👍

nextcloud_update.sh
enoch85: Sure

PR #2213, opened 2022-01-31
Fix lib.sh create_certs function.

Ark74:
Make sure certs are in place.


enoch85:
I don't understand, so the script itself needs to be in the app folder?

enoch85:
It would be much better if the path actually was correct here, not assuming that it needs to be in `./`

Ark74:
The lines you marked are not really the issue, we could even remove them and that will not make any difference.

The path would have to be changed on every openssl command and I would prefer to avoid passing paths to openssl in this specific operation as it is a 3 step validation.

enoch85:
Not perfect, but OK. :)

PR #2218, opened 2022-02-02
update janus repo url

morph027:
Refs https://gitlab.com/packaging/janus/-/issues/5

Signed-off-by: morph027 <stefan.heitmueller@gmx.com>


morph027:
I've added redirects for the old URLs so it still works but this PR updates to the new URL scheme.

szaimen (approved):

PR #2242, opened 2022-02-19
Register domain on existing accounts correctly

enoch85:
This fixes the last issue with deSEC.

cc @peterthomassen @antieon @nils-wisiol


enoch85:
@antieon Could you please confirm it's working?

PR #2251, opened 2022-03-03
Fix syntax error generation in pihole-install.sh

pq2:
This line created an empty if block in pihole-install.sh, which is syntactically not allowed


szaimen:
Thanks! I'll have a look soon!

szaimen (approved):
tested successfully so good to go from my side 👍

PR #2271, opened 2022-04-07
don't unmount the daily backup drive if only the backup check is running

szaimen:
This is best reviewed like this: https://github.com/nextcloud/vm/pull/2271/files?diff=unified&w=1


szaimen:
This is ready :)

Issue #2261, opened 2022-03-16
update.ssh script


myouha:
### Steps To Reproduce

I have run the script using the following argument:
> sudo ssh /var/script/update.ssh 22.2.5

as I had the 22.1.0.1 and did not want to make the jump to the 23 version.

I restarted the the server and did not take a screen shot of the error, it was not meaningful for me.
Now when turning off the maintenance mode, nextcloud says:  "Please use the command line updater because automatic updating is disabled in the config.php."

tried to run the command line updater.phar but got errors, and it says that the "Current version is 23.0.2."

Then the occ upgrade:
Exception: Updates between multiple major versions and downgrades are unsupported.
Update failed

### Expected Result

Upgrade to 22.2.5 success

### Actual Result

Upgrade ERROR

### Build Version

22.1.0

### Environment

By using the scripts

### Environment Details

Ubuntu 20.04 LTS

szaimen:
I guess this should be fixable by first running 
`sudo bash /var/scripts/update.sh minor`
and then
`sudo bash /var/scripts/update.sh`

enoch85:
Just tested here and it works. --> closing.

Please see docs [here](https://docs.hanssonit.se/s/W6fMouPiqQz3_Mog/virtual-machines-vm/d/W7Du9uPiqQz3_Mr1/nextcloud-vm-machine-configuration).

tuxick:
> Just tested here and it works. --> closing.

I got similar error, required updating version number in config.php before occ upgrade would work

myouha:
> 

Thanks a lot, this trick worked.

myouha:
> > Just tested here and it works. --> closing.
> 
> I got similar error, required updating version number in config.php before occ upgrade would work

This trick worked for me too. Thanks for taking the time to reply.

PR #2277, opened 2022-04-17
always update minor before major version

enoch85:
This improves stability of the update, and covers the cases when Nextcloud makes a X.**X**.XX release.


enoch85:
cc @szaimen What do you think?

szaimen:
Sry for my late answer.

In general this should not be needed but I get where the idea comes from

enoch85:
When NC released 22.X.1 then this is needed to since the script can't handle point releases which contains a "major feature" release.

enoch85:
> In general this should not be needed but I get where the idea comes from

https://github.com/nextcloud/vm/issues/2261#issuecomment-1069080250

szaimen:
> > In general this should not be needed but I get where the idea comes from
> 
> [#2261 (comment)](https://github.com/nextcloud/vm/issues/2261#issuecomment-1069080250)

alright 👍

PR #2281, opened 2022-04-25
Fix Desktop client bug

enoch85:
Tested and works!


enoch85:
cc @szaimen FYI

Tested and works on my own server.

Issue #2224, opened 2022-02-06
increase security of turnserver?


szaimen:
I've recently head a look at the talk docs and I've found this section:
https://github.com/nextcloud/spreed/pull/5788/commits/71a82ec05feb8d838869a72d657e5d1435155ae9

What do you think about the security impact? Should we implement those, too?

cc @morph027 on this :)

morph027:
Good idea. Not sure how the setup is within the VM now, if it's connected to more than one subnet etc. ... Might depend on the user setup.

szaimen:
If I understand the docs correctly, we should be able to enable this if the hpb gets installed. Otherwise probably not...

enoch85:
I have no opinion. Do what you feel is best.

enoch85:
@szaimen @morph027 is this something that you will work on anytime soon you think? :)

szaimen:
I rather not think so... I am also not sure if there is really a way to limit this to a defined set of ip-addresses because I don't know in detail how the whole stack works together...

morph027:
Sorry, needed to catch up on my Github stream ;)

After reading through, we actually should only allow the HBP address, which - if i'm right - is the ip address of the machine the whole stack runs on. So this depends on the fact if HBP is enabled ans installed, but within the coturn script section. I'll have a closer look.

szaimen:
So if the HPB is installed, coturn will always only connect directly to the HPB and to nothing else?

morph027:
Looks like. In fact, the TURN server should never connect to anything else when not running with HBP, as it's purpose is to handle inbound connections from the outside to the outside. I will test-drive the settings on my personal instance.

szaimen:
okay, I was not sure hence my question. Thanks for investigating this! :)

morph027:
Tested some video calls and everything seems to work okay.

Only thing i've added is:

```
allowed-peer-ip=127.0.0.1
denied-peer-ip=0.0.0.0-0.255.255.255
```

to turnserver config.

enoch85:
Would you mind making a PR? :)

enoch85:
This is done.

PR #2293, opened 2022-05-06
refs #2224 increase security of turnserver

morph027:
Signed-off-by: morph027 <stefan.heitmueller@gmx.com>


morph027:
I've tested this successfully with 2 users (one mobile via LTE, one notebook via WIFI) on my own HBP/CoTURN instance.

enoch85:
Thanks @morph027! <3

PR #2300, opened 2022-05-08
Adjust openssl serial file use on 3.0

Ark74:
openssl 3.0.x (jammy's version) does not create serial file anymore, so only moving on 20.04 (openssl 1.1.1).

### DIFF
--- a/apps/opensearch_certs.sh
+++ b/apps/opensearch_certs.sh
@@ -40,6 +40,6 @@ mv client.pem \
    root-ca-key.pem -t tls_store
 
 # openssl 3.0 workaround
-if [ $(lsb_release -sr) = 20.04 ]; then
+if [ "$(lsb_release -sr)" = "20.04" ]; then
    mv root-ca.srl tls_store
 fi
--- a/apps/opensearch_certs.sh
+++ b/apps/opensearch_certs.sh
@@ -37,5 +37,9 @@ rm admin-key-temp.pem \
 # Store
 mv client.pem \
    client-key.pem \
-   root-ca.srl \
    root-ca-key.pem -t tls_store
+
+# openssl 3.0 workaround
+if [ $(lsb_release -sr) = 20.04 ]; then
+   mv root-ca.srl tls_store
+fi


Ark74:
@enoch85 not sure if adding the whole lib.sh is worth only for using `$DISTRO` using lsb_release for now.

Ark74:
Tested and works, cheers!

apps/opensearch_certs.sh
```diff
@@ -37,5 +37,9 @@ rm admin-key-temp.pem \
 # Store
 mv client.pem \
    client-key.pem \
-   root-ca.srl \
    root-ca-key.pem -t tls_store
+
+# openssl 3.0 workaround
+if [ $(lsb_release -sr) = 20.04 ]; then
```
enoch85, line 43: @Ark74 Please follow SC recomendations. 

Also, this is on Ubuntu 22.04, so we should really find a valid fix for that instead of doing a workaround like this. :)

Ark74: SC?

Does that mean that 20.04 is deprecated?

enoch85 (changes requested):
See comment in-code

Ark74:
Well, if both (20.04/22.04) are still supported then there is no legacy stuff here, as openssl 1.1.1 will be supported for the next 3 years and that's how it works,  openssl 3.0 most likely will never reach 20.04.

Another thing will be to leave serial file there, moving to the "store" folder has no functional but organization/esthetic purpose.

Oh, sc I see, sure, no problem there.

enoch85:
OK, but from what I can see, this won't work on 22.04, right? That defeats the whole purpose of this script, since all the new VMs will run 22.04 within a few days/weeks.

Ark74:
I get I confuse you, this MR fix it for 22.04 (the only OS having issues) and 20.04 will keep working as it should.
Meaning this change is intended for both to be working.

The current issue only happens on 22.04, 20.04 is still working fine (even without this MR).

Ark74:
Updating name, as it might confuse the goal of MR.

enoch85:
Yes, I get that. But the fix shouldn't be working around that it doesn't work with the current code in 22.04. The fix should be to fix it on both 20.04 and 22.04. :)

Ark74:
openssl 3 (jammy) does not create a serial file anymore, so no need to move anything, the serial file moving is only required on openssl 1.1.1 (focal).

Issue #2316, opened 2022-05-13
iNotify not available when installing SMB


enoch85:
cc @szaimen 

```
Build process completed successfully
Installing '/usr/lib/php/20190902/inotify.so'
install ok: channel://pecl.php.net/inotify-3.0.0
configuration option "php_ini" is not set to php.ini location
You should add "extension=inotify.so" to php.ini
touch: cannot touch '/etc/php/8.1/mods-available/inotify.ini': No such file or directory
grep: /etc/php/8.1/mods-available/inotify.ini: No such file or directory
/var/scripts/smbmount.sh: line 356: /etc/php/8.1/mods-available/inotify.ini: No such file or directory
/var/scripts/smbmount.sh: line 357: /etc/php/8.1/mods-available/inotify.ini: No such file or directory
WARNING: Module inotify ini file doesn't exist under /etc/php/7.4/mods-available
WARNING: Module inotify ini file doesn't exist under /etc/php/7.4/mods-available
Setting the max folder variable to 524288...
fs.inotify.max_user_watches=524288
vm.overcommit_memory = 1
fs.inotify.max_user_watches = 524288
Installing the files_inotify app...
Error: App "INotify file watcher" cannot be installed because the following dependencies are not fulfilled: The library inotify is not available.
files_inotify disabled
files_inotify 0 removed
```

szaimen:
Fix in https://github.com/nextcloud/vm/pull/2317

PR #2317, opened 2022-05-13
fix inotify

szaimen:
Fix https://github.com/nextcloud/vm/issues/2316

Disclaimer: I did not test this but it should work


enoch85:
Not tested but makes sense.

PR #2342, opened 2022-05-25
chore: Included githubactions in the dependabot config

nathannaveen:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+      interval: "weekly"


enoch85:
Thanks for this PR!

I see no real gain for it in this repo since we don't work with specific versions in the same way as a docker container for example.

@szaimen what do you say?

szaimen:
It is for updating the github actions so it makes sense imho

PR #2380, opened 2022-08-19
Add menu item for Google Domains

93-infinity:
1. Add menu item for Google Domains in alphabetical position
2. Add configuration section for Google Domains.
3. Instructions: copied section from "strato" section because it mentions getting "ddns for your domain", not "ddns account" like the other account types
4. Guide: use Google's own DDNS instructions page
5. Protocol: use "dyndns2" not "googledomains" even though it is is supported natively in ddclient 3.9+. Reason: ddns updates always succeed if protocol set to "googledomains" even when using junk data. Google DDNS support page says both update methods using ddclient are supported.


enoch85:
Thanks for your contribution, is this tested and works?

93-infinity:
Yes, I tested it to be working to successfully to update my google domain DDNS

enoch85:
Weclome as a contributor to Nextcloud!

PR #2391, opened 2022-08-30
move the notify-crontab to its own script

### LINKED ISSUE #2356: Make the `files_inotify` a script instead
Seems like the `reboot` option isn't working in the crontab created to run `files_inotify` job.

Running it manually works, but it doesn't seem to work when a reboot is made. So I think it would be better to run it in a script on a time schedule instead. For example once per day, or every 10 minutes.

What needs to be done:

1. `pgrep` to check if the job is already running. If not, run it in a `nohup`
2. Update all the current cronjobs, to call the script instead of the cronjob itself with the update script)

@szaimen Since it's your contribution, do you think you could have a look at it?

Thanks! :bowing_man:

szaimen:
Close #2356


enoch85:
@szaimen Thanks! Could you please fix the failing tests, and I'll finish it off.

szaimen:
> @szaimen Thanks! Could you please fix the failing tests, and I'll finish it off.

done

enoch85:
Thanks! Will have a look tomorrow! :1st_place_medal:

apps/smbmount.sh
```diff
@@ -408,9 +408,14 @@ files_inotify app and set up the cronjob for this external storage."
...
-            # Add crontab for this external storage
+            # Download script
+            download_script ADDONS notify-crontab
+            chmod 700 "$SCRIPTS"/notify-crontab.sh
+            chown root:root "$SCRIPTS"/notify-crontab.sh
+
+            # Add crontab
             print_text_in_color "$ICyan" "Generating crontab..."
-            crontab -u www-data -l | { cat; echo "@reboot sleep 20 && php -f $NCPATH/occ files_external:notify -v $MOUNT_ID >> $VMLOGS/files_inotify.log"; } | crontab -u www-data -
+            crontab -u root -l | { cat; echo "@reboot $SCRIPTS/notify-crontab.sh $MOUNT_ID"; } | crontab -u root
```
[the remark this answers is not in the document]
szaimen: does sourcing the lib works as different user?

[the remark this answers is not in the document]
enoch85: Aah, that little detail.
 Valid point!

enoch85:
Not tested, but looks good to me.

enoch85:
OK, so coding wise almost done, but the issue that remains is how to fetch the correct `MOUNT_ID`. I was thinking somehting like `nextcloud_occ app:files_external:list` but in NC 25 I get `Command "files_external" is not defined.`.

Any idea @szaimen?

enoch85:
OK, so for an idea at least:

`"$(nextcloud_occ files_external:list | tail 1)"` but since I don't know how the output looks, it's hard to guess the final command.

szaimen:
> OK, so coding wise almost done, but the issue that remains is how to fetch the correct `MOUNT_ID`. I was thinking somehting like `nextcloud_occ files_external:list ` but in NC 25 I get `Command "files_external" is not defined.`.
> 
> Any idea @szaimen?

Did you enable the files_external app?

enoch85:
> Did you enable the files_external app?

Haha yes. :D

enoch85:
@szaimen Can you check the output of the above command and form a variable that will be the actual `MOUNT_ID`? I don't have any SMB mount installed.

Thanks!

szaimen:
You can use a local external storage mount as example. That should have the same output...

enoch85:
OK, grepped it from DB.

Can you please test if this works: https://github.com/nextcloud/vm/pull/2391/files#diff-67bceb0888b398c45857e1bc41f753da1a696075ff08b9b82f8136bc6af0139aR1124-R1134

Thanks!

nextcloud_update.sh
```diff
@@ -1121,6 +1121,18 @@ then
     print_text_in_color "$ICyan" "Nextcloud crontab updated to run every 5 minutes."
 fi
 
+# Replace iNotify checker with a script instead
+if crontab -u www-data -l | grep -q "files_external:notify -v"
```
szaimen, line 1125: Why not get the ids from the crontab? That would be a much safer way imho...

enoch85: The crontab might not be true (not used, leftover, etc), but the DB is. 

Also, will several mounts generate several crontab-rows?

szaimen: > The crontab might not be true (not used, leftover, etc), but the DB is.

However the user might not have axtivated the feature for all mounts and does not want to do so.

> Also, will several mounts generate several crontab-rows?

Yes. Each mount for which the feature was activated will have its own row.

enoch85: Hmm... It makes it a bit more complicated. 

Maybe the best option here is to just ditch the replacement to the new way of doing it.

PR #2394, opened 2022-09-09
increase turnserver quota

szaimen:
Seems like this increases available connections behind NAT. See [this report](https://github.com/nextcloud/all-in-one/issues/1121#issuecomment-1241802920).


enoch85:
Thanks!

morph027:
hm, thought 0 would be infinite, but i'm fine as long as it's working ;)

PR #2399, opened 2022-10-13
fix onlyoffice by adding secret

szaimen:


enoch85:
Thanks @szaimen! Was puling my hair on this one.

PR #2403, opened 2022-10-21
delete jwt stuff during reinstall

enoch85:
@szaimen seems like the secrets aren't updated. @szaimen not tested, please test this

Signed-off-by: Daniel Hansson <mailto@danielhansson.nu>

### DIFF
--- a/apps/onlyoffice_docker.sh
+++ b/apps/onlyoffice_docker.sh
@@ -30,6 +30,8 @@ else
     # Remove config.php value set when install was successful
     nextcloud_occ config:system:delete allow_local_remote_servers
     nextcloud_occ config:system:delete onlyoffice
+    nextcloud_occ config:system:delete onlyoffice jwt_secret
+    nextcloud_occ config:system:delete onlyoffice jwt_header
     # Show successful uninstall if applicable
     removal_popup "$SCRIPT_NAME"
 fi


apps/onlyoffice_docker.sh
```diff
@@ -30,6 +30,8 @@ else
     # Remove config.php value set when install was successful
     nextcloud_occ config:system:delete allow_local_remote_servers
     nextcloud_occ config:system:delete onlyoffice
```
szaimen, line 32: @enoch85 this line should already delete the complete array so deleting each value manually should not be necessary...

enoch85: Hmm, strange because it didn't update for a customer I just made a reinstall with. Even if set in the end of the script. I had to run the whole thing with `-x` to see the secret and then paste it in the GUI.

Maybe we should `print_text_in_color` for the secret itself just so that it can be handy to copy from the output?

szaimen: Even if not removed, it should get overwritten with the new value here:
https://github.com/nextcloud/vm/blob/703fe1a3bbdc57c0f876c6767acecf01edf47304/apps/onlyoffice_docker.sh#L261-L262

So not sure what got wrong on your customers reinstall...

enoch85: >  Even if set in the end of the script

> it should get overwritten with the new value here:

Me neither. Let's leave it as is.

PR #2422, opened 2022-12-28
make apps availble on deSEC

### LINKED ISSUE #1980: Implement deSEC in Collabora | OnlyOffice docker versions
It would remove the need for adding a domain and all that extra stuff that you currently need to do to get it up and running.

Maybe make it an option when TLS is generated, own domain (CNAME) or deSEC?

enoch85:
Signed-off-by: Daniel Hansson <mailto@danielhansson.nu>

Fix https://github.com/nextcloud/vm/issues/1980

- [x] Double check that TLS certs are generated for the subdomain. Right now I get this in the logs: `cURL error 60: SSL: no alternative certificate subject name matches target host name 'onlyoffice.testingdomainenoch85.dedyn.io' (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://onlyoffice.testingdomainenoch85.dedyn.io/ConvertService.ashx`

### Nice to have 
- [ ] If a specific port is set, also set this port in the `docker run` command.


addons/deSEC/onlyoffice_docker_desec.sh
```diff
@@ -17,7 +17,7 @@ debug_mode
 # Check if root
```
szaimen, line 17: Why did you move the script? Shouldnt be there still a way to runn it without desec?

szaimen (changes requested):

enoch85:
This is still WIP.

szaimen:
> This is still WIP.

Okay

addons/deSEC/onlyoffice_docker_desec.sh
enoch85: It's not ready yet, but thanks for pointing that out. Actually missed that one.

enoch85:
@szaimen Feel free to begin initial tests if you want. :+1:

menu/additional_apps.sh
```diff
@@ -125,7 +125,7 @@ to finish the setup once this script is done." "$SUBTITLE"
     ;;&
     *"Talk"*)
         print_text_in_color "$ICyan" "Downloading the Talk script..."
-        run_script APP talk
+        desec_app_tls_menu "DESEC talk_desec" "APP talk"
```
szaimen, line 128: honestly, I find it a bit strange that this will now always go through this menu. It will be yet another step

lib.sh
```diff
@@ -850,6 +850,27 @@ to validate them with the $f method. We have exhausted all the methods. Please c
...
 }
 
+# Let the user choose to setup a specific app with either deSEC, or regular TLS.
+# desec_app_tls_menu "DESEC collabora_docker_desec" "APP collabora_docker"
+desec_app_tls_menu() {
+choice=$(whiptail --title "$TITLE" --menu \
+"Choose TLS setup. Please note, to run the deSEC option, deSEC needs to be configured and setup already.\n
+$MENU_GUIDE\n\n$RUN_LATER_GUIDE" "$WT_HEIGHT" "$WT_WIDTH" 4 \
+"deSEC TLS setup" "(If you configured deSEC already. Works with custom port.)" \
+"Regular TLS setup" "(If deSEC isn't installed, setup normal TLS)" 3>&1 1>&2 2>&3)
```
szaimen, line 860: Can we make Regular TLS setup the default if you want to keep this pre-menu?

menu/additional_apps.sh
enoch85: Do you have a better proposal?

lib.sh
enoch85: Sure

enoch85:
Talk works out of the box. 

Onlyoffice gets 504 timeout, but everything is generated as it should.

Collabora is OK when hitting "Save" in the UI, but no document is loaded when opening it.

menu/additional_apps.sh
enoch85: ping @szaimen

szaimen: Can we maybe check for the dessec folder and only show that menu if dessec is already configured?

enoch85: Hmm, should work.

enoch85: Thought about this, what about users that doesn't have deSEC installed and wants to use it?

PR #2426, opened 2023-01-16
Update system-restore.sh and backup-viewer.sh

pq2:
Compare https://github.com/nextcloud/vm/blob/master/not-supported/daily-backup-wizard.sh#L437

Fixed issue with inconsistent quotes related to ENCRYPTION_KEY
sed did not replace the normal ticks, which caused it to be considered part of the password, which is why during restoration it reported that the password was incorrect.

Signed-off-by: Florian Helmschmidt <github@nhelmschmidt.de>

### DIFF
--- a/not-supported/system-restore.sh
+++ b/not-supported/system-restore.sh
@@ -64,7 +64,7 @@ Please do that before you can view backups."
     exit 1
 fi
 # Get needed variables
-ENCRYPTION_KEY="$(grep "ENCRYPTION_KEY=" "$DAILY_BACKUP_FILE" | sed 's|.*ENCRYPTION_KEY="||;s|"||')"
+ENCRYPTION_KEY="$(grep "ENCRYPTION_KEY=" "$DAILY_BACKUP_FILE" | sed 's|.*ENCRYPTION_KEY=\x27||;s|\x27||')"
 DAILY_BACKUP_MOUNTPOINT="$(grep "BACKUP_MOUNTPOINT=" "$DAILY_BACKUP_FILE" | sed 's|.*BACKUP_MOUNTPOINT="||;s|"||')"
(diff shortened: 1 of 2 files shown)


not-supported/system-restore.sh
```diff
@@ -64,7 +64,7 @@ Please do that before you can view backups."
     exit 1
 fi
 # Get needed variables
-ENCRYPTION_KEY="$(grep "ENCRYPTION_KEY=" "$DAILY_BACKUP_FILE" | sed 's|.*ENCRYPTION_KEY="||;s|"||')"
+ENCRYPTION_KEY="$(grep "ENCRYPTION_KEY=" "$DAILY_BACKUP_FILE" | sed 's|.*ENCRYPTION_KEY=\x27||;s|\x27||')"
```
szaimen, line 67:
```suggestion
ENCRYPTION_KEY="$(grep "ENCRYPTION_KEY=" "$DAILY_BACKUP_FILE" | sed 's|.*ENCRYPTION_KEY=||;s|\'||g;s|"||g')"
```

szaimen (commented):
Thanks for the PR! Just a small suggestion

not-supported/system-restore.sh
pq2: This pull request should not be considered as "this is how the problem should be solved" PR. I know far too little about sed to say what is the best solution here and it would take me far too much time to figure it out. I just tried my solution and it worked. Do whatever you think is best. I probably should have just reported this as an issue instead of a PR.

szaimen: all right. just tested and my proposal seems to work

szaimen:
```
export TEST='ENCRYPTION_KEY="dfjio'dkfjl'"'
echo "$(echo "$TEST" | sed "s|.*ENCRYPTION_KEY=||;s|'||g;s|\"||g")"
dfjiodkfjl
```

enoch85:
@pq2 Thanks! 👍🏼

@szaimen Do you approve?

szaimen (approved):
Ah yes, forgot that I can :D

PR #2461, opened 2023-03-31
Change Updatenotification

enoch85:
Follow up on: https://github.com/nextcloud/vm/commit/7c7b5f59473da6fd37ec7576abbbb7f22010a37d

@szaimen Any thoughts on this? I ask becasue you implemented it.

My idea here is to run updates less frequently. Once per month is enough.


enoch85:
> Hm... Is there a specific reason why the update script should only run once a month and not every week? This means some instances might be vulnerable in worst case for a whole month without any updates installed...

Yes, it's about uptime. I'm aware of the downside, but if you want to update more frequently then you can just change the crontab. Once per month is enough from a security standpoint if you ask me. In worst case, some CVE might not be patched instantly (aka next week) but 3 weeks more is not a problem I'd say - maybe if you're a government, but not a "regular user".

nextcloud_install_production.sh
```diff
@@ -564,7 +564,7 @@ nextcloud_occ config:app:set updatenotification notify_groups --value="[]"
 print_text_in_color "$ICyan" "Configuring update notifications specific for this server..."
 download_script STATIC updatenotification
 check_command chmod +x "$SCRIPTS"/updatenotification.sh
-crontab -u root -l | { cat; echo "59 $AUT_UPDATES_TIME * * * $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
+crontab -u root -l | { cat; echo "59 $AUT_UPDATES_TIME * 1-12 6 $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
```
szaimen, line 567: I would adjust this. The logic is there to only send the notification once even though the acript runs daily.

nextcloud_update.sh
```diff
@@ -794,6 +794,8 @@ if [ -f "$SCRIPTS"/updatenotification.sh ]
 then
     download_script STATIC updatenotification
     chmod +x "$SCRIPTS"/updatenotification.sh
+    crontab -u root -l | grep -v "$SCRIPTS/updatenotification.sh" | crontab -u root -
```
szaimen, line 797: I wouldn adjust this

nextcloud_update.sh
```diff
@@ -794,6 +794,8 @@ if [ -f "$SCRIPTS"/updatenotification.sh ]
...
+    crontab -u root -l | { cat; echo "59 $AUT_UPDATES_TIME * * 5 $SCRIPTS/updatenotification.sh > /dev/null 2>&1"; } | crontab -u root -
```
szaimen, line 798: Same

static/updatenotification.sh
```diff
@@ -56,7 +56,7 @@ then
         notify_admin_gui \
         "New minor Nextcloud Update!" \
         "Nextcloud $NCMIN just became available. Since you are running Automatic \
-Updates on Saturdays at $AUT_UPDATES_TIME:00, you don't need to bother about updating \
+Updates every month on Saturdays at $AUT_UPDATES_TIME:00, you don't need to bother about updating \
```
szaimen, line 59: Looks good

szaimen (changes requested):

szaimen (commented):
So if you really want to change it to monthly adjusting only the wording shoupd be enough

PR #2464, opened 2023-04-01
Add Imaginary Docker for previews

enoch85:
@szaimen This is ready for testing. 

Some points I'm not sure about:
1. Are `previewgenerator` even needed after the rebuild?
2. Does it need a cronjob?
3. Which PreviewsProviders are supported? ( added the old ones, and looked at the h2non README + Nextcloud docs)
4. Do I need to specify0 `jpeg_qulity` and such?
5. Can I choose where all the previews ends up with a `--flag`? Optimal would be the data disk and not `appdata`
6. Does this replace `previewgenerator`  once rebuilt and installed?

Thanks!


enoch85:
TODO:
1. Only rebuild if Preview generator was previously installed
2. Don't Install Preview generator "just in case", just set it as active for the rebuild

szaimen:
> 1. Are `previewgenerator` even needed after the rebuild?

No

> 2. Does it need a cronjob?

No

> 3. Which PreviewsProviders are supported? ( added the old ones, and looked at the h2non README + Nextcloud docs)

All these https://github.com/nextcloud/server/blob/701cc1eabac96cff5ccbb342b2f036d58995b2d6/lib/private/Preview/Imaginary.php#L60 but there is no way to customize them.

> 4. Do I need to specify0 `jpeg_qulity` and such?

Yes, jpeg_quality will still be used: https://github.com/nextcloud/server/blob/701cc1eabac96cff5ccbb342b2f036d58995b2d6/lib/private/Preview/Imaginary.php#L124

> 5. Can I choose where all the previews ends up with a `--flag`? Optimal would be the data disk and not `appdata`

No, they will still end up in appdata->previews

> 6. Does this replace `previewgenerator`  once rebuilt and installed?

It works completely different but in general yes.

addons/imaginary.sh
```diff
@@ -0,0 +1,144 @@
...
+    fi
+    # Show successful uninstall if applicable
+    removal_popup "$SCRIPT_NAME"
+fi
+
+# Generate docker-compose.yml
+mkdir -p "$SCRIPTS"/imaginary-docker
+if [ ! -f "$SCRIPTS/imaginary-docker/docker-compose.yml" ]
+then
+    touch "$SCRIPTS/imaginary-docker/docker-compose.yml"
```
szaimen, line 57: Why use docker-compose for this? I wouldnt say it is necessary here

addons/imaginary.sh
```diff
@@ -0,0 +1,144 @@
...
+else
+    exit
+fi
+
+# Set providers
+nextcloud_occ config:system:set enabledPreviewProviders 0 --value="OC\\Preview\\JPEG"
+nextcloud_occ config:system:set enabledPreviewProviders 1 --value="OC\\Preview\\PNG"
+nextcloud_occ config:system:set enabledPreviewProviders 2 --value="OC\\Preview\\WEBP"
+nextcloud_occ config:system:set enabledPreviewProviders 3 --value="OC\\Preview\\HEIF"
+nextcloud_occ config:system:set enabledPreviewProviders 4 --value="OC\\Preview\\TIFF"
```
szaimen, line 107: You dont need to add this as it is enabled automatically by enabling the imaginary preciew provider. Same with others.

addons/imaginary.sh
```diff
@@ -0,0 +1,144 @@
...
+nextcloud_occ config:system:set enabledPreviewProviders 10 --value="OC\\Preview\\MarkDown"
+nextcloud_occ config:system:set enabledPreviewProviders 11 --value="OC\\Preview\\OpenDocument"
+nextcloud_occ config:system:set enabledPreviewProviders 12 --value="OC\\Preview\\Krita"
+nextcloud_occ config:system:set enabledPreviewProviders 13 --value="OC\\Preview\\BMP"
+nextcloud_occ config:system:set enabledPreviewProviders 14 --value="OC\\Preview\\Movie"
+nextcloud_occ config:system:set enabledPreviewProviders 15 --value="OC\\Preview\\Imaginary"
+nextcloud_occ config:system:set preview_imaginary_url --value="http://127.0.0.1:9000"
+
+# Set values
+nextcloud_occ config:app:set previewgenerator squareSizes --value="32 256"
```
szaimen, line 122: This is not taken into account by imaginary afaik

addons/imaginary.sh
```diff
@@ -0,0 +1,144 @@
...
+nextcloud_occ config:app:set previewgenerator widthSizes  --value="256 384"
+nextcloud_occ config:app:set previewgenerator heightSizes --value="256"
+nextcloud_occ config:system:set preview_max_x --value="2048"
+nextcloud_occ config:system:set preview_max_y --value="2048"
+nextcloud_occ config:system:set jpeg_quality --value="60"
+nextcloud_occ config:system:set preview_max_memory --value="128"
```
szaimen, line 128: 256 is the default so not sure why lowering it to 128?

addons/imaginary.sh
```diff
@@ -0,0 +1,144 @@
...
+nextcloud_occ config:app:set preview jpeg_quality --value="60"
+
+# Add logs
+touch "$VMLOGS"/previewgenerator.log
+chown www-data:www-data "$VMLOGS"/previewgenerator.log
+
+# Rebuild
+print_text_in_color "$ICyan" "Scanning Nextclouds appdata directory to rebuild all previews all previews. This will take a while..."
+nextcloud_occ files:scan-app-data -vvv
+nextcloud_occ preview:generate-all --verbose >> "$VMLOGS"/previewgenerator.log
```
szaimen, line 138: Not needed...

szaimen:
Maybe also this discussion would be interesting for you to read: https://github.com/nextcloud/photos/issues/1492#issuecomment-1485200306

addons/imaginary.sh
szaimen: Same with height size

addons/imaginary.sh
szaimen: Same with heightsize

szaimen:
Additionally, this hint might be useful: https://help.nextcloud.com/t/26-0-php-extension-sysvsem/158819/5

addons/imaginary.sh
enoch85: I think it's a cleaner solution.

addons/imaginary.sh
enoch85: Ok, thanks!

addons/imaginary.sh
enoch85: Hmm, don't remember actually.

addons/imaginary.sh
enoch85: But I want to scan all the existing data as well... I took it from this post: https://okxo.de/speed-up-nextcloud-preview-generation-with-imaginary/

enoch85:
Thanks for your review @szaimen, very helpful! 

I will have a closer look tonight.

addons/imaginary.sh
enoch85: But it's needed for the "rebuild", right?

addons/imaginary.sh
enoch85: But yeah, maybe not, since we actually have a `funtion` for stopping... Will think about this.

enoch85:
Hmm, ok, so basically Imaginary does not need a cronjob since it doesn't `pre-generate` previews, it happens on the fly. And that's why the app Preview Generator app is not needed, since there's no need to `pre-generate` previews anymore due to extreme server load - since Imaginary is capped to maximum 50 concurrent threads...? 

Did I get it right? :)

addons/imaginary.sh
szaimen: Hm... Imaginary is not coupled with the previewgenerator...

szaimen:
> Hmm, ok, so basically Imaginary does not need a cronjob since it doesn't `pre-generate` previews, it happens on the fly

Indeed. 

> And that's why the app Preview Generator app is not needed, since there's no need to `pre-generate` previews anymore due to extreme server load - since Imaginary is capped to maximum 50 concurrent threads...?

No, the limit was implemented with https://github.com/nextcloud/server/pull/18210 in Nextcloud 26. But indeed is due to this the previewgenerator not necessary anymore :)

addons/imaginary.sh
enoch85: Ok, so a better way forward would be to uninstall previewgeneration and cleanup all generated previews, then the run a appfolder scan, and let imaginary handle the rest?

enoch85:
I think this is ready, can you please double check @szaimen?

apps/imaginary.sh
```diff
@@ -0,0 +1,153 @@
...
+services:
+  imaginary:
+    image: nextcloud/aio-imaginary:latest
+    container_name: imaginary
+    restart: always
+    environment:
+       PORT: 9000
+    command: -concurrency 50 -enable-url-source -log-level debug
+    ports:
+      -  127.0.0.1:9000:9000
```
szaimen, line 107: I still think using docker-compose here is unecessary

apps/imaginary.sh
```diff
@@ -0,0 +1,153 @@
...
+fi
+
+# Install dependencies for Imaginary
+check_php
+install_if_not php"$PHPVER"-sysvsem
+
+# Set default limits
+# https://github.com/nextcloud/server/pull/18210/files#diff-3bbe91e1f85eec5dbd0031642dfb0ad6749b550fc3b94af7aa68a98210b78738R1121
+nextcloud_occ config:system:set preview_concurrency_all --value="8"
+nextcloud_occ config:system:set preview_concurrency_new --value="4"
```
szaimen, line 133: these are the defaults so why do you add them?

apps/imaginary.sh
```diff
@@ -0,0 +1,153 @@
...
+
+# Set providers (https://github.com/nextcloud/server/blob/master/lib/private/Preview/Imaginary.php#L60)
+# https://github.com/nextcloud/vm/pull/2464#discussion_r1155074227
+# This is handled by Imagniary itself
+nextcloud_occ config:system:set enabledPreviewProviders --value="OC\\Preview\\Imaginary"
```
szaimen, line 138:
```suggestion
nextcloud_occ config:system:set enabledPreviewProviders 0 --value="OC\\Preview\\Imaginary"
```

szaimen: also you will need to add additionaly like the Movie Preview Provider and others that are not included in imaginary IIRC

apps/imaginary.sh
```diff
@@ -0,0 +1,153 @@
...
+nextcloud_occ config:system:set preview_imaginary_url --value="http://127.0.0.1:9000"
+
+# Set general values
+nextcloud_occ config:system:set preview_max_x --value="2048"
+nextcloud_occ config:system:set preview_max_y --value="2048"
+nextcloud_occ config:system:set jpeg_quality --value="60"
+nextcloud_occ config:system:set preview_max_memory --value="256"
```
szaimen, line 145: why set the default?

apps/imaginary.sh
```diff
@@ -0,0 +1,153 @@
...
+    # Remove apps
+    APPS=(php-imagick libmagickcore-6.q16-3-extra imagemagick-6.q16-extra)
+    for app in "${APPS[@]}"
+    do
+        if is_this_installed "$app"
+        then
+            apt-get purge "$app" -y
+        fi
+    done
+    if is_this_installed ffmpeg && ! is_app_installed integration_whiteboard
```
szaimen, line 70: I think ffmpeg is still needed for movie previews

apps/imaginary.sh
```diff
@@ -0,0 +1,153 @@
...
+    fi
+    apt-get autoremove -y
+    if yesno_box_yes "Do you want to remove all previews that were generated until now?
+This will most likely clear a lot of space! Also, pre-generated previews are not needed anymore once Imaginary are installed."
+    then
+        countdown "Removing the preview folder. This can take a while..." "5"
+        rm -rfv "$NCDATA"/appdata_*/preview
+        print_text_in_color "$ICyan" "Scanning Nextclouds appdata directory after removing all previews. \
+This can take a while..."
+        nextcloud_occ files:scan-app-data -vvv
```
szaimen, line 82: This should be enough:

```suggestion
        nextcloud_occ files:scan-app-data preview -vvv
```

enoch85: Thanks

apps/imaginary.sh
enoch85: Yup, correct. Just copying from the old script. :+1:

apps/imaginary.sh
enoch85: So that you easier can change if you want later on. Correct values in config.php, even if it's default.

apps/imaginary.sh
enoch85: As I said before, easier to change later on.

apps/imaginary.sh
enoch85: But you said here that it's not needed? https://github.com/nextcloud/vm/pull/2464#discussion_r1155074227 Wich ones should I add you mean?

szaimen: The ones that are covered by imaginary are not needed but the ones that are not are needed. does this make sense?

apps/imaginary.sh
```diff
@@ -116,16 +123,28 @@ nextcloud_occ config:system:set preview_concurrency_new --value="4"
...
+nextcloud_occ config:system:set enabledPreviewProviders 6 --value="OC\\Preview\\HEIC"
+nextcloud_occ config:system:set enabledPreviewProviders 7 --value="OC\\Preview\\HEIF"
+nextcloud_occ config:system:set enabledPreviewProviders 8 --value="OC\\Preview\\SVG"
+nextcloud_occ config:system:set enabledPreviewProviders 9 --value="OC\\Preview\\TIFF"
+nextcloud_occ config:system:set enabledPreviewProviders 10 --value="OC\\Preview\\WEBP"
+nextcloud_occ config:system:set enabledPreviewProviders 11 --value="OC\\Preview\\XML"
+nextcloud_occ config:system:set enabledPreviewProviders 12 --value="OC\\Preview\\PDF"
+nextcloud_occ config:system:set enabledPreviewProviders 13 --value="OC\\Preview\\Ilustrator"
+nextcloud_occ config:system:set enabledPreviewProviders 14 --value="OC\\Preview\\Movie"
+nextcloud_occ config:system:set enabledPreviewProviders 15 --value="OC\\Preview\\Imaginary"
```
szaimen, line 140: As I said only the ones that are not covered by imaginary need to get enabled:

```suggestion
nextcloud_occ config:system:set enabledPreviewProviders 0 --value="OC\\Preview\\Imaginary"
nextcloud_occ config:system:set enabledPreviewProviders 1 --value="OC\\Preview\\Image"
nextcloud_occ config:system:set enabledPreviewProviders 2 --value="OC\\Preview\\MarkDown"
nextcloud_occ config:system:set enabledPreviewProviders 3 --value="OC\\Preview\\MP3"
nextcloud_occ config:system:set enabledPreviewProviders 4 --value="OC\\Preview\\TXT"
nextcloud_occ config:system:set enabledPreviewProviders 5 --value="OC\\Preview\\OpenDocument"
nextcloud_occ config:system:set enabledPreviewProviders 6 --value="OC\\Preview\\Movie"
```

apps/imaginary.sh
```diff
@@ -87,7 +87,7 @@ This will most likely clear a lot of space! Also, pre-generated previews are not
             rm -rfv "$NCDATA"/appdata_*/preview
             print_text_in_color "$ICyan" "Scanning Nextclouds appdata directory after removing all previews. \
 This can take a while..."
-            nextcloud_occ files:scan-app-data preview -vvv
+            nextcloud_occ files:scan-app-data -vvv
```
szaimen, line 90: is this not working or why did you remove it again?

apps/imaginary.sh
enoch85: But there are more than you added...?

return '/(image\/(bmp|x-bitmap|png|jpeg|gif|heic|heif|svg\+xml|tiff|webp)|application\/(pdf|illustrator))/'; Shouldn't those be added as well?

apps/imaginary.sh
szaimen: the trick is to remove only the content and not the folder itself

enoch85: OK, gotcha. :+1:

szaimen (approved):
LGTM but didnt test

PR #2470, opened 2023-04-06
Don't run Imaginary as the root user

enoch85:


szaimen:
I think this whole pr is not needed since imaginary inside the container runs as mobody user iirc

enoch85:
> I think this whole pr is not needed since imaginary inside the container runs as nobody user iirc

So [this line](https://github.com/nextcloud/all-in-one/blob/main/Containers/imaginary/Dockerfile#L21) would mean that no matter if you run the "top layer" as root, it still uses `nobody` and nothing can be executed?

enoch85:
To test:

1. Run this PR and execute: `docker attach <container id>`
2. Run the *original script* and execute: `docker attach <container id>`

Any difference?

Issue #1986, opened 2021-05-22
Make LVM snapshot default when no issues get reported


szaimen:
This is just a reminder that we don't forget to make the LVM snapshot default when no issues get reported.
https://github.com/nextcloud/vm/blob/39e64fe07920bea14f064abaf71847cd5c7165a3/nextcloud_install_production.sh#L68
After we do this, everyone wil be able to use the built-in backup solution.

enoch85:
I'd say when development stopped on that part for some time because it's rock solid, then maybe. :)

szaimen:
I'd say it is already pretty stable but yeah

enoch85:
We could do this for Ubuntu 22.04 making the OS disk 45 GB in size, or extend the drive so there's only 5 GB left and keep it 40 GB in total size.

Would that work?

enoch85:
cc @small1

szaimen:
> We could do this for Ubuntu 22.04 making the OS disk 45 GB in size, or extend the drive so there's only 5 GB left and keep it 40 GB in total size.
> 
> Would that work?

From my side, yes 👍

szaimen:
Does it happen every time you run the update script?

szaimen:
Could be a bug with lvm...
https://blog.roberthallam.org/2017/12/solved-logical-volume-is-used-by-another-device/

Could you please try the following commands and post the output of those here (if it should still happen)?

```
lvremove -v /dev/ubuntu-vg/NcVM-snapshot

dmsetup info -c | grep NcVM | grep snapshot

# more to come when we have more info based on the guide linked above

```

szaimen:
@enoch85 do you have some feedback here? It is hard to debug without a way to reproduce this issue...

enoch85:
As it's not in the released version yet, please add a PR with the fix you proposed, and I'll run one of the auto update VMs with the new setup.

szaimen:
I can try. But after reading through the code, did you try to reboot the affected server once after you got the notification that the update failed because of the failed lvremove?
![image](https://user-images.githubusercontent.com/42591237/142603189-1e79a4f8-b170-43c3-8b18-196b594f790f.png)

enoch85:
I've only seen this once, and I'm not sure if the server was rebooted or not. 

If you think it can be improved, then do so, else leave it for now.

szaimen:
Thanks for the feedback! 
[...]

PR #2476, opened 2023-04-24
automatically reboot server if ncvm-snapshot could not get removed

szaimen:
See https://github.com/nextcloud/vm/issues/1986#issuecomment-1503182990


enoch85:
Thanks!

PR #2479, opened 2023-04-27
Update Imaginary values

enoch85:
cc  @szaimen Your take on this?

It's according to docs. Are there more formats supported by Imagniary? I saw that Krita was another supported format?

### DIFF
--- a/apps/imaginary.sh
+++ b/apps/imaginary.sh
@@ -102,7 +102,7 @@ install_docker
 
 # Pull and start
 docker pull nextcloud/aio-imaginary:latest
-docker run -t -d -p 127.0.0.1:9000:9000 --restart always --name imaginary nextcloud/aio-imaginary -concurrency 50 -enable-url-source -log-level debug
+docker run -t -d -p 127.0.0.1:9000:9000 --restart always --name imaginary nextcloud/aio-imaginary –cap-add=sys_nice -return-size -concurrency 50 -enable-url-source -log-level debug
 
 # Test if imaginary is working
 countdown "Testing if it works in 3 sedonds" "3"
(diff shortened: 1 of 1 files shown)


apps/imaginary.sh
```diff
@@ -102,7 +102,7 @@ install_docker
 
 # Pull and start
 docker pull nextcloud/aio-imaginary:latest
-docker run -t -d -p 127.0.0.1:9000:9000 --restart always --name imaginary nextcloud/aio-imaginary -concurrency 50 -enable-url-source -log-level debug
+docker run -t -d -p 127.0.0.1:9000:9000 --restart always --name imaginary nextcloud/aio-imaginary –cap-add=sys_nice -return-size -concurrency 50 -enable-url-source -log-level debug
```
szaimen, line 105: return size is already added by default
```suggestion
docker run -t -d -p 127.0.0.1:9000:9000 --restart always --name imaginary nextcloud/aio-imaginary –cap-add=sys_nice -concurrency 50 -enable-url-source -log-level debug
```

szaimen: Actually -concurrency 50 is imaginarys default as well afaik...

apps/imaginary.sh
```diff
@@ -141,6 +141,7 @@ nextcloud_occ config:system:set enabledPreviewProviders 4 --value="OC\\Preview\\
 nextcloud_occ config:system:set enabledPreviewProviders 5 --value="OC\\Preview\\OpenDocument"
 nextcloud_occ config:system:set enabledPreviewProviders 6 --value="OC\\Preview\\Movie"
 nextcloud_occ config:system:set enabledPreviewProviders 7 --value="OC\\Preview\\PDF"
+nextcloud_occ config:system:set enabledPreviewProviders 8 --value="OC\\Preview\\HEIC"
```
szaimen, line 144:
```suggestion
```

szaimen, line 144:
```suggestion
```

enoch85:
Why not PDF and HEIC? I saw that you added support for PDF in the last release.

szaimen:
> Why not PDF and HEIC? I saw that you added support for PDF in the last release.

because it is already included here: https://github.com/nextcloud/server/blob/5e96228eb1f7999a327dacab22055ec2aa8e28a3/lib/private/Preview/Imaginary.php#L60

enoch85:
> > Why not PDF and HEIC? I saw that you added support for PDF in the last release.
> 
> because it is already included here: https://github.com/nextcloud/server/blob/5e96228eb1f7999a327dacab22055ec2aa8e28a3/lib/private/Preview/Imaginary.php#L60

Ah ok, but still strange that PDFs weren't generated without adding it as enabled. I'll need to double check.

Anyway, are there any more formats that can be enabled?

szaimen (approved):
LGTM

PR #2487, opened 2023-05-05
Improve solvable_files.sh from Nextcloud GmbH

enoch85:


enoch85:
cc @szaimen maybe you can report that the original script didn't work. Made some improvements, but can't push to Nextcloud GHMB.

The defaults here are:
- `fix`
- `fix use_birthday`
- `verbose`
- `noscan`

szaimen:
> cc @szaimen maybe you can report that the original script didn't work. Made some improvements, but can't push to Nextcloud GmbH.

You could create an issue here: https://github.com/nextcloud-gmbh/mtime_fixer_tool_kit/issues

PR #2497, opened 2023-05-27
deprecated harden-ssh

szaimen:
This disallowed my ssh user from logging in. But maybe a restart would have fixed it. Needs investigation.


enoch85:
@szaimen Close?

szaimen:
> @szaimen Close?

Still need to do investigation on this...

szaimen:
Lets merge this 👍

Issue #2519, opened 2023-07-18
Implement Talk recording


enoch85:
@szaimen How hard you you think it would be to implement talk recording in the VM using the docker config from AIO? Just noticed it today, great work! :raised_hands: 

Would it be sufficient to download [this](https://github.com/nextcloud/all-in-one/tree/main/Containers/talk-recording) and run start.sh?

szaimen:
In theory it shouldnt be that difficult but you will need to do some changes to the talk config first. You can compare with https://github.com/nextcloud/all-in-one/blob/main/Containers/talk/start.sh.

Afterwards something like this could be used in principle: https://github.com/nextcloud/all-in-one/blob/70b46093ab2a5915a7844d30dd4067e966640ecb/manual-install/latest.yml#L184-L198 (but I would prefer again to not use docker-compose but instead normal docker run command for that).

szaimen:
> Since you are maintainer here as well, do you think you could start something small just to give me pointers that I can work on, kind of like with imaginary?

Sorry, I dont have the time to look into this currently. I hoped that my pointers would have been enough? I would suggest that you start the PR and I add some comments. WDYT?

> I mean, we already have Talk and Signaling just like in AIO, so I guess it's just a matter of changing the name of the $vars?

yes, basically

enoch85:
Sure, will make an effort when *I* have time 😄

PR #2520, opened 2023-07-19
Implement Talk recording

enoch85:
Fix https://github.com/nextcloud/vm/issues/2519

cc @szaimen

### DIFF
--- a/apps/talk.sh
+++ b/apps/talk.sh
@@ -4,7 +4,7 @@
 
 true
 SCRIPT_NAME="Nextcloud Talk"
-SCRIPT_EXPLAINER="This script installs Nextcloud Talk whcih is a replacement for Teams/Skype and similar.\
+SCRIPT_EXPLAINER="This script installs Nextcloud Talk which is a replacement for Teams/Skype and similar.\
 You will also be offered the possibility to install the so-called High-Performance-Backend, which makes it possible to host more video calls than it would be with the standard Talk app. \
 It's called 'Talk Signaling' and you will be offered to install it as part two of this script."
 # shellcheck source=lib.sh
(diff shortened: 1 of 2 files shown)


apps/talk.sh
```diff
@@ -486,3 +486,35 @@ else
...
+shm_size=2g \
+-e NC_DOMAIN=${TURN_DOMAIN} \
+-e TZ=${TIMEZONE} \
+-e RECORDING_SECRET=${TURN_RECORDING_SECRET} \
+-e INTERNAL_SECRET=${TURN_INTERNAL_SECRET}
+
+# Talk recording
+if [ -d "$NCPATH/apps/spreed" ]
+then
+    if [ "$TALK_RECORDING_ENABLED" = 'yes' ]
```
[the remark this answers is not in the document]
enoch85: @szaimen Never found the check you run here, but I guess it's something like `is_this_docker_running talk_recording`?

apps/talk.sh
```diff
@@ -486,3 +488,33 @@ else
     msg_box "Congratulations, everything is working as intended! The installation succeeded.\n\nLogging can be found by typing: journalctl -lfu signaling"
     exit 0
 fi
+
+####### Talk recording
+print_text_in_color "$ICyan" "Setting up Talk recording..."
```
szaimen, line 493: I would recommend to make this optional as talk recording is very resource intensive.

enoch85: Yeah sure, been thinking about that too.

apps/talk.sh
```diff
@@ -325,7 +327,7 @@ debug = false
 hashkey = $(openssl rand -hex 16)
 blockkey = $(openssl rand -hex 16)
 [clients]
-internalsecret = $(openssl rand -hex 16)
+internalsecret = $(TURN_INTERNAL_SECRET)
```
szaimen, line 330:
```suggestion
internalsecret = ${TURN_INTERNAL_SECRET}
```

enoch85:
@szaimen What do you think? Did I understand you correctly or is something wrong here?

Not tested.

apps/talk.sh
```diff
@@ -483,6 +485,43 @@ then
...
+
+# It's pretty recource intensive
+cpu_check 4
+
+print_text_in_color "$ICyan" "Setting up Talk recording..."
+
+# Pull and start
+docker pull nextcloud/aio-talk-recording:latest
+docker run -t -d -p "$TURN_RECORDING_HOST":"$TURN_RECORDING_HOST_PORT":"$TURN_RECORDING_HOST_PORT" \
+--restart \
```
szaimen, line 505:
```suggestion
--restart always \
```

apps/talk.sh
```diff
@@ -483,6 +485,43 @@ then
...
+--name talk-recording nextcloud/aio-talk-recording \
```
szaimen, line 506:
```suggestion
--name talk-recording \
```

szaimen:
> @szaimen What do you think? Did I understand you correctly or is something wrong here?
> 
> Not tested.

lgtm in general. However If you compare the talk-hpb with https://github.com/nextcloud/all-in-one/blob/main/Containers/talk/start.sh, the turn section can be removed

enoch85:
> However If you compare the talk-hpb with https://github.com/nextcloud/all-in-one/blob/main/Containers/talk/start.sh, the turn section can be removed

Hmm, not so sure about that. I've seen it in an enterprise setup as well. As you already know, we use @morph027 setup for Signaling, and the `JANUS_KEY` is there, not added by me, and it works as is. This is just an addition.

I suppose you mean this: https://github.com/nextcloud/vm/blob/master/apps/talk.sh#L341-L343

szaimen:
> I suppose you mean this: https://github.com/nextcloud/vm/blob/master/apps/talk.sh#L341-L343

yes

enoch85:
> > I suppose you mean this: https://github.com/nextcloud/vm/blob/master/apps/talk.sh#L341-L343
> 
> yes

I'll leave it, don't know the side effects if I remove it.

szaimen:
> I'll leave it, don't know the side effects if I remove it.

The side-effect of leaving it could be that the recording server does not work.

enoch85:
> > I'll leave it, don't know the side effects if I remove it.
> 
> The side-effect of leaving it could be that the recording server does not work.

Actually, we have that in another setup (not your container, and my colleges installation), and it's working there. Do you mean that your container doesn't handle it, or what's the reason for it stop working?

Also, it should only be removed if talk recording is turned on?

szaimen:
> Actually, we have that in another setup (not your container, and my colleges installation), and it's working there. Do you mean that your container doesn't handle it, or what's the reason for it stop working?
> 
> Also, it should only be removed if talk recording is turned on?

The config is simply wrong as far as I know. The talk-hpb does not need a turn server afaik. Turn is only used by the android and ios apps directly.

enoch85:
> The config is simply wrong as far as I know.

Can you give your take on this?

I will merge this now for easier testing. We can continue the discussion in this PR though.

PR #2535, opened 2023-07-31
fix maintenance:mode

enoch85:
cc @szaimen


enoch85:
I think this will fix the issue.

szaimen:
Thanks 💙 

I guess I'll see if this works tomorrow morning and will let you know :)

PR #2541, opened 2023-08-10
Change to ElasticSearch for FTS

enoch85:
This will replace OpenSearch with Elasticsearch. Based on Elasticsearch own Docker image, with some extra added security.

FYI @Ark74 @small_1


enoch85:
Hmm, maybe we can use Elasticsearch directly? https://geshan.com.np/blog/2023/06/elasticsearch-docker/

PR #2556, opened 2023-08-27
Make PostgreSQL connection persistent

enoch85:
Not yet introduced afaik.

cc @szaimen Do you know?


enoch85:
https://help.nextcloud.com/t/persistent-database-postgresql-connections/24307/5?u=enoch85

szaimen:
Indeed there is no install value for this available afaik. In AIO I set this after install thus.

enoch85:
Hmm OK, so maybe a `config:set` is better?

Doesn't the DB need to be persistent from the beginning, and what are the benefits/disadvantages?

szaimen:
> Hmm OK, so maybe a `config:set` is better?

yes, should be sufficient

> Doesn't the DB need to be persistent from the beginning, and what are the benefits/disadvantages?

No, I dont think it needs to be persistent from the beginning. The value can be changed during runtime.

szaimen:
I cannot recommend doing  this at all. It causes problems thus it was reverted in aio: https://github.com/nextcloud/all-in-one/pull/3479

enoch85:
> I cannot recommend doing this at all. It causes problems thus it was reverted in aio: [nextcloud/all-in-one#3479](https://github.com/nextcloud/all-in-one/pull/3479)

Thanks for the heads up! 

Though, we also have this issue in the VM without making it persistent, so I think the issue is deeper than this... Closing anyway.

Issue #2587, opened 2023-11-08
Install php8.1-sqlite3 by default


pjrobertson:
This is for searching in the Collectives app. See the discussion at: https://help.nextcloud.com/t/searching-in-collectives/116551/7?u=pjrobertson

Basically, to allow searching within collectives content, you need to have the `ext-pdo_sqlite` PHP extension installed. It's not installed by default on a nextcloud VM.

I recommend adding the following line somewhere in the install process:

`apt-get install php8.1-sqlite3`

Thanks!

PR #2590, opened 2023-11-09
Install sqlite for Collectives

enoch85:
Fix https://github.com/nextcloud/vm/issues/2587


enoch85:
@OkhamG So you mean the whole SQLite database is needed? Do you have a link to docs?

Thanks!

PR #2601, opened 2023-12-13
Adjust file ownership and permissions

admarty:
Adjust file ownership and permissions before reboot


enoch85:
Thanks for your PR, but why do you need this?

We already have a script that sets permissions for everything.

admarty:
I just spin up a VM Ubuntu-22-LTS to install using the above script and here is the result:

image

enoch85:
Yes, those are the expected permissions.

enoch85:
Here's the script in question: https://github.com/nextcloud/vm/blob/master/static/setup_secure_permissions_nextcloud.sh

Issue #2481, opened 2023-04-30
Option to keep .bash_history


github-tomster:
**Is your feature request related to a problem? Please describe.**
usually one can use the bash history unfortunately not on this VM 🙇🏻 
`/var/scripts/history.sh` prevents that 🗑️ 

**Describe the solution you'd like**
is there a non obvious reason for `truncate -s0 "$HOME/.bash_history"
` ❓ 

**Describe alternatives you've considered**
can one disable `history.sh` without breaking the VM ❓ 

**Additional context**
i searched the repo but i am not sure if this ist the first issue/question regarding the missing history 🤷🏻

enoch85:
> can one disable history.sh without breaking the VM ❓

Yes, just comment it out in the `.bash_profile`

The idea behind it was to remove commands which the clumsy user had been making containing passwords and stuff - see it as a security mesure.

github-tomster:
@enoch85 
thank you for confirming 👍🏻 
maybe a hint/comment like:
```
bash /var/scripts/nextcloud.sh
# comment out following line to keep bash_history at own risk!
bash /var/scripts/history.sh
```
could prevent such noise 🤷🏻

enoch85:
Uninitiated user: Didn't even know history was saved per default
Power user: Ask the same question as you, finds the answer and fixes it. ;)

github-tomster:
@enoch85 
was the answer already out there 🤔 
i mean the 'will it break the nextcloudvm environment' question ❓

pjrobertson:
I just came across this 'issue' of .bash_history being cleared and was confused why. Some digging led me to the `history.sh` file and then here. I've created a PR to at least add this comment in the history.sh file see: https://github.com/nextcloud/vm/pull/2604

PR #2604, opened 2023-12-22
Add comment/documentation on why .bash_history is cleared by default

pjrobertson:
Also see https://github.com/nextcloud/vm/issues/2481


enoch85:
Thanks! :+1:

PR #2612, opened 2024-01-24
fix pi-hole script

szaimen:


not-supported/pi-hole.sh
```diff
@@ -192,6 +192,9 @@ sed -i "s|getStaticIPv4Settings$|echo getStaticIPv4Settings|" "$SCRIPTS"/pihole-
 PIHOLE_INTERFACE="$IFACE"
 export PIHOLE_INTERFACE
 
+# Fix php versions getting hold for pi-hole install script
+apt-mark unhold php"$PHPVER"*
```
szaimen, line 196: @enoch85 maybe we should add this line also below here?
https://github.com/nextcloud/vm/blob/3149aad66463cdead32ae2eb9b1f98fc56e7b164/nextcloud_update.sh#L350

szaimen: Otherwise the php8.1 packages are never going to be upgraded...

PR #2651, opened 2024-07-04
Create Http azure

abdukhliq:
az network nsg rule create --resource-group "[sandbox resource group name]" --nsg-name my-vmNSG --name allow-http --protocol tcp --priority 100 --destination-port-range 80 --access Allow


enoch85:
Sorry, I don't understand this PR. Can you please explain why this is needed in a separate file?

PR #2663, opened 2024-09-12
Automatically update Geoblock files

enoch85:
@szaimen GeoIP is your implementation, but it doesn't automatically fetch latest geofile. To solve that, I made this.

I found that https://www.miyuru.lk/geoiplegacy actually posts the month and year it's been updated, so I thought that we could scrape that info with `curl`, and then compare with the current file on the system - and that's what this change does.

Please give your opinion. Maybe there's a smarter way?


enoch85:
Note to self: check connectivity before proceeding.

enoch85:
Tested and works, I'll merge this now!

szaimen:
Hi @enoch85,

sorry for coming back to you. I was very busy. 

The reason why I did not implement it like this was this warning on https://www.miyuru.lk/geoiplegacy:
image
So I used Github as a relay basically.

szaimen:
Btw, yesterday was my server automatically updated and it ran into the following issue:
image
This lead to ipv4 connections not working correctly anymore. 
So it seems like there is something wrong with your implementation.
Shall I open a bug report?

enoch85:
> Hi @enoch85,
> 
> sorry for coming back to you so late. I was very busy.
> 
> The reason why I did not implement it like this was this warning on [image] So I used Github as a relay basically.

Valid point. I actually asked the owner of the website myself, and he's fine with it! :)

> This lead to ipv4 connections not working correctly anymore.

So the filename _has_ to be `GeoIP.dat`? I find it strange that you can't set you own name in the config...

szaimen:
> > This lead to ipv4 connections not working correctly anymore.
> 
> So the filename _has_ to be `GeoIP.dat`? I find it strange that you can't set you own name in the config...

No, the problem is that the file that is on my system does not match the config. Somehow this got messed up after your PR...

enoch85:
So, what's the output of:

```
ls -l /usr/share/GeoIP
cat /etc/apache2/apache2.conf
```

szaimen:
I posted the information in the screenshot above already...

enoch85:
Sorry... 

I'll have another look and check if we can fix it, probably an easy fix. :)

enoch85:
Can you try to run this:

```
#!/bin/bash

# Get newest dat files for geoblock.sh
if grep -q "^#Geoip-block" /etc/apache2/apache2.conf
then
    if get_newest_dat_files
    then
        if grep -c GeoIP.dat /etc/apache2/apache2.conf
        then
            if [ ! -f /usr/share/GeoIP/GeoIPv4.dat ]
            then
                if download_geoip_dat 4 v4
                then
                    sed -i "s|GeoIPDBFile /usr/share/GeoIP/GeoIP.dat|GeoIPDBFile /usr/share/GeoIP/GeoIPv4.dat|g" /etc/apache2/apache2.conf
                fi
            fi
        fi
        check_command systemctl restart apache2
    fi
fi
```

PR #2668, opened 2024-09-15
download latest GeoIP.dat no matter what

enoch85:
cc @szaimen


enoch85:
Tested and works for me at least, let's not break any more installations --> merge

Issue #2674, opened 2024-09-17
Nextcloud VM GeoBlock broken


aaaskew:
### Steps To Reproduce

1. Use menu.sh to add Geo blocking
2. Select just countries
3. Select my country

### Expected Result

Geoblocking IPs stops people outside the country from connecting

### Actual Result

Everyone on IPv4 (I think) is blocked with

Forbidden
You don't have permission to access this resource.

Additionally, a 403 Forbidden error was encountered while trying to use an ErrorDocument to handle the request.

### Screenshots, Videos, or Pastebins

![ipv4](https://github.com/user-attachments/assets/05fee553-c01c-4849-a835-a62fee29aa8e)

![ipv6](https://github.com/user-attachments/assets/e58252f4-9b54-4b93-a382-04606c11adce)

### Additional Context

This affects anyone using GeoBlock installed already or fresh install VMs.

2 issues:

1. On fresh install the install pops-up 2 error messages. These might not actually be a problem? I don't know.

2. The GeoIP configuration in apache.conf points to `/usr/share/GeoIP/GeoIPv4.dat` that does not exist.

`/usr/share/GeoIP/GeoIP.dat` and `/usr/share/GeoIPv6.dat` do exist however so I expect the apache.conf configuration needs changing for fresh installs, but also for people running update.sh?

### Build Version

29

### Environment

By downloading the VM

enoch85:
Hmm, this is a new one...

I was trying to be cautious since we're downloading from an external source. I don't want the users to get any virus. :| 

Maybe if you can find a better source for the download which we can trust? Then I can make a new PR for this.

enoch85:
This is what happens in the background:

```
Checking connection to https://www.miyuru.lk/geoiplegacy...
++ curl -LI https://www.miyuru.lk/geoiplegacy -o /dev/null -w '%{http_code}\n' -s
+ CURL_STATUS=200
+ [[ 200 = \2\0\0 ]]
+ return 0
+ curl -s https://www.miyuru.lk/geoiplegacy
+ grep -q 'September 2024'
+ grep -c GeoIP.dat /etc/apache2/apache2.conf
1
+ '[' '!' -f /usr/share/GeoIP/GeoIPv4.dat ']'
+ download_geoip_dat 4 v4
+ site_200 https://dl.miyuru.lk/geoip/maxmind/country/maxmind4.dat.gz
[...]

PR #2675, opened 2024-09-17
Rewrite Geoblock (use new mmdb)

enoch85:
@szaimen @aaaskew  Please test this PR. Please note it requires some new functions, please see the lib file in this PR.

FIx https://github.com/nextcloud/vm/issues/2674

### DIFF
--- a/network/geoblock.sh
+++ b/network/geoblock.sh
@@ -55,12 +55,17 @@ else
     fi
     # Remove  Apache config
     sed -i "/^#Geoip-block-start/,/^#Geoip-block-end/d" /etc/apache2/apache2.conf
+    if [ -f "$GEOBLOCK_MOD_CONF" ]
+    then
+        a2disconf geoblock
+        rm -f "$GEOBLOCK_MOD_CONF"
+    fi
     # Show successful uninstall if applicable
     removal_popup "$SCRIPT_NAME"
     # Make sure it's clean from unused packages and files
     apt purge libmaxminddb0* libmaxminddb-dev* mmdb-bin* apache2-dev* -y
     apt autoremove -y
-    rm -rf /usr/share/GeoIP
+    #rm -rf /usr/share/GeoIP keep these to save downloads...
     check_command systemctl restart apache2
 fi
 
--- a/network/geoblock.sh
+++ b/network/geoblock.sh
@@ -53,9 +53,11 @@ else
     # Show successful uninstall if applicable
     removal_popup "$SCRIPT_NAME"
     # Make sure it's clean from unused packages and files
-    apt purge libmaxminddb0* libmaxminddb-dev* mmdb-bin* -y
+    apt purge libmaxminddb0* libmaxminddb-dev* mmdb-bin* apache2-dev* -y
     apt autoremove -y
+    rm -f /usr/lib/apache2/modules/mod_maxminddb.so
     rm -rf /usr/share/GeoIP
+    systemctl apache2 restart
 fi
 
 # Download GeoIP Databases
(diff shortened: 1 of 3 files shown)


network/geoblock.sh
```diff
@@ -160,22 +172,31 @@ then
...
+echo "MaxMindDBFile CITY_DB /usr/local/share/maxminddb/GeoLite2-City.mmdb"
+} > /etc/apache2/mods-available/maxminddb.conf
+
+
 GEOIP_CONF="#Geoip-block-start - Please don't remove or change this line
-<IfModule mod_geoip.c>
-  GeoIPEnable On
-  GeoIPDBFile /usr/share/GeoIP/GeoIPv4.dat
-  GeoIPDBFile /usr/share/GeoIP/GeoIPv6.dat
+<IfModule mod_maxminddb.c>
```
szaimen, line 184: Are you sure this mod is still going to run on ubuntu 22.04? If not, we are going to have a regression...

nextcloud_update.sh
```diff
@@ -582,19 +582,20 @@ fi
 # Get newest dat files for geoblock.sh
 if grep -q "^#Geoip-block" /etc/apache2/apache2.conf
 then
-    if get_newest_dat_files
+    if grep -c GeoIPDBFile /etc/apache2/apache2.conf
     then
-        if grep -c GeoIP.dat /etc/apache2/apache2.conf
+        msg_box "We have updated GeoIP to a new version which isn't compatible with the old one. Please reinstall with the menu script to get the latest version."
```
szaimen, line 587: oh :/

enoch85: #progress sometimes hurt. :/

enoch85:
Canät test it right now since I've reached the limit. Have to wait until tomorrow.

network/geoblock.sh
enoch85: Works on Ubuntu 22.04, no errors anyway. Will continue to test. It' not done just yet. :)

enoch85:
This is now fully tested from me. Will merge later today, would be nice with some further testing though. Please dig in @szaimen @aaaskew 😄

aaaskew:
Many thanks for the change. Unfortunately I can't update to the new module. I get 'Update limit for Maxmind GeoDatabase reached! Please try again tomorrow.'.

I found the following: From March 2024; Every account is limited to 1,000 total direct downloads (30 for GeoLite accounts) in a 24-hour period. (https://support.maxmind.com/hc/en-us/articles/4408216129947-Download-and-Update-Databases)

This appears to come from lib.sh and I assume this is due to the hard-coding of an Account ID and license. This means that anyone will be using your credentials to download the database.

The download appears to be done with a Docker image? Not too sure about how this works but do we need to download the City IP database? I thought we are only using the country database:

echo "GEOIPUPDATE_EDITION_IDS=GeoLite2-City GeoLite2-Country"

I believe part of the reason for MaxMind deprecating the old database format is they could not control or monazite it for corporate users with the old APIs.

I have created my own login and can download the database myself. Maybe a few more instructions would be good it add to explain that if the download fails, the database could be sideloaded? I would think if the database already exists in the correct location then the update would not fail the download. People can create their own MaxMind account and license so that the downloads will be per deployment and therefore no longer hit a limit in lib.sh.

It will make the use of this module more of a pain to setup but I think it is the only realistic way to make sure it works for everyone.

enoch85:
Thanks for your input, I have now disabled the City DB. :) You might be correct that that's part of the problem as that takes one extra download. 

Regarding license, yes, it's the only way. You need an account. Anyone will be able to abuse it, but I hope they don't. To limit the attempts I chose to save the `mmdb` files instead of getting new ones every time. Now in the beginning of this merge, it will take a few days to push out the new DB, but ones the dust has settled I hope that we won't be rate limited as frequently.

The `mmdb` files ends up in `/usr/share/GeoIP` and you could create your own account and fetch put them there, or you just wait. :) Eventually it will work.

Another option I've thought about is to download some old files to this repo, and use them initially, then when the user updates it will update those files. But I don't know yet, we'll see how this pans out.

enoch85:
Wierd, what happens if you run this manually?

```
    if grep "Geoip-block-start" /etc/apache2/apache2.conf
    then
        sed -i "/^#Geoip-block-start/,/^#Geoip-block-end/d" /etc/apache2/apache2.conf
    fi
 ```

aaaskew:
That works but is the logic correct earlier.

I assume that line 26 of geoblock.sh should be saying 'If old geoblock is not installed AND new geoblock is not installed' then this is a fresh install, otherwise, uninstall?

enoch85:
> If old geoblock is not installed AND new geoblock is not installed' then this is a fresh install, otherwise, uninstall?

Well, good point. You're absolutely right!

abdullahdevrel:
I work for IPinfo, where we have a free IP to Country database with more lenient licensing terms and better accuracy.

This ticket has been merged and closed, so I have opened a new issue ticket pitching our free IP to Country database with more lenient license terms, which are particularly helpful for packaging and distributing data.

Our database is licensed under CC-BY-SA 4.0, which permits distribution (even for commercial purposes). We only require an attribution statement. This is one aspect of the database there is more. Please share your thoughts on my feature request ticket: https://github.com/nextcloud/vm/issues/2682

PR #2678, opened 2024-09-19
Update locate_mirror

enoch85:
Replace with a more modern take on it https://github.com/vegardit/fast-apt-mirror.sh

### DIFF
--- a/addons/locate_mirror.sh
+++ b/addons/locate_mirror.sh
@@ -48,7 +48,7 @@ else
         # Backup
         cp -f /etc/apt/sources.list /etc/apt/sources.list.backup
         # Replace
-        if fast-apt-mirror current --apply
+        if fast-apt-mirror current --apply # TODO is fast-apt-mirror.sh set better here?
         then
             msg_box "Your Ubuntu repo was successfully changed to $FASTEST_MIRROR"
         fi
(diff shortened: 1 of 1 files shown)


enoch85:
cc @aaaskew  Please test.

enoch85:
This is now tested and works, will merge to sort the bug and improve later.

aaaskew:
This has worked for me in Ubuntu 24 but my old upgraded to Ubuntu 22 system did something weird.

I am not sure if this is due to network issues or what it downloaded but it appears it tried and failed to upgrade the sources.list (fortunately)

This is what I saw:

Screenshot 2024-09-20 at 16 41 59

enoch85:
Hmm, maybe we should limit it to Ubuntu 24 then... Can you please try again?

Also, you have backups in `etc/apt/` ;)

aaaskew:
Tried again with Ubuntu 24 and 22 and it works for me with the re-test. I am assuming the error was due to a DNS or download failure that disappeared on retry. The sources.list did not actually get changed when that screenshot occured. The script just went a little mad. I assume it is due to some output being taken as what should be displayed in the dialog but that text is actually the script itself.

So consider the re-test ok for me.

enoch85:
Nice! Thanks for confirming. :+1:

PR #2681, opened 2024-09-21
also check for old installations (and reinstall)

enoch85:
@aaaskew


aaaskew:
Sorry but I would assume && as the operator rather than ||

enoch85:
Sorry, I'm tired today. 😲

Issue #2682, opened 2024-09-22
Feature request: Using IPinfo IP to Country database


abdullahdevrel:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

I am requesting to integrate [IPinfo's IP to Country database](https://ipinfo.io/products/free-ip-database) to the project. The database has the following features:

- It is updated daily, with zero compromise to accuracy. There is no range clustering, and the database provides full accuracy.
- The data granularity reaches individual IP level.
- The database comes in MMDB database format.
- It is licensed under CC-BY-SA 4.0, permitting commercial usage.
- Available file formats include: CSV, MMDB, JSON
- The data is tabular and unnested, making it very easy to use. The dataset includes both IPv4 and IPv6 in a single file.

**Describe the solution you'd like**

I would like the NextCloud VM to switch its IP to the Country database to IPinfo's free database because it is more accurate and reliable and has more lenient licensing terms.

**Describe alternatives you've considered**
I have not considered any alternatives.

**Additional context**

## Database schema

| Field Name       | Example                      | Data Type | Description                                       |
|------------------|------------------------------|-----------|---------------------------------------------------|
| `country`        | JP                           | TEXT      | ISO 3166 country code of the location             |
| `country_name`   | Japan                        | TEXT      | Name of the country                               |
| `continent`      | AS                           | TEXT      | Continent code of the country                     |
| `continent_name` | Asia                         | TEXT      | Name of the continent                             |

Documentation: https://ipinfo.io/developers/ip-to-country-database

We also have a free database that combines country and ASN information in a single database: https://ipinfo.io/developers/ip-to-country-asn-database

Samples are available here: https://github.com/ipinfo/sample-database/tree/main/IP%20to%20Country

The database can be downloaded simply by accessing the [storage URI with an access token](https://ipinfo.io/developers/database-filename-reference).

```bash
curl -L https://ipinfo.io/data/free/country.mmdb?token=<YOUR_TOKEN> -o country.mmdb
```

Let me know what you think. Thanks

enoch85:
@abdullahdevrel Thank for your proposal. Please check the PR and approve "attribution requirements".

[...]

PR #2683, opened 2024-09-22
Use IPInfo as secondary source

enoch85:
Fix https://github.com/nextcloud/vm/issues/2682

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -417,7 +417,6 @@ download_geoip_mmdb() {
     } > /tmp/dockerenv
     unset MwKfcYATm43NMT
     unset i9HL69SLnp4ymy
-    unset x8v8GyVQg2UejdPh
     install_docker
     if docker run --name maxmind --env-file /tmp/dockerenv -v "$GEOBLOCK_DIR":"$GEOBLOCK_DIR" ghcr.io/maxmind/geoipupdate
     then
@@ -433,6 +432,7 @@ download_geoip_mmdb() {
             return 1
         fi
     fi
+    unset x8v8GyVQg2UejdPh
 }
 
 # Check if process is runnnig: is_process_running dpkg
--- a/lib.sh
+++ b/lib.sh
@@ -425,7 +425,7 @@ download_geoip_mmdb() {
     else
         docker rm -f maxmind
         rm -f /tmp/dockerenv
-        print_text_in_color "$ICyan" "Update limit for Maxmind GeoDatabase reached! We're now trying to get the Country Database from https://ipinfo.io instead."
+        print_text_in_color "$ICyan" "Rate limit for Maxmind GeoDatabase reached! We're now trying to get the Country Database from https://ipinfo.io instead."
         if ! curl -sfL https://ipinfo.io/data/free/country.mmdb?token="$x8v8GyVQg2UejdPh" -o "$GEOBLOCK_DIR"/GeoLite2-Country.mmdb
         then
             msg_box "Sorry, we couldn't get the needed IP geolocation database from any source, please try again in 24 hours."
--- a/lib.sh
+++ b/lib.sh
@@ -156,6 +156,7 @@ HTTP2_CONF="/etc/apache2/mods-available/http2.conf"
 # GeoBlock
 GEOBLOCK_MOD_CONF="/etc/apache2/conf-available/geoblock.conf"
 GEOBLOCK_MOD="/etc/apache2/mods-available/maxminddb.load"
+GEOBLOCK_DIR="/usr/share/GeoIP"
 # PHP-FPM
 PHPVER=8.3
 PHP_FPM_DIR=/etc/php/$PHPVER/fpm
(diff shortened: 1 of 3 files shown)


enoch85:
Hmm, this doesn't seem to work. Allowing "SE" in the list still blocks it when trying to reach the site.

@abdullahdevrel Are your `mmdb` different from Maxminds?

enoch85:
Example from CSV:

```
start_ip	end_ip	country	country_name	continent	continent_name
1.0.0.0	1.0.0.255	AU	Australia	OC	Oceania
1.0.1.0	1.0.3.255	CN	China	AS	Asia
1.0.4.0	1.0.7.255	AU	Australia	OC	Oceania
1.0.8.0	1.0.15.255	CN	China	AS	Asia
1.0.16.0	1.0.31.255	JP	Japan	AS	Asia
```

In the Apache conf, we use different "tags". This might not work after all?

enoch85:
Found this now: https://community.ipinfo.io/t/migrating-from-maxmind-geolite2-country-to-ipinfo-country-mmdb-edition/56

enoch85:
So, since IPInfo doesn't run it's own Apache module, I'm afraid this won't work. :|

abdullahdevrel:
@enoch85 Thank you for reviewing. The scheme is different from MaxMind's, and you have found the post that I would have linked. 

I believe the Apache module (https://github.com/maxmind/mod_maxminddb) is schema agnostic, like all the other MMDB reader libraries out there (https://community.ipinfo.io/t/list-of-mmdb-reader-libraries/2821).

Looking at this bit of code:

https://github.com/nextcloud/vm/blob/6a111c790e6d0207a5778d14c71af26bbf3a100e/network/geoblock.sh#L224-L227

So, with respect to our database schema, I assume this will be:

```
  MaxMindDBEnv MM_CONTINENT_CODE DB/continent
  MaxMindDBEnv MM_COUNTRY_CODE DB/country
```

Can you see if it would work?

enoch85:
> @enoch85 Thank you for reviewing. The scheme is different from MaxMind's, and you have found the post that I would have linked.
> 
> I believe the Apache module (https://github.com/maxmind/mod_maxminddb) is schema agnostic, like all the other MMDB reader libraries out there (https://community.ipinfo.io/t/list-of-mmdb-reader-libraries/2821).
> 
> Looking at this bit of code:
> 
> https://github.com/nextcloud/vm/blob/6a111c790e6d0207a5778d14c71af26bbf3a100e/network/geoblock.sh#L224-L227
> 
> So, with respect to our database schema, I assume this will be:
> 
> ```
>   MaxMindDBEnv MM_CONTINENT_CODE DB/continent
>   MaxMindDBEnv MM_COUNTRY_CODE DB/country
> ```
> 
> Can you see if it would work?

Hmm, even if it would work it would mean we would have to maintain (or at least put checks in) to know which DB is used. So that would also mean that we would have to choose from either IPInfo or Maxmind. 

I will test anyway so that we'll know if that's the case.

enoch85:
And yes, I can confirm, the change works.

Now I need to get my head around how we want to do this...

enoch85:
Found a solution!

```
<IfModule mod_maxminddb.c>
  MaxMindDBEnable On

  # Check for IPinfo mmdb
  <IfFile "/usr/share/GeoIP/IPInfo-Country.mmdb">
    MaxMindDBFile DB /usr/share/GeoIP/IPInfo-Country.mmdb
    MaxMindDBEnv MM_CONTINENT_CODE DB/continent
    MaxMindDBEnv MM_COUNTRY_CODE DB/country
  </IfFile>
  # Check for Maxmind mmdb
  <IfFile "/usr/share/GeoIP/GeoLite2-Country.mmdb">
    MaxMindDBFile DB /usr/share/GeoIP/GeoLite2-Country.mmdb
    MaxMindDBEnv MM_CONTINENT_CODE DB/continent/code
    MaxMindDBEnv MM_COUNTRY_CODE DB/country/iso_code
  </IfFile>
</IfModule>
```

enoch85:
OK, so this is tested and works!

@abdullahdevrel Please confirm that the  "attribution requirements" are enough.

lib.sh
```diff
@@ -414,18 +416,29 @@ download_geoip_mmdb() {
...
         docker rm -f maxmind
         rm -f /tmp/dockerenv
+        # Since only one mmdb file can exist at the same time due to Apache "if" confitions, remove IPInfos config
+        rm -f "$GEOBLOCK_DIR"/IPInfo-Country.mmdb
     else
         docker rm -f maxmind
         rm -f /tmp/dockerenv
-        msg_box "Update limit for Maxmind GeoDatabase reached! Please try again tomorrow."
-        return 1
+        print_text_in_color "$ICyan" "Rate limit for Maxmind GeoDatabase reached! We're now trying to get the Country Database from https://ipinfo.io instead."
```
enoch85, line 430: @abdullahdevrel This is what I need approval on.

enoch85: I'll merge this now, please come back with a suggestion if the above is not enough.

PR #2691, opened 2024-10-13
geoblock: update csv list url

szaimen:


enoch85:
Thanks! So, what do you think of the automatic magic now implemented? :)

szaimen:
It is really nice that we can retreive the db now automatically! 🎉 
Thanks a lot for implementing :)

PR #2704, opened 2024-11-28
Enable Brotli support

enoch85:


enoch85:
FYI  @szaimen 

Maybe something for AIO?

Issue #1819, opened 2021-02-15
Torrent


ShalokShalom:
Hi there

Since your downloads are quite sizable and since my download just failed, do I suggest offering torrent files next to the casual downloads. 

This also reduces the load on your server, if that's any concern.

ShalokShalom:
Yeah, good call. 

Let's try this, I havent done this myself, while I am open doing so. 

I do work on this, once my instance of Nextcloud is productive. :)

lukaspavelka:
Maybe i can also help with it, if you need my help please let me know

enoch85:
@lukaspavelka 

> Maybe i can also help with it, if you need my help please let me know

Sure, would be great if a torrent could be created. We need to script it so that it can be done automatically each time a new release is made.

lukaspavelka:
Maybe some basic stuff where to start is: 
`for ova in $(find . -type f -name "*.ova" | awk -F '.*/var/www' '{print $2}'); do bash -c "create-torrent $ova --urlList 'http://127.0.0.1/$ova/{}' -o ${1%.ova}.torrent" sh {}  \; -print; done
`

Or implement php-torrent solution from here: 
[https://github.com/christeredvartsen/php-bittorrent](url)

enoch85:
Thanks @lukaspavelka!

Would be great if you could have a look exactly how it's done and then send a PR this way. :+1:

lukaspavelka:
@enoch85 can you specify what exactly need to be done in what order so i can prepare some solution for it? I need more specifics.

enoch85:
When I do releases I work with VMware Workstation (on Windows, yeah I know). So for each new release I would need a script, Powershell maybe (?)(I don't know) which creates a torrent for each new release I make. 

The torrent itself should then be auto uploaded to some public tracker so that downloads could start immediately. Another alternative would be to just create the torrent so that I could upload it using any BitTorrent client out there. 

I hope that makes sense, I'm not that good with creating torrents, only VMs. ;)

szaimen:
I mean the coolest thing would be if there was an app for Nextcloud that would integrate a torrent server into Nextcloud but that's most likely at least currently not doable...

szaimen:
Here was a peoposal that might work: https://help.nextcloud.com/t/torrent-server-for-file-share/88707/5

enoch85:
[...]

PR #2728, opened 2025-02-25
Update & Fix torrent creation

celeroncool:
Hello!

I saw that the torrent creation script wanted some love :)

## Modifications:
- Make script use default transmission download directory so it works out of the box.
- Make script fix default transmission systemd init file per https://github.com/transmission/transmission/issues/6991
- Do not download the VM file if it already exists to save bandwidth.

## Open questions:
- At its current stage, the script has no idea if the source files have been updated -> torrent will eventually be out of date.
- Would it be at all possible to have a SHA256 stamp available for download from your server, so we can first hash the local file, compare against your file and download new version?
- Naming the file always NextcloudVM.zip causes other issue, with the torrent naming & distribution. As the torrent file is created, there is no way to update the file later. Would it be better to name the torrent to include what version of Nextcloud VM it is and also what platform it is?

## Related Issue
https://github.com/nextcloud/vm/issues/1819

PS: I will seed this file for the foreseeable future :)


celeroncool:
I changed this to a draft so I can also finalize updating the trackers.txt file, as it has multiple offline trackers.
Would it be OK to use: https://github.com/ngosang/trackerslist/blob/master/trackers_best.txt

celeroncool:
I updated the tracker list & tested working with new tracker list:
image

Could

PR #2729, opened 2025-02-27
cosmetic fixes to torrent script

enoch85:


enoch85:
@celeroncool Hmm even with these fixes, the torrent just stalls for me. Any idea?

celeroncool:
Could you check the hash of the file so we could compare on what is created on my side when I create the torrent?

### Running the script inside clean Ubuntu 24.04 VM creates a torrent with info hash v1: 14fc1e1b47f0606fe8b5b304f6b4a1d730dffb8f
sha256sum from same file:
84b1dae36c0962721ec3ab46485e8b5d5b1d06e5485be522eb52f6df5552028c  NextcloudVM.zip

### For some reason downloading it with curl inside freshly provisioned NCVM produces following output:
7a47c12b236e3029660faaf4d4c51af3a5c807db077c28fc7c8c504ee15a4eb9  NextcloudVM.zip

### Downloading it directly with browser from link at https://download.kafit.se/s/dnkWptz8AK4JZDM -> download produces this hash:
21fdaed9049b2220d6fbf8179563d091177bf7c045971340fa60ddc0b9ffe0b6  NextcloudVM.zip

So there is something funky going on with nextcloud zip compression because the hash changed between downloads.

I then only download the actual OVA file inside the nextcloud VM:
dd38137e64a310e581fc295e1b42dd042bc07fc7fea3aa13bee9180d10085b14  100GB_Nextcloud-VM_www.hanssonit.se.ova

Same file downloaded on my machine:
dd38137e64a310e581fc295e1b42dd042bc07fc7fea3aa13bee9180d10085b14  100GB_Nextcloud-VM_www.hanssonit.se.ova

Now this file has torrent info hash:
0b36c11183e9b84f08e23e7b6b378fa80534e767

celeroncool:
Also, my editor added unecessary "" around the cat command for the torrent creation, so it would not add the trackers...
So a working command is:
```
transmission-create -o $TRANSMISSION_DL_DIR/nextcloudvmhanssonit.torrent -c "https://www.hanssonit.se/nextcloud-vm" -t $(cat /tmp/trackers.txt) "$TRANSMISSION_DL_DIR"/"$NC_ZIP"
```

enoch85:
You're probably right, We can't download the zip content, and we need to use the direct OVA for this. You torrent works, and I downloaded it.

I will make some more changes to get it just as I want it. :)

Thanks a lot for looking into this, it helps a lot. ⭐

PR #2752, opened 2025-06-10
Remove "Prefer/force IPv4 for apt"

UnderEu:
I can understand the original intention, autonomous systems had issues with the current Internet Protocol a.k.a. IPv6 over the past years but this protocol is now very well established, mature and, IMO, there is no reason to a single application dictates how an operating system (or package manager in this regard) will behave regarding the system's Internet connectivity especially if the target environment has no issues with it and, if the opposite, masks any ongoing issue particular to the host OS, ISP and/or the server on the other end.

Circumventing things in the intention to get them done quick does not fix the root cause and it's better to put the effort once to fix things in a definitive way than sweeping it under the carpet to deal with it later and over and over. Any application shouldn't touch anything external to and/or broader than its own scope unless it's strictly necessary and the user is aware of any change & acknowledges it.

### DIFF
--- a/nextcloud-startup-script.sh
+++ b/nextcloud-startup-script.sh
@@ -584,10 +584,6 @@ nextcloud_occ maintenance:repair --include-expensive
 apt-get autoremove -y
 apt-get autoclean
 
-# Remove preference for IPv4
-rm -f /etc/apt/apt.conf.d/99force-ipv4
-apt-get update
-
 # Success!
 msg_box "The installation process is *almost* done.
 


GoetzGoerisch:
Thank you @UnderEu for proposing this improvement.

@enoch85 to your concern: Why would one want to force IPv4 if connecting using the legacy protocol is a large issue in many parts of the world nowadays?

There was the same issue with Pi-Hole a couple of month back: https://github.com/pi-hole/pi-hole/pull/6144

f2k1de:
> I totally agree with you, but at the same time we want the scripts to work _everywhere_, not only on [50% of the servers](https://www.google.com/intl/en/ipv6/statistics.html).
> 
> I'm not convinced.

@enoch85, please stop implying that IPv4 is available on every system. That's not the case and even a decreasing trend.

enoch85:
OK, I rest my case. IPv6 for the people!

PR #2767, opened 2025-10-13
Switch Adminer source back to official upstream

ItzSteveHuh:
Updated the Adminer installation script to download from the official [adminer.org](https://www.adminer.org/) Source instead of the deprecated adminerevo.org fork. The original upstream project is now maintained again. Commented out the previous download and extraction steps for reference.


enoch85:
Thanks for your efforts, just noticed this PR. I will actually implement AdminerNeo as suggested by users.

PR #2816, opened 2026-04-08
Fix app re-enable after upgrade (#2797)

### LINKED ISSUE #2797: Apps are disabled during update but not re-enabled
### Steps To Reproduce

1. Go to 'ssh console'
2. Use update_nextcloud
3. Wait to finish
4. Enter Nextcloud with admin
5. See many installed and previously enabled apps, now are disabled
6. Manually re enable installed apps

### Expected Result

The installed and enabled apps should be enabled after update

### Actual Result

Many installed apps are disabled after update, even when they are compatible apps.

### Screenshots, Videos, or Pastebins

_No response_

### Additional Context

The console output shows many lines saying: "Disabled incompatible app: [app name]"

After that, for each disabled app, I can see the following lines:
```
Note: This command triggers the upgrade actions associated with a new version. The new version's updated source files must be deployed in advance.
See the upgrade documentation: https://docs.nextcloud.com/server/32/go.php?to=admin-update for more information.
Restoring tasks from /mnt/NCBACKUP/apps...
Setting secure permissions...
Creating possible missing Directories
chmod Files and Directories
chown Directories
chmod/chown .htaccess
No such app enabled: [app name]
No upgrade required.
```

Additionally, for each disabled app, a line appears in Nextcloud log:

```
Error    no app in context    AppPathNotFoundException Could not find path for [app name]
```

### Build Version

Nextcloud 32.0.5

### Environment

By using the scripts, By downloading the VM

### Environment Details

Ubuntu 22.04 with PHP 8.3 in a ESXi 7 VM

notarealnumber:
Fixes #2797

## Problem

The `enabled_apps_before_upgrade.txt` is saved to `$BACKUP` in line 369, but the backup
rotation (formerly lines 1157-1164) moves everything to `$BACKUP-OLD/<timestamp>/`, deletes
`$BACKUP`, and recreates it empty. When the re-enable logic runs at line 1366, the file is
gone. The condition silently evaluates to false and no apps are re-enabled.

## Root cause

`$BACKUP` pointed to a fixed path (`/mnt/NCBACKUP`) that was reused and rotated during
the same script run.

## Fix

- **lib.sh**: `$BACKUP` now includes a timestamp (`/mnt/NCBACKUP/<timestamp>`), giving each
  run its own directory. No rotation needed, no file loss.
- **nextcloud_update.sh**: Remove the rotation block. Add cleanup of backups older than one
  year. Fix the legacy migration block (`/var/NCBACKUP` → new structure). The re-enable
  logic works unchanged since `$BACKUP` now persists throughout the script.

## Additional fixes

- Eliminates a race condition where `$(date +%Y-%m-%d-%H%M%S)` was called separately for
  `mkdir` and `rsync`, potentially producing two different timestamps.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -19,7 +19,7 @@ NCDATA=/mnt/"$POOLNAME"
 SNAPDIR=/var/snap/spreedme
 GPGDIR=/tmp/gpg
 SHA256_DIR=/tmp/sha256
-BACKUP=/mnt/NCBACKUP
+BACKUP=/mnt/NCBACKUP/"$(date +%Y-%m-%d-%H%M%S)"
 NC_APPS_PATH=$NCPATH/apps
 VMLOGS=/var/log/nextcloud
 
--- a/nextcloud_update.sh
+++ b/nextcloud_update.sh
@@ -359,7 +359,8 @@ then
     export NEEDRESTART_SUSPEND=1
 fi
 
-# Save the list of enabled apps before upgrade (to re-enable them after upgrade)
+# Create the backup dir (/mnt/NCBACKUP/"$(date +%Y-%m-%d-%H%M%S)") and save
+# the list of enabled apps before upgrade (to re-enable them after upgrade)
 # https://github.com/nextcloud/vm/issues/2797
 print_text_in_color "$ICyan" "Saving list of enabled apps before upgrade..."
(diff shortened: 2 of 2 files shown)


enoch85:
Thanks for the PR!

I missed that this was already posted. And that's merged so I'll close this one.

PR #2833, opened 2026-08-03
Use dracut to generate initramfs

pq2:
tpm2-unlock.sh wasn't working for me using Ubuntu 26.04. After the reboot the initramfs could not be loaded and the device started a busybox shell.

After using dracut instead of update-initramfs as described [here](https://ubuntu.com/server/docs/how-to/security/tpm-backed-luks-decryption-with-clevis/) it worked.

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/not-supported/tpm2-unlock.sh
+++ b/not-supported/tpm2-unlock.sh
@@ -20,7 +20,7 @@ debug_mode
 root_check
 
 # Check if already installed
-if is_this_installed clevis-luks || is_this_installed clevis-tpm2 || is_this_installed clevis-initramfs
+if is_this_installed clevis-luks || is_this_installed clevis-tpm2 || is_this_installed clevis-initramfs || is_this_installed clevis-dracut
 then
     msg_box "It seems like clevis-luks is already installed. We are trying to do the configuration again."
 else
@@ -71,7 +71,7 @@ then
 fi
 
 # Install needed tools
-apt-get install clevis-tpm2 clevis-luks clevis-initramfs -y
+apt-get install clevis-tpm2 clevis-luks clevis-initramfs clevis-dracut -y
 
 # Execute the script
 print_text_in_color "$ICyan" "Setting up automatic unlocking via TPM2..."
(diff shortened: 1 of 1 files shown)


enoch85:
Thanks for the PR! 🚀

not-supported/tpm2-unlock.sh
```diff
@@ -71,21 +71,21 @@ then
...
 We will now uninstall all needed packets again, so that you are able to start over."
-    apt-get purge clevis-tpm2 clevis-luks clevis-initramfs -y
+    apt-get purge clevis-tpm2 clevis-luks clevis-initramfs clevis-dracut -y
     apt-get autoremove -y
     msg_box "All installed packets were successfully removed."
     exit 1
 fi
 print_text_in_color "$ICyan" "Updating initramfs..."
-if ! update-initramfs -u -k 'all'
+if ! dracut -f
```
szaimen, line 88: Hi, I am not sure if this will continue to work on Ubuntu 24.04. Maybe we should use dracut on 26.04 and update-initramfs on ubuntu 24.04. WDYT @pq2?

pq2: I'm not entirely sure about that either. While `clevis-dracut` is available in the repositories for 24.04 ([Source](https://packages.ubuntu.com/clevis-dracut)), dracut doesn't seem to be the default tool for creating the initramfs in 24.04. It can't hurt to use both, depending on the version, as you suggested.

szaimen: Yes lets do that. Would you be able to create a PR?

enoch85: Note, we have variables for OS checks.

szaimen: https://github.com/nextcloud/vm/pull/2834

pq2: Thanks for creating the PR. I was too late.

